PR #104, opened 2012-07-19
Bcb5support

patrickmmartin:
Hi Phil, as discussed, here is the sum total of the changes for BCB5 (and the free 5.5 compiler).

My first pull request, so here goes...

Above all: please note the breaking change in

include/internal/catch_commandline.hpp L 64

this breaks VS 2010 at very least, but otherwise BCB5 is broken, as best I can determine.

I presume omit this line if you do want to accept the patch, but I've left it in for our reference.  

The remaining controversy is from BCB getting caught out in resolving mapConsoleColour

include/internal/catch_console_colour_impl.hpp L 57

Finally, there's a working BCB project with a command line build available via a Borland make compatible makefile.

Kind regards, Patrick

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,8 @@ projects/VS2010/TestCatch/_UpgradeReport_Files/
 projects/VS2010/TestCatch/TestCatch/TestCatch.vcxproj.filters
 projects/VisualStudio/TestCatch/UpgradeLog.XML
 UpgradeLog.XML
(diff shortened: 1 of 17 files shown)


horenmar:
@patrickmmartin If you still want to, you can submit a new PR adding BCB support against current Catch, but I am going to close this PR, as Catch has changed a LOT in the meantime.

patrickmmartin:
Sure

PR #178, opened 2013-06-28
Update Makefile for SelfTest

mloskot:
This PR updates Wichert's Makefile from #141:
- adds new .cpp files
- renames output filename to catch_self_test
- adds check target


horenmar:
Since the current plan is to keep only cmake, I am closing this.

Issue #83, opened 2012-05-11
Avoid warnings for old-style casts and shadowing.


mgeier:
When compiling with GCC 4.4.5 there are warnings with certain settings (line numbers are for single_include header file 176eb81563):

GCC option `-Wshadow`:

```
catch/catch.hpp|77| warning: declaration of ‘line’ shadows a member of 'this'
catch/catch.hpp|77| warning: declaration of ‘file’ shadows a member of 'this'
```

This should come from `catch_common.h` and could be fixed by using `line_` and `file_` for the constructor arguments.

```
catch/catch.hpp|4859| warning: declaration of ‘colour’ shadows a previous local
catch/catch.hpp|4855| warning: shadowed declaration is here
```

This should come from `../internal/catch_reporter_registrars.hpp`.
I think this is a little harder to resolve, because the same variable name was chosen on purpose.
Maybe some macro trickery could be used to avoid variables with the same name:

```
#define CONCATENATE_STUFF_HELPER(a, b) a ## b
#define CONCATENATE_STUFF(a, b) CONCATENATE_STUFF_HELPER(a, b)
#define TEXTCOLOUR(colour) TextColour CONCATENATE_STUFF(tmp, __LINE__)( TextColour::colour );
```

This could then be used like `TEXTCOLOUR( ResultError );`. This compiles for me but I don't know if it works because it seems that colours are only supported for Windows.

GCC option `-Wold-style-cast`:

```
catch/catch.hpp|580| warning: use of old-style cast
```

This should come from `catch_tostring.hpp` and could be fixed like this: `return toString( static_cast<unsigned long>(value) );`

PR #188, opened 2013-07-14
Avoid GCC warnings (-Wold-style-casts and -Wshadow)

mgeier:
I worked around a few warnings which my GCC (4.7.2) spat out.

See also issue #83.


mgeier:
A few years have passed ... in the meantime, all warnings have disappeared, that's great!

PR #199, opened 2013-09-27
Temporarily disabled VC warning C4702.

Fraser999:
The warning is for unreachable code (around [line 166 of clara.h](https://github.com/philsquared/Catch/blob/master/include/internal/clara.h#L166) shown below):

```
virtual void setFlag( C& obj ) const {
    typename RemoveConstRef<T>::type value;
    convertInto( true, value );  // This throws if value is of type std::string
    function( obj, value );  // warning C4702
}
```

This ultimately comes from `bindProcessName`, where `convertInto` has args of type `bool` and `std::string`.


ben-crowhurst:
+1  These warnings are driving us crazy.

horenmar:
I tested this and v1.7.1 does not trigger C4702 warning. Unless someone still has trouble, I am going to close this soon.

vadz:
I can confirm that the warning is gone with 1.7.1, at least when using MSVS 2015. Thanks!

philsquared:
Thanks, @vadz

PR #205, opened 2013-10-21
Added VS2013 (RC) test project

cdmh:
VS2013(RC) project based on previous VS projects


horenmar:
We are planning to move over to having only `CMakeLists.txt` inside the repository with IDE-specific files being generated by CMake.

PR #206, opened 2013-10-21
Eliminated objects that hold constant strings. Aids step-debugging in MSVC

cdmh:
This is a re-submission of the second half of my previous pull request https://github.com/philsquared/Catch/pull/203

I've eliminated std::string objects that hold constant strings, and reverted to char const*. This helps step-debugging in MSVC where "step-into" a CHECK (for example) steps into several ctors and dtor for creation/destruction of temporaries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.0 build 10 (master branch)*
+*v1.0 build 11 (master branch)*
 
 # New release with significant changes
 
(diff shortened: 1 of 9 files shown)


michaelbprice:
My opinion, the provided rationale for making this kind of change is not sufficient.

There are other ways to solve this problem (probably better for the sanity of developers too). See https://msdn.microsoft.com/en-us/library/dn457346.aspx#BKMK_C___Just_My_Code

There may however be other reasons to make this change.  I recommend against merging this pull request and closing it instead.

nabijaczleweli:
Using raw pointars (obtained from possibly-malicious sources, *gasp*)! :-1:

horenmar:
While Catch does allocate `std::string` quite freely and could stand to do so less, it would be better to do so with something like `string_view`, not raw C-strings.

PR #210, opened 2013-10-27
use __COUNTER__ instead of __LINE__ for unique name in INFO macro

### LINKED ISSUE #209: Can't use INFO macro in other macro
For example, if you want to use approximate comparison and print arguments.

```
#define CHECK_IS_APPROX(E1, E2, rtol)\
    {\
        INFO(#E1 << " :=\n" << E1)\
        INFO(#E2 << " :=\n" << E2)\
        CHECK( isApprox(E1, E2, rtol) );\
    }
```

This leads to multiple declarations of Catch::ScopedMessage.

Slonegg:
Resolve #209


philsquared:
I implemented a version of this a while ago on the develop-v2 branch (see #351) and hadn't realised I never got around to merging it into master. I've now done that.

PR #215, opened 2013-11-07
Added headers to targets in CMakeLists.txt

Fraser999:
I've added all the headers currently in the source tree to the CMakeLists file.  This is a convenience feature which allows these to show up in IDEs - it doesn't affect the build in any way.

I'd personally favour having the lists in a CMake file per directory rather than cluttering up the main CMakeLists.txt, but that's personal preference.  At the moment, there's only a single CMake file in the project and perhaps it's better to keep it that way?

I've also added a small function which checks that no source (.cpp) or header (.hpp / .h) files have been omitted from the hard-coded CMake variables.  It only issues a developer warning if this does happen, so unless it's a required source file which is missing, it won't actually break the build.


horenmar:
Most of this PR has found its way into master one way or another, but I will look into  3431e25befa3f07c1ce03a63053bb483ac02af1b because that functionality might be useful in the future.

horenmar:
I have incorporated functionality from 3431e25 into master, so I am closing this.

PR #216, opened 2013-11-10
Visual Studio integration part 1 - allow catch_message.hpp to be included by more than 1 source file

colonelsammy:
As discussed with Phil by email.

The counter for MessageInfo is now templated - this forces the compiler/linker to use greedy instantiation (or whatever process it needs to use) to remove duplicates of the static from the executable.

Also tidied up the VS2010 TestCatch project and delele the SelfTest executable in 'make clean' on *nix.

Tested on Windows (VS2010 & MingW) and Ubuntu 12.10.  Self tests all pass.

single include has _not_ been regenerated.


BMBurstein:
Is this still live?

janisozaur:
This looks like a nice feature, even though I don't use VS. In case this goes in, please make sure to optimize PNGs before you add them to git (so as to keep only one instance of already optimized screens).

philsquared:
Sorry, @colonelsammy, I never did around to this. Was it still relevant at all?

colonelsammy:
I think it is so old that your chances of merging are about zero...lots has changed in Catch since then! I havn't used it recently so I've not updated it...I'll leave my clone with the changes on my repo if anyone is interested.

I think your new employer has a plug-in that runs Catch? Perhaps we should encourage folks to use that ;-)

philsquared:
Ha! Yes ReSharper C++ has a test runner that now supports Catch.
That's not an option for everyone, though.
I was more thinking whether its worth still trying to support VS2010 for this - given that later VSs are very different in this regard too.
While there are always plenty of people stuck on older IDEs I doubt it's worth it in this case, given how much was involved - but that's where I was asking if it was still relevant to you. Sounds like not, so I'll leave it be.

BMBurstein:
Does this mean that there is a solution for newer editions of VS? Specifically VS2015?

horenmar:
@BMBurstein 3rd party only, in the aforementioned ReSharper C++ plugin. For VS 2015 without plugins, not right now.

PR #232, opened 2013-12-17
re-added signal handling to the latest Catch (fetched from latest upstream) (re-adding #160)

formiaczek:
1. #160 for the latest master branch.
2. Also added time-out for the whole test suite - In order to make use of it, before including <catch...> one needs to define something like:
# define TIMEOUT_FOR_ALL_TESTS_IN_SECONDS 3 \* 60

This is extremely useful to have: instead the whole testing might be hang if the code / test code enters the never-ending loop. Above will make it stop and fail instead, giving a bit of a clue on where the hanging code might be.

Implementation uses pthreads, so it's not available for MSVC compilers (unless with pthreads-win32), so for non-gnu compilers it only produces a warning if above is defined. This can be easily changed / updated once catch moves to C++11 (with std::thread), but can perhaps be fixed / updated by someone else here (i.e. using Windows native threads etc.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.0 build 19 (master branch)*
+*v1.0 build 23 (master branch)*
 
 Build status (on Travis CI) [![Build Status](https://travis-ci.org/philsquared/Catch.png)](https://travis-ci.org/philsquared/Catch)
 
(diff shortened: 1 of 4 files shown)


horenmar:
Signal handling is in, timeouts are problematic (you can only get them nonportably, the easy way on Linux means much more dangerous signal handling, Windows is even worse) and for now probably best left for outside scripting.

Closing this, if someone has a sane idea for timeouts, new PR can be opened.

PR #237, opened 2014-01-13
JUnit reporter: Added an ISO8601 timestamp to the timestamp attribute.

sevas:
The current JUnit reporter uses the _"tbd"_ string for the `timestamp` attribute in testsuite nodes.

I use Catch in a context where libraries are built for several platforms (nothing really original here), and then aggregated in one package. The testresults are also aggregated, and I display the full set of results using the Jenkins JUnit publisher.

The main problem with this is that the JUnit parser in Jenkins somehow discards testsuite nodes if their name were already encountered **and** the timestamp is something it does not understand.

For instance, if I run the same 10 testcases on 3 platforms, the aggregated results report (as well as the _test trends_ graph) only shows me 10 test results (instead of the 30 I'm expecting). I'm pretty sure it is also due to something fishy inside the JUnit parser, but modifying Catch is way more easy.

I did not find any discussion about that topic in the Catch google group.

This simple patch replaces that _"tbd"_ string with an actual timestamp string. I used the ISO8601 format because [it seems to be what the JUnit spec is expecting](http://windyroad.com.au/dl/Open%20Source/JUnit.xsd).

This patch currently solves the aforementioned issue with Jenkins testresults aggregation.

I did not regenerate the single header redistribuable as part of this PR because my assumption is that releases are made from the master branch.

### DIFF
--- a/include/reporters/catch_reporter_junit.hpp
+++ b/include/reporters/catch_reporter_junit.hpp
@@ -89,6 +89,9 @@ namespace Catch {
             else
                 xml.writeAttribute( "time", suiteTime );
 
+            time_t rawtime;
+            struct tm *timeinfo;
+
 #ifdef CATCH_PLATFORM_WINDOWS
             // %z does not return the offset from UTC with MSVC's strftime implementation
             // Timestamps will look like: "2014-01-14T17:22:09"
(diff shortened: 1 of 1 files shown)


horenmar:
I reimplemented the functionality a bit differently (forced UTC in timestamps because %z is not backward compatible, factored it out of the junit writer), but thank you for the PR, I wouldn't even know it needs changing otherwise.

PR #240, opened 2014-02-12
add VS2012 project and solution files

Kosta-Github:


Kosta-Github:
It is a bit outdated now; closing ...

PR #241, opened 2014-02-12
add support for `duration threshold`

Kosta-Github:
This adds a new option `-q <seconds>` for specifying a `duration threshold` value for printing out duration timing results in the basic console reporter: only if the given threshold (given as a `double` value in seconds) is exceeded the duration is printed out through the console reporter.

This PR replaces my earlier PR #189 since that was not up-to-date anymore with all the recent changes in the `master` branch...


Kosta-Github:
It is a bit outdated now; closing ...

PR #247, opened 2014-02-22
switch from ->* to <= in expression decomposer

tclamb:
Thanks for the awesome test framework. The idea is really clever!

One problem I had noticed is that comparisons were asymmetric. Tests such as `REQUIRE( 1+1 == 2 )` would fail, but `REQUIRE( 2 == 1+1 )` would pass. This pull request makes tests symmetric.

The expression decomposition seemed like magic, so I spent some time figuring it out and writing my own [minimal working implementation](https://gist.github.com/tclamb/9149587). During that, I realized expressions containing common binary operators can be used on the LHS if we switch from `->*` to a comparison operator.

This pull request allows expressions containing any of `*`, `/`, `%`, `+`, `-`, `<<` and `>>` to be used on the LHS.

`<=` was chosen arbitrarily from comparison operators, but any of `<=`, `>=`, `<`, and `>` have the same precedence.

No tests broke, and the corresponding originally failing test in `[tricky]` was updated and added to the main suite.

### DIFF
--- a/include/internal/catch_capture.hpp
+++ b/include/internal/catch_capture.hpp
@@ -33,7 +33,7 @@
     do { \
         Catch::ResultBuilder __catchResult( macroName, CATCH_INTERNAL_LINEINFO, #expr, resultDisposition ); \
         try { \
-            ( __catchResult->*expr ).endExpression(); \
+            ( __catchResult <= expr ).endExpression(); \
         } \
         catch( ... ) { \
             __catchResult.useActiveException( Catch::ResultDisposition::Normal ); \
--- a/include/internal/catch_result_builder.h
+++ b/include/internal/catch_result_builder.h
@@ -41,8 +41,8 @@ namespace Catch {
                         ResultDisposition::Flags resultDisposition );
 
         template<typename T>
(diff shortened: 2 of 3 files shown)


tavianator:
You might be able to avoid the `-Wparentheses` warnings by using `<<` instead of `<=`.  Haven't tried though.

philsquared:
Hi,
It's been a while - this dropped off my radar for a bit - bit since the discussion picked up again recently in #359 (albeit for different reasons) I finally took a closer look at this.
I think I may very well adopt your proposal now - (not sure when it will actually make it in, though). Thanks for bringing it up here, @tclamb 

As for the `-Wparentheses` warning, @Morwenn  - I don't seem to get it for `<=` - but I do for `<<`, funnily enough.
I'm not sure what's different between mine and yours is. Both the code-base and the compiler have moved on since this was originally raised - so maybe that's it?

Morwenn:
@philsquared For sure it's been a while. I will try again this weekend with the current code base and check whether I still have the warnings.

tclamb:
Glad to see this make it in. :+1: Don't forget about the test in `projects/SelfTest/TrickyTests.cpp`.

PR #285, opened 2014-06-05
Fix compilation error with old GCC versions.

arturocastro:
Compilation failed with GCC 4.3.3 under QNX 6.4.1. This use of
auto_ptr's operator= is not supported. Change to reset() which has the
same functionality with new pointers, applying to both auto_ptr and
unique_ptr.


horenmar:
As far as I can tell, these changes are already in Clara in master, thus I will be closing the PR.

PR #310, opened 2014-08-25
Deleting trailing whitespaces

jubalh:
Deleting trailing whitespaces


jubalh:
I updated this PR so it can be automerged again!

philsquared:
Thanks for doing that (twice!).
However I have addressed the issue closer to the root by adding a script (scripts/fixTrailingWhitespace.py) that will do this automatically. I just need to remember to run it every now and then (I'll probably call it from the developBuild.py script, actually).

I've run the script now and checked in all the fixed files - so I'm going to close this PR on that basis. If you think I've missed something feel free to re-open or otherwise let me know.

PR #313, opened 2014-08-26
Added VS2013

HungMingWu:


horenmar:
We are moving to `CMakeLists.txt` only, with IDE-specific projects to be generated by Cmake.

PR #343, opened 2014-10-12
use variadic macros to make tests with templates code easier

rnburn:
Hi, would you consider modifying some of the macros so that they will work more easily with template code with multiple arguments. For example, a check like this

REQUIRE(f<A, B>() == 3);

which wouldn't work now unless wrapped with an extra set of parenthesis could be made to work without any trouble by using a variadic version of REQUIRE

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.0 build 53 (master branch)*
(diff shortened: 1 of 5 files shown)


horenmar:
Well, it has been a while, but this has been done separately for #860 and will be part of Catch 2, so I am going to close this.

PR #351, opened 2014-11-10
Made test cases based on __COUNTER__ where available instead of __LINE__

ned14:
This solves a big problem for me where my existing Boost.Test based unit testing places one test per file and #includes all of them via a build step into the one test binary, and as a result __LINE__ collides repeatedly. This pull request uses __COUNTER__ to generate the unique test case names instead on those compilers which have it, else it falls back to __LINE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.1 build 7 (develop branch)*
+*v1.1 build 8 (develop branch)*
 
 Build status (on Travis CI) [![Build Status](https://travis-ci.org/philsquared/Catch.png)](https://travis-ci.org/philsquared/Catch)
 
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.1 build 8 (develop branch)*
+*v1.1 build 13 (develop branch)*
 
 Build status (on Travis CI) [![Build Status](https://travis-ci.org/philsquared/Catch.png)](https://travis-ci.org/philsquared/Catch)
 
 [Please see this page if you are updating from a version before 1.0](docs/whats-changed.md)
+
+<a href="https://raw.githubusercontent.com/philsquared/Catch/develop/single_include/catch.hpp">[The latest, single header, version can be downloaded directly using this link]</a>
 
 ## What's the Catch?
 
--- a/docs/Readme.md
+++ b/docs/Readme.md
@@ -10,6 +10,7 @@ Before looking at this material be sure to read the [tutorial](tutorial.md)
 * [Build systems](build-systems.md)
 * [Supplying your own main()](own-main.md)
 * [Configuration](configuration.md)
+* [String Conversions](tostring.md)
 * [Why are my tests slow to compile?](slow-compiles.md)
 
 Other
--- a/docs/assertions.md
+++ b/docs/assertions.md
@@ -17,12 +17,12 @@ The ```CHECK``` family are equivalent but execution continues in the same test c
 Evaluates the expression and records the result. If an exception is thrown it is caught, reported, and counted as a failure. These are the macros you will use most of  the time
 
 Examples:
-
-```c++
+```
 CHECK( str == "string value" );
 CHECK( thisReturnsTrue() );
 REQUIRE( i == 42 );
 ```
+
 * **REQUIRE_FALSE(** _expression_ **)** and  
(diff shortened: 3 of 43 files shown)


ned14:
Fixed unit test breakage and merged latest develop. Pull request now passes Travis.

philsquared:

ned14:
No problem. This was a nice clean simple patch with obvious benefits and no bad consequences, at least if CATCH keeps a similar design to now (if you remember I was urging the use of C++ modules to improve compilation times which are for me at least significant, maybe half that of Boost.Test v2, I haven't tested Boost Test v3 yet supposedly it is radically better).

philsquared:
I'm reimplementing this slightly differently - mostly in terms of the compiler detection (so it fits in with the rest of my compiler-capability detection). I also wanted to make sure I'm familiar with every usage - as macros can be tricky if you don't keep your eye on them! I'll credit you in the commit message, though.

As for modules - having now watched Gabriel's cppcon talk I'm somewhat optimistic about them again :-) Catch will need some concessions towards them - particularly in the use of `CATCH_CONFIG_MAIN` and `CATCH_CONFIG_RUNNER`. Are you suggesting the use of `__LINE__` in test names might be a problem with them?

philsquared:
In fact I've pushed it on the develop-v2 branch now (I also have it on master but don't want to push it straight away there until it's been "on the road" a bit).

ned14:
Use of __LINE__ in unit test names was problematic for me because I was keeping my unit tests in separate files but including each of those into a single master source file (this allows me to permute the unit tests via macro magic to increase coverage). I appreciate that is nasty and not usual, but it let me keep Boost.Test compatibility and gave users the option to choose between Boost.Test and CATCH.

Besides, __COUNTER__ is so widely available nowadays and it "just works" that it's not a bad replacement for __LINE__. CATCH's use of counter, unlike almost any other use of counter, is not ODR violating and I think it may be one of the very, very few good idea uses of counter. Next time you see Chandler getting worked up about the travesty of counter and why it's so wrong, make sure to suggest this use case for it :)

Thanks for the merge Phil. Am I going to have to re-hack your develop-v2 branch to support multithreaded use, or is that support planned for soon?

philsquared:
Ah - so it's not modules, per se.

As for the multi-threaded support: the good news is that this is one of the main goals of v2 - and is in active development! I'm not ready for checking in yet as I'm trying to find a decent work-around for lack of proper thread_local support on most compilers!

ned14:
Am I right your fundamental problem with thread local is automatic destruction of TLS contents on thread exit on windows without using dllmain?

philsquared:
Funnily enough no. Or rather I had already anticipated that one and was designing around it.
The problem for me right now is that clang doesn't actually support `thread_local` (well, supposedly it does with a certain version of gcc's stdlib, or something).
Secondarily, although related, it looks like `thread_local` is not well supported beyond that either.

ned14:
Well if you don't care about TLS destruction, proprietary TLS works just fine. Here's a snippet:

```
#ifndef BOOST_AFIO_THREAD_LOCAL
# ifdef __cpp_thread_local
#  define BOOST_AFIO_THREAD_LOCAL thread_local
# elif defined(_MSC_VER)
#  define BOOST_AFIO_THREAD_LOCAL __declspec(thread)
# elif defined(__GNUC__)
#  define BOOST_AFIO_THREAD_LOCAL __thread
# else
#  error Unknown compiler, cannot set BOOST_AFIO_THREAD_LOCAL
# endif
#endif
```
The obvious problem with this is that the type must be an integral type. This leads to the natural workaround:

```
static BOOST_AFIO_THREAD_LOCAL object *local_object;
if(!local_object)
{
  local_object=new object();
  register_with_static_freeer(local_object);
}
```

... which is fine if not too many threads are created and destroyed over time.

An alternative which Boost uses is thread local ptr class which on POSIX uses the destructor callback supplied by pthread_key_create to delete the storage on thread exit. This isn't legally doable on Windows without DllMain tricks.

Now, let me stress the *legally* part. Boost.Thread does a UB evil to implement TLS destruction on thread exit on Windows without requiring a hook into DllMain. If you'd like to know more let me know ...

philsquared:
Sorry - missed your reply before.
Using the proprietary keywords is what I'm currently doing. I'm working around the destruction issue by forcing the user to guard the scope in which threaded testing can occur - so there are well defined points at which the thread-local object gets created and destroyed, rather than it being done lazily.

philsquared:
This is now on master, too

PR #353, opened 2014-11-14
Adds doc on how to make types printable

ryanhaining:
I spent a little while working my way through the source to find the comments about `toString`.  Overloading this function was the only way I was able to print a `std::pair` without reopening `namespace std`.  It's possible I've missed something, I'll be happy to amend it if you'd like to point me in the right direction.  I just created the document I was looking for before having to dive into the code.


ryanhaining:
Yeah that all looks perfect, thanks!

PR #363, opened 2015-01-04
Fix missing brackets in tutorial.md

sadika9:
Add missing closing brackets
  Under "What did we do here?" section


philsquared:
Thanks for that. I Couldn't merge it directly as half of it was already done - so I just redid the other half.

PR #366, opened 2015-01-12
Detect C++11 support for Visual Studio 2015.

wilx:
This patch changes `include/internal/catch_compiler_capabilities.h` so that it detects C++11 features in Visual Studio 2015.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.0 build 53 (master branch)*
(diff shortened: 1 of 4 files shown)


philsquared:
Thanks for that.
I've actually reorganised catch_compiler_capabilities.h - and added the nullptr detection for VS2015 as part of that. The CPP11_OR_GREATER, though, I don't think VS2015 is quite ready for yet (and I have diminished its use in Catch anyway).

PR #372, opened 2015-01-20
(Linux) Support breakpoint upon assertion failure.

milliburn:
Add CATCH_PLATFORM_LINUX (based on presence of `__LINUX__` or `__linux__`).
For this platform,
- define a conditional check for debugger presence by having a temporary child PTRACE_ATTACH to its parent (failure should indicate a tracer (ie. debugger) is already present). Modified version of [this](http://stackoverflow.com/a/3599394/1281937).
- based on that conditional, upon assertion failure
  - execute a Breakpoint Exception (x86) inline; or
  - raise SIGINT (other architectures; the breakpoint will occur somewhere inside raise()).


horenmar:
I merged in #585, so I am closing this.

PR #374, opened 2015-02-11
Add --force-colour option to force colour output.

peterhuene:
Adding a --force-colour option to force colour output on POSIX systems,
provided a debugger is not attached.

This allows for Catch to output colours even if STDOUT is not a tty,
which can be the case when the test executable is being spawned by a
parent process (e.g. CMake's ctest).

### DIFF
--- a/include/internal/catch_commandline.hpp
+++ b/include/internal/catch_commandline.hpp
@@ -170,6 +170,10 @@ namespace Catch {
             .describe( "set a specific seed for random numbers" )
             .bind( &setRngSeed, "'time'|number" );
 
+        cli["--force-colour"]
+            .describe( "force colourised output" )
+            .bind( &ConfigData::forceColour );
+
         return cli;
     }
 
--- a/include/internal/catch_config.hpp
+++ b/include/internal/catch_config.hpp
@@ -37,6 +37,7 @@ namespace Catch {
             noThrow( false ),
             showHelp( false ),
             showInvisibles( false ),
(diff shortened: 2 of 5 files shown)


peterhuene:
This is probably missing tests and I wasn't sure if I needed to regenerate the single include file or not.  I looked at the contributing doc but didn't see guidance around either of those things.  Please let me know what I should do to make this PR acceptable; thanks!

peterhuene:
Looks like Travis is going to fail with a segfault in the tests (presumably doesn't like my `getConfig` call).  I can look into that.

peterhuene:
Also, I'm completely open to using the UK spelling for the option rather than the American one.  Let me know and I'll amend.

peterhuene:
Changed to the UK spellings of "colour" and "colourised" for consistency.

philsquared:
Thanks for that Peter. All looks fine - and pretty much the way I would have done it :-)

kristopherjohnson:
I'd like a --force-no-colour option. When running in CLion, apparently isatty() returns true but the output just has the ugly escape characters.

I don't mind writing the patch myself, but is there a way to force non-colourised output already?

kirbyfan64:
`sed 's/\x1B\[[0-9;]*[JKmsu]//g'` :)

kristopherjohnson:
My --no-colour patch is here: https://github.com/philsquared/Catch/pull/502

PR #381, opened 2015-03-13
Return non-const ref from Catch::Session::cli()

jh3141:
A non-const reference to the `CommandLine` object returned from `Catch::Session::cli()` allows a custom `main` implementation to add additional options to the command line, which cannot be done with a const reference.

### DIFF
--- a/include/catch_runner.hpp
+++ b/include/catch_runner.hpp
@@ -184,7 +184,7 @@ namespace Catch {
             }
         }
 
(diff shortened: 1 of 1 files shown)


jh3141:
(For the motivation behind this, see how I use the forked version here: https://github.com/jh3141/kind-lang/blob/master/tests/main.cpp)

horenmar:
The new CLI in Catch2 is properly composable, so I am going to close this.

PR #390, opened 2015-03-25
Just balance the parentheses around the links.

vadz:
No real changes, just get rid of annoyingly unbalanced parentheses in a couple of links.


vadz:
Oops, I didn't notice the existing #363 which does the same thing (the only difference is that this one removes the IMO accidental pair of parentheses after the first link). Sorry for the duplicate, please feel free to close it and apply #363 instead if you prefer to keep these parentheses.

philsquared:
Thanks for that. I Couldn't merge it directly as half of it was already done - so I just redid the other half.

PR #397, opened 2015-04-12
Add a section about multiple file tests.

Mazyod:
Personally, I chased my tail for a while because this was missing. There are a few raised issues about this, so I really think this will help others get started with the library quicker.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -219,6 +219,14 @@ Scenario: vectors can be sized and resized
       Then: the capacity changes but not the size
 ```
 
+## Spliting Test Cases Across Multiple Files
+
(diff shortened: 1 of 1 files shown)


philsquared:
Thanks for this. You're absolutely right that this is something that is often missed - and something needs to go in the tutorial about it.
I had a slightly different view on how to present it, though - so I've actually written this instead:
https://github.com/philsquared/Catch/blob/master/docs/tutorial.md#scaling-up

Thanks for the prompt.

martinmoene:
@philsquared perhaps: s/real world tests/real-world tests/ ?

philsquared:
yep - good call - thanks

PR #398, opened 2015-04-13
Added Google's support for floating comparison. Refer to http://stackove...

student-t:
Reference: http://stackoverflow.com/questions/17333/most-effective-way-for-float-and-double-comparison

Added support for floating point comparison.

### DIFF
--- a/include/catch.hpp
+++ b/include/catch.hpp
@@ -24,6 +24,7 @@
 
 #include "internal/catch_notimplemented_exception.h"
 #include "internal/catch_context.h"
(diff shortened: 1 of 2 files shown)


martinmoene:
Catch already has support for [floating point comparisons](https://github.com/philsquared/Catch/blob/master/docs/assertions.md#floating-point-comparisons). Why add this one?

cheers,
Martin

student-t:
Sorry, I didn't know it had one. Closed.

PR #401, opened 2015-04-15
Replace tabs with spaces in main include files

stinos:


horenmar:
There is a ton of places that use tabs and tons that use spaces, so there would have to big a large commit that changes them all at once. While this might happen in the future, changing them couple of lines at a time is pointless, so I am going to close this.

Thank you for your effort though.

Issue #410, opened 2015-04-29
REQUIRE_FALSE does not abort test


BMBurstein:
If an assertion marked REQUIRE_FALSE fails, the test case is not aborted

philsquared:
Thanks for reporting this - good catch.
I've merged your PR into master.

PR #411, opened 2015-04-29
Fix REQUIRE_FALSE so that it aborts test case

BMBurstein:
Fix for issue #410


philsquared:
Thanks for that - good catch!
I've merged this back into master.

PR #416, opened 2015-05-16
Added FAIL_AT, MARK_FAILED and MARK_FAILED_AT

rollbear:
I went back and forth between alternatives several times, but in the end I decided on going the macro way instead of exposing a generic report() function. The latter would expose internal data types, making Catch implementation more rigid. This solution is also more in line with what you find in other frame works like google-test and boost unit test.

I hope I've not butchered things too badly. I tried to stay true to your coding style.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -26,6 +26,18 @@ The message is always reported but does not fail the test.
 
 The message is reported and the test case fails.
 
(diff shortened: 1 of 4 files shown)


blastrock:
+1 ! :+1:

philsquared:

PR #419, opened 2015-05-20
Show full path of test run

BMBurstein:
When printing the path of failure, print the path of the test being run (the leaf), not just the section where the error occurred.


BMBurstein:
I accidentally based this off an older revision (I am new to git). The only file I actually changed is include/reporters/catch_reporter_bases.hpp

philsquared:
See response to your forum question: https://groups.google.com/forum/#!topic/catch-forum/4ApDoQ2hXFQ

PR #422, opened 2015-05-24
Fix -Wshadow warnings with GCC.

wilx:
This patch fixes `-Wshadow` warnings that I see when I use Catch in log4cplus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.1 build 3 (master branch)*
+*v1.1 build 4 (master branch)*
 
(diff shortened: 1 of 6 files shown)


wilx:
I guess this is moot at this point. I am closing this.

philsquared:
Sorry, I think I missed this the first time around.
When you say it's moot, do you mean it's no longer relevant, or that you have just given up (or just that the PR would need refreshing)?
I'm trying to catch up on my backlog a bit...

wilx:
@philsquared I think it is no longer relevant. I have since then moved to unpatched version of Catch and I have not had any problems so far.

PR #427, opened 2015-05-27
Fixed "char * const" versus "const char *" confusion

jespert:
There are some confusions between `char const  *` (pointer to a constant char, equivalent to `const char *`) and `char * const` (the pointer is constant, but the char is mutable) throughout the code base. This introduces some unnecessary castings and complicates the manual use of `Catch::Session().run(argc, argv)` when using some very common definitions of `main()` or arrays of string literals. Below are a couple of examples that fail to build with the current master branch.

Example 1:

```
#define CONFIG_CATCH_RUNNER
#include "catch.hpp"

int main(int argc, const char **argv)
{
  Catch::Session().run(argc, argv);
  return 0;
}
```

Example 2:

```
#define CONFIG_CATCH_RUNNER
#include "catch.hpp"

int main()
{
  const char *argv[] = { "my_tests", 0 };
  Catch::Session().run(1, argv);
  return 0;
}
```

NOTE: In this pull request I have already rebuilt the single header with the script. I am not sure if I had to do anything else (or if I should not have done this).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.1 build 3 (master branch)*
+*v1.1 build 4 (master branch)*
 
 Build status (on Travis CI) [![Build Status](https://travis-ci.org/philsquared/Catch.png)](https://travis-ci.org/philsquared/Catch)
 
--- a/docs/own-main.md
+++ b/docs/own-main.md
@@ -16,7 +16,7 @@ If you just need to have code that executes before and/ or after Catch this is t
 #define CATCH_CONFIG_RUNNER
 #include "catch.hpp"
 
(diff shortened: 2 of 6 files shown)


philsquared:
Thanks @jespert.
However please note that ```int main(int argc, const char **argv)``` is not standard c++! (see: http://en.cppreference.com/w/cpp/language/main_function). Some compilers may accept it as an extension.
However your second example is valid - and there's no reason I should be stopping that from working, so I'll change ```Session::run``` to accept ```char const* const*```.

horenmar:
The `Session::run` function has been fixed since and I just fixed up documentation to use standard compliant `int main(int argc, char* argv[])`, so I am going to close this.

PR #429, opened 2015-05-29
defines BDD acronym, and replaces TDD with BDD

luxe:
I think BDD was meant instead of TDD.  I don't see anything about Test-Driven Development further down the documentation.  
Also, the acronym BDD might be obvious to some, but I had to look it up while reading.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -13,7 +13,7 @@ The rest of this tutorial will assume that the Catch single-include header (or t
 
 # Writing tests
 
-Let's start with a really simple example. Say you have written a function to calculate factorials and now you want to test it (let's leave aside TDD for now). 
+Let's start with a really simple example. Say you have written a function to calculate factorials and now you want to test it (let's leave aside BDD for now). 
 
 ```c++
 unsigned int Factorial( unsigned int number ) {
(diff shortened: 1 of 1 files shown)


kirbyfan64:
Aren't they still distinct?

itrion:
I agree that acronyms should be used after the introduction of its full meaning, because no all the people have to know it. BDD and TDD are not the same thing. BDD is for an outside in, top down approach and TDD is for inside out or bottom up. You can think in behavior-driven development as a kind of evolution of test-driven development but at least for me not all test are suitable to be written in terms of scenarios so it is

Catch offers the two approaches and The good part is that both of them can be use in any case. In the example of the factorial, it is a unit test case, and not an scenario for that reason I think the name should be kept.

philsquared:
Ok. A few things going on here.
 Thanks, @luxe, for bringing this up.
However TDD was the intended term in the first mention. The meaning being, regardless of how you normally write your tests (before or after the code being tested): here we're going to write the code first.

I agree that I shouldn't just drop an acronym like TDD in there with no explanation, introduction, or even an expansion. I'll think about rewording it. However I intend to write a bit more about TDD in general and could just link to that (perhaps, for now, I'll just link to wikipedia or something).

Later I mention BDD - but here I'm talking about something quite different. Again: guilty as charged - I didn't offer any explanation for the term. Perhaps a simple link, for now, will suffice - but really that should be expanded on a bit more.

Specifically I say, "BDD-style" to refer to the give-when-then structure. given-when-then is not unique to BDD, nor does it define BDD. It is, however, often associated with BDD - and BDD is seen as popularising it - so I'll continue to say BDD-Style - but should offer a little more qualification

@itrion: BDD was originally meant to be "TDD done well" - with a little extra rigour and focus. BDD, itself, has evolved to emphasis the human communication more - with the impact on testing and design more of an artifact of the process. As such I don't think it's out of place to describe a way of doing TDD that emphasises a bit of structure on the way what is going on is communicated to be "BDD style". But we're getting a bit deep :-)

In short I won't take this PR as-is, but I'll try and clarify the wording a little. Thanks guys!

luxe:
Thanks.

PR #436, opened 2015-06-10
Improve error messages

timonbaetz:
```C++
class Error {
public:
  operator bool() const { return false; }
};

std::ostream& operator << ( std::ostream& os, const Error& value ) { 
  os << "some error string"; return os; 
}

Error getError() { return Error(); }

TEST_CASE("Some test", "[SomeTest]") {
    REQUIRE(getError());
}
```
Without the change outputs:
```
... FAILED:
  REQUIRE( getError() )
with expansion:
  false
```
With the change outputs:
```
... FAILED:
  REQUIRE( getError() )
with expansion:
  some error string
```

### DIFF
--- a/include/internal/catch_expression_lhs.hpp
+++ b/include/internal/catch_expression_lhs.hpp
@@ -71,7 +71,7 @@ public:
     void endExpression() {
         bool value = m_lhs ? true : false;
         m_rb
-            .setLhs( Catch::toString( value ) )
+            .setLhs( Catch::toString( m_lhs ) )
             .setResultType( value )
             .endExpression();
     }


philsquared:

philsquared:
I've had a look at this now and I'm not convinced this is the right way to go.
I'll try and explain my reasons:

1. What the assertion is *testing* is the boolean value that the object converts to. When Catch prints the *actual* value it should be the value that is actually being tested. That's not to say that the error string is not useful in this case - although I'd argue that that interpretation is specific to your usage. But the message is extra context, rather than the value being tested (I'll show you how you can have your cake and eat it in a moment).

2. By taking your approach it breaks for certain other things that convert to bool - for example objects using the "safe bool idiom" (which typically return a pointer to a private member function). These now get printed as integers - which is not quite as friendly - even though we evaluated them as bools.

I promised you your cake - and you can eat it by capturing the context using `CAPTURE` - e.g.:

```
    auto err = getError();
    CAPTURE( err );
    REQUIRE( err );
```

For me, with you example, that prints:
```
  REQUIRE( err )
with expansion:
  false
with message:
  err := some error string
```

How does that sound?

timonbaetz:
Yes, I get your point and thx for the detailed answer. I am using that approach to extract validation functionality:
```
Error isValid(const Foo& foo);
...
REQUIRE(isValid(foo1));
REQUIRE(isValid(foo2));
```
But i guess that should work with ```CAPTURE``` as well

oh, and thx for great work.

PR #448, opened 2015-06-24
Fixed typo in docs

Sumith1896:
I don't know if this is mistake in docs or intentional. 
If not a mistake, feel free to ignore.
cc @philsquared


philsquared:
Thanks @Sumith1896,

It is intentionally that way. The square brackets are to indicate that the second argument is completely optional - so the comma should be inside them (as you won't include the comma if not using a second argument.
It's a little confusing because the second argument is the tags, which is a string containing identifiers within square brackets - but they are different.

Sumith1896:
Cool. Thanks for building Catch, we moved to Catch and we are very happy :smile:

PR #456, opened 2015-07-03
Fix for white background in windows powershell when using colour output.

RandomInEqualities:
In powershell the text background appears white

bad_colours

In the current implementation the original background is overwritten when coloring text. This commit splits the foreground and background color attributes. I have tested it with both power-shell and command line.


philsquared:
Thanks for this. I've been meaning to look into it for a while.
I think it's still not a complete solution. If the background colour is not black then the current foreground colours may not be readable. Still it's better than the previous situation - especially for the Powershell case!

PR #457, opened 2015-07-04
Fixed code analysis warning C28251.

tn0502:
Fix for C28251 "Inconsistent annotation for function: this instance has an error":

Inconsistent annotation for `OutputDebugStringA`: this instance has no annotations. See `c:\program files (x86)\windows kits\8.1\include\um\debugapi.h(85)`. `catch_debugger.hpp`	85

### DIFF
--- a/include/internal/catch_debugger.hpp
+++ b/include/internal/catch_debugger.hpp
@@ -82,7 +82,7 @@
 #endif // Platform
 
 #ifdef CATCH_PLATFORM_WINDOWS
-    extern "C" __declspec(dllimport) void __stdcall OutputDebugStringA( const char* );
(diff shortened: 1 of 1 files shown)


philsquared:

tn0502:
No problem. I can rebase and create another request if you prefer?

philsquared:
Sure - seems like that should be easy for you to do. Thanks.
My only concern with this (without looking into it too deeply) is whether this would introduce portability issues, since it relies on an MS extension. Although it's protected by a Windows `#define`, there are non MS compilers that target Windows.
But those compilers must also consume Windows.h, so either they must support those extensions, or there must be conditional code in there too.

Issue #464, opened 2015-07-17
Catch_xmlwriter generates invalid xml if we CAPTURE an uint8_t that has a control character value


espenalb:
Steps to reproduce:

```
#include "catch.hpp"

TEST_CASE("this test fails intentionally in order to get the CAPTURE into the log") {
  uint8_t a = 1;
  CAPTURE(a);
  CHECK(a == 2);
}
```

Running the test fails as expected, but if we generate a JUnit report, the xml file contains invalid characters, causing my CI server (Jenkins) to reject the entire XML file.

philsquared:
Thanks for reporting this.
I took a look at the xml encoding code and realised it was overdue some attention - so I rewrote it - taking on board your control characters proposal (I actually encode them as entities, though).
Please let me know of this addresses your issue (dev build 9)

horenmar:
There have been major changes in XML writer recently that fixed a lot of things, so this is likely resolved.

I am going to close this soon if there is no feedback that this still doesn't work.

philsquared:
if I test @espenalb's original example myself with `-r junit` (or `-r xml`) I get:

`a := &#x01;`

- which I presume is what is expected?

philsquared:
A little wrinkle here - this is valid XML 1.1 (I believe). but not 1.0. We've been using XML 1.1 for a while now (for this reason) but that turns out to be problematic as it is not widely supported.
So I'm looking at reverting to XML 1.0 and instead of encoding these chars writing them as C-escaped sequences (e.g. \x01).
Any thoughts?
(See https://github.com/philsquared/Catch/tree/Xml1_0)

philsquared:
The Xml 1.0 changes went in in 1.8.0, so I think that concludes this issue.
Feel free to re-open if you think there is still an issue?

PR #465, opened 2015-07-17
Proposal for fixing #464 - invalid XML in output.

espenalb:
This fixes the issue with invalid characters in the XML output.

I guess it could also be fixed by overloading the toString function - that is maybe a better solution, as this quickfix has a performance cost since _all_ strings are parsed...

### DIFF
--- a/include/internal/catch_xmlwriter.hpp
+++ b/include/internal/catch_xmlwriter.hpp
@@ -169,9 +169,26 @@ namespace Catch {
             }
         }
 
-        void writeEncodedText( std::string const& text ) {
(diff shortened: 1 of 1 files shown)


philsquared:
Sorry - looks like I referenced this issue in a commit a while back but never followed up to make sure it was now addressed. Is this still outstanding or is it covered by that commit? (IIRC I did it differently as I used a more integrated approach that, I believe mitigates the extra performance overhead, amongst other things).

horenmar:
In the year and a half since this was open, XML handling has been completely overhauled and there is no test case here to recheck.

If this is still problem, please open a new issue/PR.

PR #466, opened 2015-07-18
Add matcher: Equals(std::vector)

webmaster128:
This matcher supports arbitrary `std::vector`s including custom allocators. It [is tested](https://github.com/randombit/botan/blob/master/src/tests/catchy/test_base.cpp) using [Botan](http://botan.randombit.net/)'s secure_vector type.

Note that this moves `catch_tostring.h` up in the single header, I hope this causes no trouble.

### DIFF
--- a/include/internal/catch_matchers.hpp
+++ b/include/internal/catch_matchers.hpp
@@ -8,6 +8,9 @@
 #ifndef TWOBLUECUBES_CATCH_MATCHERS_HPP_INCLUDED
 #define TWOBLUECUBES_CATCH_MATCHERS_HPP_INCLUDED
 
+#include <vector>
+#include "catch_tostring.h"
+
 namespace Catch {
 namespace Matchers {
     namespace Impl {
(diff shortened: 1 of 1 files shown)


webmaster128:
I just noticed that this moves `catch_tostring.h` up in the single header, but std::to_string was somehow not available. Maybe because it is C++11.

philsquared:

webmaster128:
Hi @philsquared, could you please have a look at this again? This Equals matcher for std::vector was a non-trivial piece of work it would be unfortunate if it get's lost now. I don't know what your plans for matchers are and if you like this implementation but I you want matchers, I think this is a real basic one and a great example for a matcher that can handle template types.

philsquared:
I'll definitely be looking at it. It's exactly the sort of thing I've been meaning to get around to adding for a while - so the work is very much appreciated.
I'm just trying to work through my backlog at the moment!
I should be able to cherry pick from your branch straight into master - so no need to retarget.

webmaster128:
Great, no need to hurry. I was just afraid it was overseen.

philsquared:
Sorry I never got back to this. I must have forgotten to make a note of it and- since it's closed with no way to reopen - it didn't even come up in review.

Anyway, if this is still an area of interest for you you might like to know that I've been reworking the Matchers implementation. It's now much simpler - with far less templates in particular (and no CRTP!).

That's the good news. The (slightly) bad news is that it will break existing custom Matchers (although its not a big job to update them). Since you've been using them in the past I thought I'd give you a heads up before this gets merged to master.

You can find them here:
https://github.com/philsquared/Catch/tree/new-matchers

(or just the single include:
https://github.com/philsquared/Catch/raw/new-matchers/single_include/catch.hpp)

Please let me know what your thoughts or - including if this is no longer relevant for you so I'll not wait for you.

webmaster128:
Thanks for the update. Even though I always loved Catch, I don't work on a project using it right now. Thus I cannot put any efford in rewriting this matcher. Please feel free to reuse whatever you find here if it helps.

philsquared:
Thanks for getting back. No worries at all. My main concern was whether I'd be breaking you.

PR #470, opened 2015-07-24
Detect nullptr support in GCC 5 better

DavidEGrayson:
This change allows Catch to work (or at least compile) on GCC 5.2 when it is not running in C++11 mode (no standard specified on the command line).  I am using i686-w64-mingw32 gcc 5.2.0 built by the MSYS2 project.

My understanding is that nullptr was introduced in C++11, so it seems like we should check to see whether we are running C++11 mode and only attempt to use nullptr if that is the case.  Every GCC since 4.7 has a __cplusplus macro we can check to see what version of the C++ standard is being used.  This change preserves the logic that was used for GCC 4, but for GCC versions greater than 4, it checks the __cplusplus macro to see if nullptr is supported.

I assume there isn't much harm in false negative with this macro.  Since my change will strictly decrease the set of compilers where Catch attempts to use nullptr, I think it should be pretty safe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![catch logo](catch-logo-small.png)
 
-*v1.2.1-develop.6*
+*v1.2.1-develop.7*
 
 Build status (on Travis CI) [![Build Status](https://travis-ci.org/philsquared/Catch.png)](https://travis-ci.org/philsquared/Catch)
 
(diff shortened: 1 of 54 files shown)


DavidEGrayson:
Sorry, GitHub is displaying a ton of commits, but I only made one.  This pull request is based on your `develop` branch, not master, and I should have specified that when making the pull request.

PR #471, opened 2015-07-24
Fix for nullptr in GCC 5.2

DavidEGrayson:
This change allows Catch to work (or at least compile) on GCC 5.2 when it is not running in C++11 mode (no standard specified on the command line). I am using i686-w64-mingw32 gcc 5.2.0 built by the MSYS2 project.

My understanding is that nullptr was introduced in C++11, so it seems like we should check to see whether we are running C++11 mode and only attempt to use nullptr if that is the case. Every GCC since 4.7 has a __cplusplus macro we can check to see what version of the C++ standard is being used. This change preserves the logic that was used for GCC 4, but for GCC versions greater than 4, it checks the __cplusplus macro to see if nullptr is supported.

I assume there isn't much harm in false negatives with this macro. Since my change will strictly decrease the set of environments where Catch attempts to use nullptr, I think it should be pretty safe.

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -67,7 +67,7 @@
 // GCC
 #ifdef __GNUC__
 
-#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6 && defined(__GXX_EXPERIMENTAL_CXX0X__) )
+#if (__GNUC__ > 4 && __cplusplus >= 201103L) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6 && defined(__GXX_EXPERIMENTAL_CXX0X__) )
 #   define CATCH_INTERNAL_CONFIG_CPP11_NULLPTR
 #endif
 


DavidEGrayson:
This [snippet of code](
https://github.com/gcc-mirror/gcc/blob/d9b3844711525446cb92c9ef85c7b7e7d3dd05b2/libstdc%2B%2B-v3/include/bits/c%2B%2Bconfig#L201-203) from GCC confirms that my pull request does the right thing:

~~~~
#if __cplusplus >= 201103L
  typedef decltype(nullptr)	nullptr_t;
#endif
~~~~

philsquared:
Thanks for this David,

Actually a bit further down there's a big `#if (__cplusplus >= 201103L)` check for all compilers, so I've changed the gcc check to be specific to that 4.6 edge case and let the more recent versions drop down to the more general condition check.
I believe this will have the same results (I've not pushed that change yet - but you'll get a notification when it is)

philsquared:
Are you able to confirm if my changes have the same effect as your PR?

DavidEGrayson:
Yep, your changes look good, and I just confirmed with testing that the latest version from https://raw.githubusercontent.com/philsquared/Catch/develop/single_include/catch.hpp works fine with gcc 5.2.0 in the default language mode.

PR #482, opened 2015-08-12
Add matcher: Not()

webmaster128:
A generic Not() matcher to get nice readable test code a la

```
CHECK_THAT(address1, Not(Equals(address2)));
```

[Used and tested in Botan.](https://github.com/randombit/botan/blob/master/src/tests/catchy/test_x509.cpp)

### DIFF
--- a/include/internal/catch_matchers.hpp
+++ b/include/internal/catch_matchers.hpp
@@ -32,6 +32,23 @@ namespace Matchers {
     };
 
     namespace Generic {
+        template<typename ExpressionT>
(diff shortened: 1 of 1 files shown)


philsquared:
Thanks. Been meaning to do that for a while!

webmaster128:
Fine, thanks. I just rebased my other matcher onto current `develop`: #466 When this is in, I can follow-up with a bunch of matchers I have in the pipeline, like Integer and Boolean matching.

PR #488, opened 2015-08-20
Fix for Coverity analysis issues

Tarnasa:
Running a Coverity scan on this project yielded two possible issues which I have resolved:
- Big parameters passed by value
- Uninitialized scalar fields

Also please note,
The signature of one of the methods (TestRegistry::LexSort) has changed, if this method represents a part of Catch's public API, then we should consider incrementing the major version as per semantic versioning.

### DIFF
--- a/include/internal/catch_assertionresult.h
+++ b/include/internal/catch_assertionresult.h
@@ -15,7 +15,7 @@ namespace Catch {
 
     struct AssertionInfo
     {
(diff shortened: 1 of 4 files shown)


michaelbprice:
@Tarnasa I just noticed that this pull request is to the `master` branch.  It might need to be made to the `develop` branch.

PR #489, opened 2015-08-21
add cmake support for easy, crossplatform and standart install

masterspline:
Cmake support will simplify installation scripts. 

And it's crossplatform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,5 @@
+project(Catch)
+cmake_minimum_required(VERSION 2.8)
+
+install( FILES single_include/catch.hpp
+	DESTINATION include )


kirbyfan64:
...

Isn't this kind of overkill, considering that CMake is a full build system and the equivalent Posix SH install script is 1 line long?

```sh
cp single_include/catch.hpp ${INCDIR:-/usr/local/include}`
```

Again, Posix shell. Works on pretty much every Unix platform. Permissions don't really matter here, hence the use of `cp`. CMake doesn't generate install targets for Windows anyway!

masterspline:
> equivalent Posix SH install script is 1 line long

+ ShaBang line + 1 line .bat file -> same 3 lines but in in 2 different files

>  CMake doesn't generate install targets for Windows anyway!

My cmake generates INSTALL.vcxproj for MsBuild and really do install.

If You wish to use shell - just do it, CMakeLists.txt do not exclude it. But it provides many small advantages:

1. Common and standart build and install steps for all projects using cmake (only configure step requires to know custom options for different projects, but build, install and test are the same). If You need to create packege or maintain it, CMakeLists.txt makes you work easy.

2. It's really crossplatform and not limited by Posix Shell.

3. CMakeLists.txt greatly simplifies external dependancy installation for C++ project using cmake ExternalProject (without it I need to write patch or custom installation script and do it crossplatform because world did not limited by Posix systems).

The only cons here is that Upstream need to support CMakeLists.txt (currently requires about zero resources).

Darelbi:
That script actually doesn't build Catch, because it needs to define Catch as source file before including it. You are just creating an empty library file. If compiled correctly Catch is 600 kb (compiled with GCC -O3)

masterspline:
> That script actually doesn't build Catch

Yes, It's just install (copy) ready catch.hpp file. Crossplatform install is all I want.

Darelbi:
Sorry, It's a period I write before reading and then I have to edit my messages XD.

masterspline:
> Installing an header is not so good, in general the installation process with Cmake is flawed. It can't replace a operative system package management

I do not suggest to do `make install` from root. CMakeLists.txt is just helper to create package or install files in user directory. Crossplatform helper. And it makes installation simple and common for all projects using cmake.

>  It is better to include a cloned source tree directly from cmake so that you can build your project and all dependencies.

If You about cloning Catch in separate directory and setting include_directories(), then it is common and known way (even in such solution CMakeLists.txt will help).

Currently I use better solution. I use cmake ExternalProject to install external dependancies (which not found by find_package()). All packages installed in same directory, so include_directories() and link_directories() called once. CMakeLists.txt greatly simplify ExternalProject usage. Makes it common for all packages.

Darelbi:
I mean compile other targets as subtargets (add_subdirectory) and having them in same folder (similiar to you, except I don't use ExternalProject but simply use add_subdirectory (and includes and linking of course), it does not assume client have installed cvs programs or internet connection active. Instead of patching a dependency you download a the new version and unzip it beside old one in a parallel folder with different version number, if the new version breaks something it's easier to rollback(you just set directory to point again to old library, you don't have to use a cvs command to "unpatch" the external library). This is of course useful in a Continuos integration environment where you have continuos builds and tests of all components and an installer is not an option because you could as well have 3 versions of a library for 3 different projects (deploy is still done, a zip file is generated with each build, of course not every build get released ^^ )

PR #490, opened 2015-08-28
Enable tuple printing for MSVC

studoot:
Microsoft C++ in VS2013 & above can support tuple printing as performed
by Catch (requires std::tuple and variadic templates), so add enabling
macro definition for those compilers.


philsquared:

PR #497, opened 2015-09-16
Switch to container-based and enable OS X on Travis

robertodr:


philsquared:
Thanks for this,
However I've now accepted https://github.com/philsquared/Catch/pull/512, which seems to cover the same ground and more (let me know if you think I've missed something, though).

robertodr:
PR #512 is way more comprehensive than mine ;) Thanks for your great work. Your framework really rocks!

PR #498, opened 2015-09-16
Allow overriding std::exception translation

mikedld:
CATCH_TRANSLATE_EXCEPTION allows to only translate exceptions which are
not inherited from std::exception (as well as std::string and char*).
There're exception classes though which inherit from std::exception but
don't provide any meaningful info as what() result, Poco::Exception being
one of those. In this case, adding translator like
```
    CATCH_TRANSLATE_EXCEPTION(Poco::Exception const& e)
    {
        return e.displayText();
    }
```
might help, but unfortunately it's not functioning as desired.

This commit makes it possible to provide translation overrides for
std::exception-derived classes.

### DIFF
--- a/include/internal/catch_exception_translator_registry.hpp
+++ b/include/internal/catch_exception_translator_registry.hpp
@@ -45,7 +45,17 @@ namespace Catch {
                 throw;
             }
             catch( std::exception& ex ) {
-                return ex.what();
+                try {
+                    try {
+                        throw;
+                    }
+                    catch(...) {
+                        return tryTranslators( m_translators.begin() );
+                    }
+                }
+                catch( std::exception& ex ) {
+                    return ex.what();
+                }
             }
             catch( std::string& msg ) {
                 return msg;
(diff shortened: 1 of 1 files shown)


mikedld:
Sorry, was the closing a silent "no"?..

nabijaczleweli:
Looks like @philsquared just mass-closed 16 issues

philsquared:
I deleted the develop branch (which is now fully merged into master).
It hadn't occurred to me that would automatically close all PRs against the branch. Makes sense, though.

I'm going to have to go back over them now and see which need reopening - probably not a bad thing as I'm behind anyway.

In the case of this one, though, I believe I committed a changed fairly recently that has the same effect. Sorry this PR slipped past my radar at the time or I would have used it/ acknowledged it.

I'd appreciate if you could check out the latest master and see if it works for you translating Poco exceptions (etc).

mikedld:
Indeed it looks like a49f088 does what I need. I'll delete the branch once I confirm master works here. Thanks!

PR #501, opened 2015-09-29
Use original background colour in windows console (instead of black).

SergiusTheBest:
Hi!

My default console background is blue and test output looks not very good. I've changed the code to use original background.

before

after

### DIFF
--- a/include/internal/catch_console_colour_impl.hpp
+++ b/include/internal/catch_console_colour_impl.hpp
@@ -60,35 +60,37 @@ namespace {
         {
             CONSOLE_SCREEN_BUFFER_INFO csbiInfo;
             GetConsoleScreenBufferInfo( stdoutHandle, &csbiInfo );
-            originalAttributes = csbiInfo.wAttributes;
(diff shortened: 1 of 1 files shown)


Kosta-Github:
what happens on a green background? :-)

Darelbi:
true, should keep black background or otherwise catch have to detect background color and choose a adifferent color for fonts not very good.

SergiusTheBest:
@Kosta-Github I guess a green background is not common.

philsquared:
Thanks for this.
However I've actually already made essentially the same changes on the Develop branch. See:
https://github.com/philsquared/Catch/blob/develop/include/internal/catch_console_colour_impl.hpp

@Kosta-Github's observation is important, though. I've considered providing a few colour sets, mapped by background colour, but not got around to it yet.

SergiusTheBest:
@philsquared Good to know. Thanks!

Kosta-Github:
@SergiusTheBest IMHO, a `green` background is as likely or unlikely as a `blue` one...

PR #502, opened 2015-09-29
Add --no-colour command-line option

kristopherjohnson:
Adds a --no-colour command-line option that will disable colourised output even if the output device/stream supports it.


rcdailey:
You misspelled "color" :)

nabijaczleweli:
It's "colour", you filthy colonial!

But "color" is a better option here, because everyone uses it now. *audible sigh*

philsquared:

nathanaeljones:
This flag would be extremely useful, as without it Catch and CLion do not get along well at all...

Issue #304, opened 2014-08-18
Dead link in tutorial.md


d-meiser:
The link pointing to a blog post on how to use catch in open source software projects (section "Where to put it?") is broken.

PR #504, opened 2015-09-30
Fix whitespace errors in include/**

xaizek:
Tried to add Catch to my repository and hit a lot of warnings from git about whitespace errors (the check is performed by pre-commit hook), this is result of fixing those warnings.


include/reporters/catch_reporter_console.hpp
```diff
@@ -258,11 +258,14 @@ namespace Catch {
             }
         }
         void lazyPrintRunInfo() {
+            // Usefulness of this message is questionable.
```
michaelbprice: These changes certainly seem outside the scope of whitespace changes.  What's the rationale for these?

michaelbprice:
I detest unnecessary whitespace like everyone else, so I like the effort.

I must say that I dislike Git's propensity to care about whitespace, and I consider that a tooling problem (in this case it seems like you have some overly aggressive pre-commit hooks).

include/reporters/catch_reporter_console.hpp
xaizek: Thanks for noticing this! I forget disabling it (it's annoying to see whole message over and over again if you do `make check` several times in a row).

xaizek:
> I must say that I dislike Git's propensity to care about whitespace, and I consider that a tooling problem (in this case it seems like you have some overly aggressive pre-commit hooks).

It's great actually, that's the way I setup my pre-commit hook to make sure unnecessary whitespace is never committed. It's better to make sure no changes add extra whitespace than remove it later, polluting repository history for no reason.

jubalh:
There is also #310 which has the same goal but got opened some time earlier. Today I updated the PR so it can get automerged.

xaizek:
Hey Michael, that one has 8 files in `include/` versus 16 here, maybe those are new files. You might want to squash this commit into #304.

philsquared:
Thanks for this - however I've now added a script to do the job. More info in my comment to PR #304

PR #512, opened 2015-10-06
Improves the Travis CI integration

daniel-j-h:
I read about https://github.com/philsquared/Catch/issues/511#issuecomment-145700281 via [your twitter](https://twitter.com/phil_nash/status/651284078138904576), and though instead of writing something, let actions speak. So here are some Travis CI improvements.

-----------------------------------------------------------------

In particular, this adds several improvements, such as:

- making use of container based infrastructure (faster startups, caching feature, 2 cpus, 4gb ram, ..)
- using ccache for faster building (we could also cache arbitrary directories)
- builds for linux and osx
- builds for various gcc, clang versions
- recent cmake installation

The Travis CI status for my fork looks like this, if you want to give it a try:
https://travis-ci.org/daniel-j-h/Catch

Related: what about [this C++11 check](https://github.com/philsquared/Catch/blob/0c1c9fa9229c7358ce16c9910004a39290b20f5f/projects/CMake/CMakeLists.txt#L9-L13)? The installed compilers can handle C++11.

-----------------------------------------------------------------

Further improvements: there are some features still open for improvements, for example we could do builds against libc++, run sanitizers, and so on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,163 @@
 language: cpp
+sudo: false
 
-compiler:
-  - clang
-  - gcc
+cache:
+  ccache: true
+  directories:
+    - $HOME/.ccache
 
 env:
-  - BUILD_TYPE=Debug
-  - BUILD_TYPE=Release
+  global:
+    - USE_CCACHE=1
+    - CCACHE_COMPRESS=1
+    - CCACHE_MAXSIZE=200M
+    - CCACHE_CPP2=1
+
+
+matrix:
+  include:
+
+    # 1/ Linux Clang Builds
+    - os: linux
+      compiler: clang
+      addons: &clang35
+        apt:
+          sources: ['llvm-toolchain-precise-3.5', 'ubuntu-toolchain-r-test']
+          packages: ['clang-3.5']
+      env: COMPILER='ccache clang++-3.5' BUILD_TYPE='Release'
+
+    - os: linux
+      compiler: clang
+      addons: *clang35
+      env: COMPILER='ccache clang++-3.5' BUILD_TYPE='Debug'
+
+
+    - os: linux
+      compiler: clang
+      addons: &clang36
+        apt:
+          sources: ['llvm-toolchain-precise-3.6', 'ubuntu-toolchain-r-test']
+          packages: ['clang-3.6']
+      env: COMPILER='ccache clang++-3.6' BUILD_TYPE='Release'
+
+    - os: linux
+      compiler: clang
+      addons: *clang36
+      env: COMPILER='ccache clang++-3.6' BUILD_TYPE='Debug'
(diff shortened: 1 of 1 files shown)


kirbyfan64:
So, just out of curiosity, what's the difference between `&clang36` and `*clang36`?

daniel-j-h:
If you read the [YAML Spec](http://yaml.org/spec/1.2/spec.pdf) (warning: 84 pages pdf) this is the so called anchoring feature. With `&clang36` we create an alias that we then can use by means of `*clang36`.

philsquared:
Thanks for this!
I didn't realise you can test against multiple versions of gcc/ clang etc on Travis. How far back does gcc, in particular, go?
In any case this would be a a great step forward. I know there was another PR relating to container-based Travis - was that from you? I hadn't had a chance to look at it.
I'd still like a build system with more control - so we can run those Python scripts too (unless you can do that in Travis?)

Darelbi:
take a look at one of my travis scripts:
https://github.com/Darelbi/Infectorpp2/blob/master/.travis.yml

basically travis is a unix shell environment, it should already have python for running scripts, if not the case, you just add the apt package into travis configuration.

I'm using only C++11 so I have tested against few compilers but you can both test against much more compilers and for the same compiler you can also test with `-std=c++11` enabled or not.

daniel-j-h:
@philsquared:

> I didn't realise you can test against multiple versions of gcc/ clang etc on Travis. How far back does gcc, in particular, go?

Take a look at the ppa `ubuntu-toolchain-r` that is used to pull in gcc (and stdlibc++, ...):
https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

From what I can tell the oldest is gcc-4.5 _in this particular ppa_! There is nothing stopping us from pulling packages from different ppas.

> In any case this would be a a great step forward. I know there was another PR relating to container-based Travis - was that from you? I hadn't had a chance to look at it.

If you mean https://github.com/philsquared/Catch/pull/497/files, no that's not by me. In contrast to this pull request, this other one does not include various compiler versions, caching, a recent cmake and so on.

> I'd still like a build system with more control - so we can run those Python scripts too (unless you can do that in Travis?)

Sure. It's a full blown Linux (in fact a Ubuntu 12.04, that's why we pull recent packages from ppas). There are whitelists for [sources](https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json) and [packages](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise), though. But usually these packages are more than enough.

philsquared:
Thanks again for this. I'm definitely going to have to look into this more - but I don't see any reason not to start with this right away.
Next steps are getting the Python scripts to run...

PR #520, opened 2015-10-20
Add a pragma to suppress parenthesis warnings for GCC

mattgodbolt:
Fixes issues with GCC where errors of the form `error: suggest parentheses around comparison in operand of ‘==’ [-Werror=parentheses]`. Seems to only happen with GCC 4.7, for whatever bizarre reason.

Should address issue #496 without the nasty side effects (and reversion).

e.g. fixes errors seen in: https://travis-ci.org/mattgodbolt/seasocks/jobs/86310974 (now https://travis-ci.org/mattgodbolt/seasocks/jobs/86313317)


mattgodbolt:
argh, apologies. I will resubmit with a patch against the develop branch

PR #527, opened 2015-10-27
restrict nullptr detection to gcc 4 (closes #526)

### LINKED ISSUE #526: Catch, nullptr and gcc 5.0: availability of std::nullptr_t misdetected w/C++98
Note: this is similar to https://github.com/philsquared/Catch/issues/487, but I think something could be done in Catch to help alleviate this.

It looks like `gcc` does not define `std::nullptr_t` when the C++ standard (through `-std`) is not specified. A simple workaround is to ensure that you compile with `std=c++11` or above, but some of us unfortunately live in a world where we must adhere to the C++98 standard.

I think the lines [here](https://github.com/philsquared/Catch/blob/d758428fe297655e558a86cc6a013fcb1bcc8c7d/include/internal/catch_compiler_capabilities.h#L69-L71) need to be amended to only operate with GCC with major version 4. With gcc 5 and above, one must explicitly opt into the C++11 standard if they want `std::nullptr_t`; that would then be detected [here](https://github.com/philsquared/Catch/blob/d758428fe297655e558a86cc6a013fcb1bcc8c7d/include/internal/catch_compiler_capabilities.h#L109-L111).

(and PS: thanks to everyone who has contributed to Catch; it's an amazing tool!)

kevinushey:
The old detection logic here assumed that all versions of gcc >= 5 would
provide nullptr and `std::nullptr_t`; this is only true in gcc >= 5 when
the standard is explicitly set (e.g. `std=c++11`), and other detection
logic handles that.

This change allows Catch-using programs to compile with gcc >= 5, when
using the default (C++98) standard.

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -66,7 +66,7 @@
 // GCC
 #ifdef __GNUC__
 
-#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6 && defined(__GXX_EXPERIMENTAL_CXX0X__) )
+#if __GNUC__ == 4 && __GNUC_MINOR__ >= 6 && defined(__GXX_EXPERIMENTAL_CXX0X__)
 #   define CATCH_INTERNAL_CONFIG_CPP11_NULLPTR
 #endif
 


kevinushey:
And I now see that this was submitted here: https://github.com/philsquared/Catch/pull/471

Sorry for the noise!

PR #530, opened 2015-10-28
Get rid of unused parameter in TestEventListenerBase::assertionEnded()

vadz:
Just don't name this parameter to avoid warnings about it being unused.

No real changes.


nabijaczleweli:
:+1:

michaelbprice:
Nice and simple. :+1:

philsquared:
Thanks for catching that.
I couldn't merge the PR as it conflicts with the single_include header (in general please don't update the header in single_include as it is generated by splicing all the other headers together).
So I've just reapplied the same fix - will make it in to the develop branch soon,.

PR #534, opened 2015-11-06
 don't forward declare as class and later define as struct

snaewe:
Signed-off-by: Stefan Naewe <stefan.naewe@gmail.com>


philsquared:
Thanks.
I also spotted this when I did a Visual Studio build - but I've made the change the other way (the `Not` struct is now a class).
So I'm going to close this PR but will mention it in the commit message.

PR #541, opened 2015-11-18
Fixed Catch download URL in Tutorial

tnull:
The site http://builds.catch-lib.net seems to be down or _very_ slow. The link (https://raw.githubusercontent.com/philsquared/Catch/master/single_include/catch.hpp) provided in README.md is working, tough.


horenmar:
This change has already been made at some point in the past, so I am closing this.

Issue #543, opened 2015-12-03
 fabs() vs std::fabs() causing build errors


Crog:
Updating from v1.1 build 3 to 1.2.1 results in numerous compilation errors. This appears to be due to a mixture of C & C++ functions i.e such as fabs() vs std::fabs() used in v1.1. The relevant C header should be included.

```
/home/smex/NGL/DEV/5000core/../3rd_party/catch/catch.hpp: In function 'bool     Catch::Detail::operator==(double, const Catch::Detail::Approx&)':
/home/smex/NGL/DEV/5000core/../3rd_party/catch/catch.hpp:2140: error: 'fabs' was not declared in this scope
cc: /opt/qnx650/host/linux/x86/usr/lib/gcc/arm-unknown-nto-qnx6.5.0/4.4.2/cc1plus error 1
```

The issue was produced under GCC 4.4.2 with an empty .CPP file that has a single 
 line to #include <catch/catch.hpp>

The errors are all due to the use of the 'C'  global namespace functions instead of the C++ std:: namespace functions such as fabs()

In many places, the C functions are okay to use as there is an inclusion of <cstring> which brings in a large amount of C including malloc() and the likes.

Crog:
To embellish the details this is using the QNX qcc compiler packaged with QNX v6.5.0. The standard headers do appear to adhere to standard namespace conventions quite well unlike some other libraries where it appears the C 'global' functions are simply pulled into the std:: namespace. Seeing regression since Catch v1.1 it makes you wonder if it were tested on such conformant platform at some stage and possibly this needs to be added to the current set of test setups.

Crog:
Attached is a patch which adds the std:: in the places identified from my unit tests. However as the code is templates there may be further code that uses the C global functions.
[Catch Patch for std namespace.txt](https://github.com/philsquared/Catch/files/50881/Catch.Patch.for.std.namespace.txt)

Crog:
This issue is covered under pull request #559

horenmar:
I just created a branch `add-std-namespace-prefixes`, that should fix most of these. The last one is in `clara.h`, which is an external dependency and will have to be fixed independently.

Can you confirm it works for you now?

philsquared:
This version of Catch will continue with it's own forked version of Clara (as Clara will move on to C++11 and be used by Catch2) - so I just fixed this in-place

PR #544, opened 2015-12-03
Added std:: where it was missing

Crog:
This is in relation to issue #543 and covers the main places where std:: is missing but there may still be further calls.

### DIFF
--- a/include/external/clara.h
+++ b/include/external/clara.h
@@ -35,6 +35,7 @@
 #define TBC_TEXT_FORMAT_H_INCLUDED
 #endif
 
+#include <cctype>
 #include <string>
 #include <vector>
 #include <sstream>
@@ -232,7 +233,7 @@ namespace Clara {
         }
         inline void convertInto( std::string const& _source, bool& _dest ) {
             std::string sourceLC = _source;
-            std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), ::tolower );
+            std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), std::tolower );
             if( sourceLC == "y" || sourceLC == "1" || sourceLC == "true" || sourceLC == "yes" || sourceLC == "on" )
                 _dest = true;
             else if( sourceLC == "n" || sourceLC == "0" || sourceLC == "false" || sourceLC == "no" || sourceLC == "off" )
--- a/include/external/clara.h
+++ b/include/external/clara.h
@@ -232,7 +232,7 @@ namespace Clara {
         }
         inline void convertInto( std::string const& _source, bool& _dest ) {
             std::string sourceLC = _source;
-            std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), ::tolower );
+            std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), std::tolower );
             if( sourceLC == "y" || sourceLC == "1" || sourceLC == "true" || sourceLC == "yes" || sourceLC == "on" )
                 _dest = true;
             else if( sourceLC == "n" || sourceLC == "0" || sourceLC == "false" || sourceLC == "no" || sourceLC == "off" )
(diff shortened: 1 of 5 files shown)


nabijaczleweli:
:+1:

Crog:
Okay, so currently Travis doesn't like it if I changed ::tolower to std::tolower. Some other fix is evidently necessary to fix that issue. In my gcc 4.4.2 compiler I am getting:

    /../external/clara.h:235: error: '::tolower' has not been declared

kirbyfan64:
Is there a `#include <ctype.h>`?

Crog:
Good call @kirbyfan64 will add the missing include into clara.h and catch_common.hpp and put the std:: back in :
    #include <cctype>

Is there an easy way to roll back changes on here (I'm fairly new to using GitHub!)

nabijaczleweli:
@Crog You can use the standard set of git commands

Crog:
Thanks, I will take a look at Git commands soon. For now I can't solve the issue of ::tolower as Travis build system doesn't have a test that replicates a strict conforming environment like I have here :(

Crog:
Cracked it. GCC reports nicely in some versions about std::tolower having ambiguous overloads. Fixing the ambiguity with the following tells the compiler which function to use and this now builds under QNX and all the Travis Test  :)

nabijaczleweli:
:+1:

include/external/clara.h
```diff
@@ -232,7 +233,7 @@ namespace Clara {
         }
         inline void convertInto( std::string const& _source, bool& _dest ) {
             std::string sourceLC = _source;
-            std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), ::tolower );
+            std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), static_cast<int(*)(int)>(std::tolower) );
```
Kosta-Github: this doesn't look good to me...

include/internal/catch_common.hpp
```diff
@@ -22,7 +24,7 @@ namespace Catch {
         return s.find( infix ) != std::string::npos;
     }
     void toLowerInPlace( std::string& s ) {
-        std::transform( s.begin(), s.end(), s.begin(), ::tolower );
+        std::transform( s.begin(), s.end(), s.begin(), static_cast<int(*)(int)>(std::tolower) );
```
Kosta-Github: same here

nabijaczleweli: Will the code work with just `&std::tolower` instead of this mostrosity?

PureAbstract: std::tolower should already be `int std::tolower(int)` - so the cast is a best superfluous, and at worst is undefined behaviour at worst.

nabijaczleweli: There's also `charT tolower( charT ch, const locale& loc );` in `<locale>`, so we're better off using a lambda anyway

philsquared: Can't use lambdas due to C++-03 compatibility.
I think the answer might be along similar lines, though - go via a(n out of line) forwarding function.

philsquared:

include/internal/catch_common.hpp
Crog: From what I can gather this was the only resolution. it is not undefined behavior. It is required to resolve the ambiguity due to the two different impleemntations of std::tolower as @nabijaczleweli  mentions there is a second and is the reason why I had issues. The cast is there to tell the compiler to select the desired function, if there isn't a matching function. 

If it really isn't liked (i wasn't toosure either) then creating a function object which calls std::tolower internally would be a practical way to resolve this.

PureAbstract: Another reasonable approach might be:
```
int (*tolowep)(int) = std::tolower;
std::transform( s.begin(), s.end(), s.begin(), tolowerp );
```
That will select the correct overload, but avoids actually writing `static_cast`.

Crog:
I have opened this request again under #559

PR #546, opened 2015-12-04
Fix uninitialized member in TestRegistry

### LINKED ISSUE #545: Uninitialized member in TestRegistry
Hi,

I'm using ctest to run catch tests monitored by valgrind. As you can see in [this CI report](https://travis-ci.org/01org/parameter-framework/jobs/94879897), valgrind has found an uninitialized variable in TestRegistry. The CI uses the lastest (master) "single include".

It turns out that [`m_currentSortOrder`](https://github.com/philsquared/Catch/blob/master/include/internal/catch_test_case_registry_impl.hpp#L119) is indeed left uninitialized.

It seems that the `m_sortedFunctions.empty()` test above serves to detect the first time that method is called; `m_currentSortOrder` could be initialized there.

dawagner:
Fixes #545.

### DIFF
--- a/include/internal/catch_test_case_registry_impl.hpp
+++ b/include/internal/catch_test_case_registry_impl.hpp
@@ -104,8 +104,10 @@ namespace Catch {
             return m_functions;
         }
         virtual std::vector<TestCase> const& getAllTestsSorted( IConfig const& config ) const {
-            if( m_sortedFunctions.empty() )
+            if( m_sortedFunctions.empty() ) {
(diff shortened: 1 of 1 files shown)


dawagner:
fixed by https://github.com/philsquared/Catch/commit/2b688e1cef85025ed71a0dc885118bb2318c21a0

philsquared:
Sorry - I hadn't spotted this PR before.
I fixed it a slightly different way, anyway.

PR #548, opened 2015-12-07
Make buildable with default main() function for Qt5 under Windows

maxvodo:
Developer environment: Qt5, MinGW 4.8.2, Windows.
Build with default `char * const` type of `argv` argument for default `main()` method fails with errors: 
```
C:/Qt/Qt5.3.2/5.3/mingw482_32/lib/libqtmaind.a(qtmain_win.o): In function `WinMain@16'
c:\work\build\qt5_workdir\w\s\qtbase\src\winmain/qtmain_win.cpp:131: undefined reference to 'qMain(int, char**)'
error: ld returned 1 exit status
```
I propose using of `CATCH_CONFIG_MAIN_ARGV_TYPE` macro with default value `char * const` for all platforms, except Windows.

### DIFF
--- a/include/internal/catch_default_main.hpp
+++ b/include/internal/catch_default_main.hpp
@@ -8,17 +8,29 @@
 #ifndef TWOBLUECUBES_CATCH_DEFAULT_MAIN_HPP_INCLUDED
 #define TWOBLUECUBES_CATCH_DEFAULT_MAIN_HPP_INCLUDED
 
+#include "catch_platform.h"
+
+#ifndef CATCH_PLATFORM_WINDOWS
+  #ifndef CATCH_CONFIG_MAIN_ARGV_TYPE
+    #define CATCH_CONFIG_MAIN_ARGV_TYPE char * const
+  #endif
+#else
+  #ifndef CATCH_CONFIG_MAIN_ARGV_TYPE
+    #define CATCH_CONFIG_MAIN_ARGV_TYPE char *
+  #endif
+#endif
+
 #ifndef __OBJC__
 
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
Why not always have it `const`?

maxvodo:
@nabijaczleweli: Signature `int main(int, char**)` lose `const` qualifier of `int main(int, char* const*)`; and it causes linkage error on my environment.

nabijaczleweli:
`::main()` is (AFAIR) guaranteed to have C linkage, which means, that you can't have a linkage error there...
...unless Qt is so shitty, that it `#define`s `main`

martinmoene:
In N4527 (2015-05-22): 

>3.6.1 Main function [basic.start.main]:
> ...
>(2.2) — a function of (int, pointer to pointer to char) returning int

Elsewhere the document mentions *“pointer to const pointer to const int"*, so I think it's safe to conclude that *pointer* means *non-const pointer* and that using `const` in `main`'s signature is wrong.

nabijaczleweli:
Same in N4140 (*de facto* C++14), interesting. Having *cv-qualified* `::main()` arguments *should* be allowed, though, just as a thought.
Does Qt `#define main` to `qMain`?

philsquared:
Thanks for bringing this up.
There's really no reason for it to be const (other than const usually being the right thing to do - but in this case it makes it non-standard - and it gets handed off to a function that takes as a `char const*` anyway.

So I've reverted it to `char*[]`.

maxvodo:
@philsquared , @nabijaczleweli: thx guys, close request.

PR #557, opened 2015-12-16
ApproxContainer for approximate comparison of stl containers.

kmactavish:
This pull request adds the `ApproxContainer<ContainerType>` class for approximate comparison of values in stl containers. The default container is `std::vector`, but it should work for most stl containers (uses iterators).

I couldn't partially specialize the global `toString` function (since this is templated on container type), so I overloaded the `std::ostream` operator instead.

I haven't updated the single include or added to the documentation yet. If you like the look of it, I can do both before it's merged. I know I'm pull requesting into master, but it is a quick patch to `v1.*` and it seems more appropriate there than `v2.*`. I can move the pull request if you disagree.

E.g.,

```
std::vector<double> a {1., 2., 3.};
std::vector<double> b {1.01, 2.05, 2.99};
CHECK(a == ApproxContainer<>(b).epsilon(0.01));
REQUIRE(a == ApproxContainer<>(b).epsilon(0.1));
```

Output for the failed `CHECK`:

```
  CHECK( a == ApproxContainer<>(b).epsilon(0.01) )
with expansion:
  { 1.0, 2.0, 3.0 }
  ==
  Approx( { 1.01, 2.05, 2.99 } )
```

### DIFF
--- a/include/internal/catch_approx.hpp
+++ b/include/internal/catch_approx.hpp
@@ -43,7 +43,7 @@ namespace Detail {
 
         friend bool operator == ( double lhs, Approx const& rhs ) {
             // Thanks to Richard Harris for his help refining this formula
-            return fabs( lhs - rhs.m_value ) < rhs.m_epsilon * (rhs.m_scale + (std::max)( fabs(lhs), fabs(rhs.m_value) ) );
+            return rhs.compare(rhs.m_value, lhs);
         }
 
         friend bool operator == ( Approx const& lhs, double rhs ) {
@@ -74,9 +74,15 @@ namespace Detail {
             return oss.str();
         }
 
-    private:
+    protected:
+        bool compare ( double lhs, double rhs ) const {
+            return fabs( rhs - lhs ) < m_epsilon * (m_scale + (std::max)( fabs(rhs), fabs(lhs) ) );
+        }
+
         double m_epsilon;
         double m_scale;
+
+    private:
         double m_value;
     };
 }
--- a/include/internal/catch_approx.hpp
+++ b/include/internal/catch_approx.hpp
@@ -42,7 +42,6 @@ namespace Detail {
         }
 
         friend bool operator == ( double lhs, Approx const& rhs ) {
-            // Thanks to Richard Harris for his help refining this formula
             return rhs.compare(rhs.m_value, lhs);
         }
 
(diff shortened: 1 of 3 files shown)


kmactavish:
I can also write a unit test before the merge ;).

include/internal/catch_approx.hpp
```diff
@@ -43,7 +43,7 @@ namespace Detail {
 
         friend bool operator == ( double lhs, Approx const& rhs ) {
             // Thanks to Richard Harris for his help refining this formula
```
nabijaczleweli: I'd move this comment to `compare()`

kmactavish: Good point

maxvodo:
vote :thumbsup:

tomvierjahn:
That looks too specific to me. Cannot this be generalized/generalizable similar to `StringMaker`?
See #954.

philsquared:
I think this would be better as a Matcher.
See https://github.com/philsquared/Catch/blob/master/docs/matchers.md

tomvierjahn:
I dare to disagree ;-)

One of your key selling points is that you can use native operators for comparison. So I would like to do the following while using the same mechanics
```
CHECK(float_a == Approx(float_b));
CHECK(container_a == Approx(container_b));
```
given that `container_.*` contain floating point data.

philsquared:
You are very daring ;-)
However the idea of "native operators" breaks down when we're introducing more specialised, higher level and fine-tunable  ways of performing comparisons.
The `Approx` type already makes "unnatural" use of the comparison operators. If Catch had had Matchers at the beginning I probably would have written it in terms of those, too.
Matchers are also a user extension point - so you can add your own Matchers to do whatever you like and they work the same way. That's not to say that something like an "Approx" Matcher for containers does not belong in the core framework - but just that we don't have to cover all possible cases.

kmactavish:
So I looked back at this, and using the Matchers looks really nice. However, to support such a general extension point, it would really be nice to map any scalar binary or unary function over all the pairs in two containers, and be able to reduce the resulting binary sequence using a number of reduction functions. I'm thinking along the lines of python's `all(map(operator.lt, a, b))` or `any(map(lambda x: x<10, a))`. It makes no sense to have a scalar and container version of every type of comparison you want to do.

However, right now, Matchers only support the unary `match` function, and depend on the second argument being stored inside the Matcher for binary comparisons. Without changing the Matcher base class, I couldn't figure out a nice way of supporting this without resorting to >= c++11 features. Ideas?

PR #559, opened 2015-12-17
Re: fixing missing std:: (see #559)

Crog:
The original #544 got prematurely closed, appears this is the only way to reopen.

**The main criticism of this pull request**:

```
static_cast<int(*)(int)>(std::tolower)
```

This is due to there being two std::tolower functions:

```
 std::tolower(int)
charT tolower( charT ch, const locale& loc );
```

A cast is necessary to tell the compiler to disambiguate which function pointer we are taking.

I propose this change be taken at this time as it resolves the compatibility issues but a new issue should be raised to discuss the best solution as whether to keep, or change to use a function object that calls the std::tolower internally or other viable alternatives.

### DIFF
--- a/include/external/clara.h
+++ b/include/external/clara.h
@@ -35,6 +35,7 @@
 #define TBC_TEXT_FORMAT_H_INCLUDED
 #endif
 
+#include <cctype>
 #include <string>
 #include <vector>
 #include <sstream>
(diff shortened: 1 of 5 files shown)


nabijaczleweli:
Instead of a function object, you can just have a free `catch::tolower()` function

horenmar:
I reimplemented this, because Clara is an external dependency for Catch and should not be changed here + the history was a mess.

PR #538, opened 2015-11-12
Added an optional absolute margin to the approximation checks

CNugteren:
The Approx class now has two options:
* The existing ```epsilon``` to set the tolerated relative error
* The new ```margin``` to set the tolerated absolute error

They can be combined together.

This most probably also fixes another user's issue in #499.

PR #561, opened 2015-12-23
Added an optional absolute margin to the approximation checks

CNugteren:
The Approx class now has two options:
- The existing epsilon to set the tolerated relative error
- The new margin to set the tolerated absolute error

They can be combined together.

This is the same as #538, but that one was closed accidentally.


horenmar:
This has been merged to master, but git doesn't see that.

PR #585, opened 2016-02-05
Implement support for breaking into debugger under Linux

vadz:
This makes `-b` command line option to work under Linux.

Only tested under amd64, but should work on any architectures, in theory.

### DIFF
--- a/include/internal/catch_debugger.h
+++ b/include/internal/catch_debugger.h
@@ -25,29 +25,29 @@ namespace Catch{
     // http://cocoawithlove.com/2008/03/break-into-debugger.html
     #ifdef DEBUG
         #if defined(__ppc64__) || defined(__ppc__)
-            #define CATCH_BREAK_INTO_DEBUGGER() \
-                if( Catch::isDebuggerActive() ) { \
+            #define CATCH_TRAP() \
                     __asm__("li r0, 20\nsc\nnop\nli r0, 37\nli r4, 2\nsc\nnop\n" \
-                    : : : "memory","r0","r3","r4" ); \
-                }
+                    : : : "memory","r0","r3","r4" )
         #else
-            #define CATCH_BREAK_INTO_DEBUGGER() if( Catch::isDebuggerActive() ) {__asm__("int $3\n" : : );}
+            #define CATCH_TRAP() _asm__("int $3\n" : : )
         #endif
     #endif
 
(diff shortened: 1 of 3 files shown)


include/internal/catch_debugger.h
```diff
@@ -25,25 +25,36 @@ namespace Catch{
...
+            #define CATCH_TRAP() _asm__("int $3\n" : : )
         #endif
     #endif
 
+#elif defined(CATCH_PLATFORM_LINUX)
+    // If we can use inline assembler, do it because this allows us to break
+    // directly at the location of the failing check instead of breaking inside
+    // raise() called from it, i.e. one stack frame below.
+    #if defined(__GNUC__) && (defined(__i386) || defined(__x86_64))
+        #define CATCH_TRAP() asm volatile ("int $3")
```
mat-tso: How about using `__builtin_trap()` available on both gcc and clang to avoid the architecture dependence ? Though it will not specifically raise SIGTRAP.

EDIT: On gcc 5.3.0 and clang 3.7.1, it raises SIGILL. It is kind of confusing to see "Illegal instruction" for an expected break. Thus I think your solution is better. :+1:

caryoscelus:
Any reason this isn't merged yet?

horenmar:
Since there is interest in this functionality (#372 is very similar), Ill try to take a look soon.

horenmar:
This PR seems to bring in about 13% runtime penalty for failing tests, even if `-b` is not specified. (Its about 200% with -b specified and no debugger attached).

Since masses of failing tests are not the usual use case, this probably is not a problem, but it seems to suggest that there is something wrong with how the debugger check is gated if `-b` is not specified.

vadz:
This is really strange, I admit I haven't tested the performance implications of this, but I just don't see how could it affect the execution time so much without `-b`. How exactly can I reproduce this, i.e. what methodology did you use for benchmarking?

FWIW with `-b` the code could be optimized by caching the result of `isDebuggerActive()` at the price of not detecting debugger being attached (or, even more rarely, detached) while the test is executing.

Thanks for looking at this!

horenmar:
I've made branch for performance improvements, `dev-performance`. A very simple benchmark is there, with about a 1M failing assertions, so I ran it against this.

Anyway, the penalty isn't that bad, but it could bear looking into later.

PR #588, opened 2016-02-06
Produce valid xml and preserve utf8 characters in attributes

mrpi:
The XML escaping had three problems:
- The escape sequence "&#x**XX**" always has to be terminated with a semicolon
- Bit-values larger than SCHAR_MAX had been converted to "&#x**FFFFFFXX**"
- The input was expected to have exactly one byte per char which is not valid for UTF8. As UTF8 is the default encoding on most modern platforms and the default in XML it seems reasonable to keep valid UTF8 encoded input untouched.

My patch is not perfect for platforms/source code working with other encodings than ASCII or UTF8. But at least the created XML should be accepted by a valid parser.
Everything else would require C++11 u8""-string literals all over the place or an xml encoder that performs character set conversion.

### DIFF
--- a/include/internal/catch_xmlwriter.hpp
+++ b/include/internal/catch_xmlwriter.hpp
@@ -22,7 +22,7 @@ namespace Catch {
    
     // (see: https://en.wikipedia.org/wiki/UTF-8#Codepage_layout)
     namespace Utf8 {
-        inline bool isSingleByteChar(unsigned char b) {
+        inline bool isSingleByteChar( unsigned char b ) {
             // Plain ASCII chars
             return b <= 0x7F;
         }
@@ -31,23 +31,23 @@ namespace Catch {
             return b >= 0x80 && b <= 0xBF;
         }
         
-        inline bool isFirstInTwoByteChar(unsigned char b) {
+        inline bool isFirstInTwoByteChar( unsigned char b ) {
             return b >= 0xC2 && b <= 0xDF;
         }
         
-        inline bool isFirstInThreeByteChar(unsigned char b) {
+        inline bool isFirstInThreeByteChar( unsigned char b ) {
             return b >= 0xE0 && b <= 0xEF;
         }
         
-        inline bool isFirstInFourByteChar(unsigned char b) {
+        inline bool isFirstInFourByteChar( unsigned char b ) {
             return b >= 0xF0 && b <= 0xF4;
         }
         
-        inline bool isInvalidChar(unsigned char b) {
+        inline bool isInvalidByte( unsigned char b ) {
             return b == 0xC0 || b == 0xC1 || b >= 0xF5;
         }
         
-        inline bool isValid(const char* str, size_t len) {
+        inline bool isValid( char const* str, size_t len ) {
             int outstandingBytesOfCurrentChar = 0;
            
             for( std::size_t i = 0; i < len; ++ i ) {
@@ -83,15 +83,15 @@ namespace Catch {
                         return false;
                 }
                                 
-                // explicit negative check (sould be fully redundant here)
-                assert( isInvalidChar( b ) == false );
+                // explicit negative check (should be fully redundant here)
+                assert( isInvalidByte( b ) == false );
             }
             
             return outstandingBytesOfCurrentChar == 0;
         }
         
-        inline bool isValid(const std::string& str) {
-            return isValid(str.c_str(), str.size());
+        inline bool isValid( std::string const& str ) {
+            return isValid( str.c_str(), str.size() );
         }
     }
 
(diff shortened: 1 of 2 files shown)


include/internal/catch_xmlwriter.hpp
```diff
@@ -16,12 +16,88 @@
...
+        
+        inline bool isValid(const std::string& str) {
+            return isValid(str.c_str(), str.size());
+        }
+    }
 
     class XmlEncode {
     public:
-        enum ForWhat { ForTextNodes, ForAttributes };
+        enum ForWhat { ForTextNodes, ForAttributes };
```
nabijaczleweli: y u trailing whitespace

projects/SelfTest/MiscTests.cpp
```diff
@@ -458,10 +458,16 @@ TEST_CASE( "XmlEncode" ) {
...
     }
     SECTION( "string with control char (1)" ) {
-        REQUIRE( encode( "[\x01]" ) == "[&#x1]" );
+        REQUIRE( encode( "[\x01]" ) == "[&#x1;]" );
     }
     SECTION( "string with control char (x7F)" ) {
-        REQUIRE( encode( "[\x7F]" ) == "[&#x7F]" );
+        REQUIRE( encode( "[\x7F]" ) == "[&#x7F;]" );
+    }
+    SECTION( "string with control char that is negativ on signed char (xFF)" ) {
```
nabijaczleweli: `"negative"`

projects/SelfTest/MiscTests.cpp
```diff
@@ -458,10 +458,16 @@ TEST_CASE( "XmlEncode" ) {
...
+        REQUIRE( encode( "[\xFF]" ) == "[&#xFF;]" );
+    }
+    SECTION( "string with utf8 multi byte char (german 'ae' umlaut)" ) {
```
nabijaczleweli: `"multibyte"`

gatopeich:
When is this going to be in a Catch release? The issue renders JUnit output unusable in many cases...

philsquared:
There were a few issues here, and several iterations attempting to address them (thanks for all involved, btw). I'm now not entirely clear if any of it is still outstanding?
Could you please confirm one way or the other, and if anything is outstanding please restate?

mrpi:
Just tested with Catch v1.5.7.
This test:
`TEST_CASE("xml compare failing")
{
     REQUIRE(std::string(u8"普通話") == u8"北方方言");
}`

produces this xml:
`<TestCase name="xml compare failing">
      <Expression success="false" type="REQUIRE" filename="/home/ludger/projects/liquidpp/test/render.cpp" line="23">
        <Original>
          std::string(u8"&#xFFFFFFE6;&#xFFFFFF99;&#xFFFFFFAE;&#xFFFFFFE9;&#xFFFFFF80;&#xFFFFFF9A;&#xFFFFFFE8;&#xFFFFFFA9;&#xFFFFFFB1;") == u8"&#xFFFFFFE5;&#xFFFFFF8C;&#xFFFFFF97;&#xFFFFFFE6;&#xFFFFFF96;&#xFFFFFFB9;&#xFFFFFFE6;&#xFFFFFF96;&#xFFFFFFB9;&#xFFFFFFE8;&#xFFFFFFA8;&#xFFFFFF80;"
        </Original>
        <Expanded>
          "&#xFFFFFFE6;&#xFFFFFF99;&#xFFFFFFAE;&#xFFFFFFE9;&#xFFFFFF80;&#xFFFFFF9A;&#xFFFFFFE8;&#xFFFFFFA9;&#xFFFFFFB1;" == "&#xFFFFFFE5;&#xFFFFFF8C;&#xFFFFFF97;&#xFFFFFFE6;&#xFFFFFF96;&#xFFFFFFB9;&#xFFFFFFE6;&#xFFFFFF96;&#xFFFFFFB9;&#xFFFFFFE8;&#xFFFFFFA8;&#xFFFFFF80;"
        </Expanded>
      </Expression>
      <OverallResult success="false"/>
    </TestCase>`

http://www.xmlvalidation.com says: 
> Errors in the XML document:
> 	4:	38	Character reference "&#xFFFFFFE6" is an invalid XML character.

horenmar:
@mrpi  If I run your test case against current master, it works. As long as its on Linux.

If I try it on Windows, I end up with "Warning C4566: character represented by universal-character-name '\u8A00' cannot be represented in the current code page (1252)" and invalid XML, but I blame that on wrong setting somewhere (which I am not able to find nor google properly).

philsquared:
Works for me too on macOS (with Clang).

philsquared:
Could anyone who originally had problems here (@mrpi?, @gatopeich?) confirm or deny whether there are still issues with this?

mrpi:
XML output looks good now!
Thanks!

philsquared:
Cool - thanks!
I'm closing this now, then

PR #592, opened 2016-02-21
Fix out_of_range exception coming from Tbc::Text

lightmare:
This fixes two bugs in Tbc::Text

**1)** The first bug is subtle, and probably un-checkable by unit test. There's a possibility of undefined behaviour at [tbc_text_format.h#L88](https://github.com/philsquared/Catch/blob/3b4edd7a4849e05fd9d006bf652190729a32906f/include/external/tbc_text_format.h#L88), here's what leads to it:

``` c++
// assume width == remainder.size()
pos = remainder.find_last_of( _attr.tabChar, width );
// assume tabChar was found
if( pos != std::string::npos ) { 
    tabPos = pos;
    if( remainder[width] == '\n' ) // remainder[width] == '\0', width won't be decremented
        width--;
    remainder = remainder.substr( 0, tabPos ) + remainder.substr( tabPos+1 ); // ***
    // now width == remainder.size() + 1
}   

if( width == remainder.size() ) { // it's greater, not equal
    spliceLine( indent, remainder, width );
}   
else if( remainder[width] == '\n' ) { // remainder[remainder.size() + 1] !!!
```

If, at the line marked `***`, there was `remainder.erase( tabPos, 1 )`, then the line marked `!!!` wouldn't be _undefined behaviour_, because `erase` doesn't deallocate memory -- it'd just be reading garbage. But because in C++11 `***` move-assigns from rvalue, you can't safely read beyond its terminating NUL -- its original internal buffer might've been thrown away and the one from the rvalue (whose size including terminating NUL may be exactly width), taken.

**2)** The second bug causes passing "-1" position to `substr` when a tab character appears at `remainder[width]`.

``` c++
// assume width == _attr.width - indent
// assume remainder[width] == _attr.tabChar
pos = remainder.find_last_of( _attr.tabChar, width );
...
// tabPos == width
if( tabPos != std::string::npos )
    indent += tabPos; // now indent == _attr.width
```

In the next round, width will be zero, and [spliceLine( indent, remainder, width-1 )](https://github.com/philsquared/Catch/blob/3b4edd7a4849e05fd9d006bf652190729a32906f/include/external/tbc_text_format.h#L102) will throw `out_of_range`.

### DIFF
--- a/include/external/tbc_text_format.h
+++ b/include/external/tbc_text_format.h
@@ -77,9 +77,9 @@ namespace Tbc {
                 pos = remainder.find_last_of( _attr.tabChar, width );
                 if( pos != std::string::npos ) {
                     tabPos = pos;
-                    if( remainder[width] == '\n' )
+                    if( width == remainder.size() || remainder[width] == '\n' )
                         width--;
-                    remainder = remainder.substr( 0, tabPos ) + remainder.substr( tabPos+1 );
+                    remainder.erase( tabPos, 1 );
                 }
 
                 if( width == remainder.size() ) {
@@ -104,7 +104,7 @@ namespace Tbc {
                     }
                     if( lines.size() == 1 )
                         indent = _attr.indent;
-                    if( tabPos != std::string::npos )
+                    if( tabPos < _attr.width - indent )
                         indent += tabPos;
                 }
             }
--- a/projects/SelfTest/TestMain.cpp
+++ b/projects/SelfTest/TestMain.cpp
@@ -300,7 +300,13 @@ TEST_CASE( "Long strings can be wrapped", "[wrap]" ) {
             == "one two three\n        four\n        five\n        six" );
     }
 
+    SECTION( "Wrapping near tab doesn't extract substrings from invalid positions", "" ) {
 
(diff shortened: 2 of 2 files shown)


horenmar:
Oooh, I think I just hit this one reproducibly. That header is supposed to be an external dependency though, and not patched in this repository, but I'll see where this belongs.

lightmare:
Yes, the file comes with Clara, but even there it looks like it's pulled from somewhere else. Thing is, all the tests of the Text class are here in Catch, not in Clara, and I needed to check that these changes don't break anything else.

(you can checkout or cherry-pick 1501339 "SelfTest checks..." that adds the test case to reproduce)

philsquared:
When I pulled Clara out of Catch Text was common to both, so I had to extract that too. Technically it is an external dependency to both, but I don't think i ever fully extracted it and it is, IIRC, still just part of Clara.
In any event I'll take a look at this one to see how to fit it in - thanks @lightmare (and @horenmar)

horenmar:
This was fixed separately in 4a04682e491b5697979e921c3a5ed9ae933d9406, with the `\t` handling getting cut. I marked the related issue as revisit, in case the handling will be put back in, but I am going to close this for now.

PR #609, opened 2016-03-10
Support closure in REGISTER_TEST_CASE macros. 

willir:
`REGISTER_TEST_CASE` accepts only raw pointer to free function,
Sometimes it is convenient to generate several test_cases based on same
function but with different data.

For example: if one has bunch of inputs for which tested library should
produce certain outputs. In other word if one wants to write something
like:

```
for(const auto &testData : testDataList) {
    REGISTER_TEST_CASE(std::bind(&testCase, testData), "test_case_" + testData.name);
}
```

In this case raw pointers doesn't work. So `REGISTER_TEST_CASE` should
support `std::function`.

This support should be enabled only if compiler supports C++11.
For that purpose `CATCH_CONFIG_CPP11_STD_FUNCTION` define should be added.

Also it is required to change type of members of class `Catch::NameAndDesc` from raw char pointer to `std::string`, to allow one register test cases with different names in the loop easily.


philsquared:
Interesting.
I'm not against this, but I have a few observations (and would need to think about it a bit more):

1. wrt, "if one has bunch of inputs for which tested library should produce certain outputs" - that sounds a lot like a property-based testing approach. I have a first class implementation of this in the works. That said I've been promising it for a while and its not ready yet (it's a big chunk of work and my time is limited).
2. IIRC I originally made `Catch::NameAndDesc` raw char*s because I found that using std::string in those self-registering constructor seemed to add a lot of compile time overhead - at least in XCode. That was some years ago (and Catch itself was smaller and simpler). I should probably revisit that to see if there is still an issue.

Thanks for proposing this. I'll try and get back to you on it.

PR #614, opened 2016-03-19
Update to have correct download URLs

rbuckland:


horenmar:
Changes to the download url have been made since, so I am closing this.

PR #616, opened 2016-03-19
Remove concatenation of empty string in ResultBuilder::reconstructExpression()

RossBencina:
When if( m_exprComponents.op == "" ) is true, there is no need to concatenate with m_exprComponents.op.


horenmar:
Agreed.

Issue #617, opened 2016-03-19
ResultBuilder::reconstructExpression() checks for impossible value: "!"


RossBencina:
The following comparison is present in `ResultBuilder::reconstructExpression()`:

catch_result_builder.hpp: 147

```
   else if( m_exprComponents.op != "!" ) {
```

However, as far as I can tell, `m_exprComponents.op` can never be equal to `"!"`.

`m_exprComponents.op` is constructed as an empty string, and after that is only set via calls to `ResultBuilder::setOp`. There are two calls to `setOp`:

```
catch_capture.hpp(139):                .setOp( "matches" ) \
catch_expression_lhs.hpp(95):      .setOp( Internal::OperatorTraits<Op>::getName() );
```

Here are the possible return values for `OperatorTraits<Op>::getName()`:

```
template<Operator Op> struct OperatorTraits             { static const char* getName(){ return "*error*"; } };
template<> struct OperatorTraits<IsEqualTo>             { static const char* getName(){ return "=="; } };
template<> struct OperatorTraits<IsNotEqualTo>          { static const char* getName(){ return "!="; } };
template<> struct OperatorTraits<IsLessThan>            { static const char* getName(){ return "<"; } };
template<> struct OperatorTraits<IsGreaterThan>         { static const char* getName(){ return ">"; } };
template<> struct OperatorTraits<IsLessThanOrEqualTo>   { static const char* getName(){ return "<="; } };
template<> struct OperatorTraits<IsGreaterThanOrEqualTo>{ static const char* getName(){ return ">="; } };
```

**Proposed Fix**

(Available as PR: https://github.com/philsquared/Catch/pull/618)

Remove the check and delete the default branch, which can never be reached:

```
diff --git a/include/internal/catch_result_builder.hpp b/include/internal/catch_result_builder.hpp
index d453fec..101f922 100644
--- a/include/internal/catch_result_builder.hpp
+++ b/include/internal/catch_result_builder.hpp
@@ -144,7 +144,7 @@ namespace Catch {
             return m_exprComponents.lhs.empty() ? m_assertionInfo.capturedExpression : m_exprComponents.op + m_exprComponents.lhs;
         else if( m_exprComponents.op == "matches" )
             return m_exprComponents.lhs + " " + m_exprComponents.rhs;
-        else if( m_exprComponents.op != "!" ) {
+        else {
             if( m_exprComponents.lhs.size() + m_exprComponents.rhs.size() < 40 &&
[...]

PR #618, opened 2016-03-20
m_exprComponents.op is never assigned the value `!`, so don't check for it (issue #617)

RossBencina:
See issue #617 for detailed analysis.


philsquared:
Resolved elsewhere, as mentioned in #617

PR #623, opened 2016-03-29
Fix unreachable code warning from MSVC

nre-ableton:
MSVC is correct -- this code is never reached. The matching
convertInto() function which is called throws an exception, and the
subsequent function() call is not reached.

When building Catch-based projects with aggressive MSVC warning levels
(eg, with /Wall /W4 /WX flags), then Catch is otherwise warning-free
except for this single instance.

### DIFF
--- a/include/external/clara.h
+++ b/include/external/clara.h
@@ -373,8 +373,13 @@ namespace Clara {
             }
             virtual void setFlag( C& obj ) const {
                 typename RemoveConstRef<T>::type value;
-                convertInto( true, value );
-                function( obj, value );
+                try {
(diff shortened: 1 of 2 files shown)


lightmare:
`convertInto(bool, T&)` throws, but it's only used if `!IsBool<T>`.
There's also `convertInto(bool, bool&)` overload which doesn't throw.

nre-ableton:
@lightmare good catch (no pun intended!). I'll amend and add a try/catch instead.

nre-ableton:
@philsquared great, thanks! I'll close this PR.

Issue #608, opened 2016-03-09
Do not use exit() when detecting duplicate test names


rcdailey:
Slightly related to #472, I find it very difficult to detect when Catch is failing when it calls `exit()` internally. Example:

```
    void enforceNoDuplicateTestCases( std::vector<TestCase> const& functions ) {
        std::set<TestCase> seenFunctions;
        for( std::vector<TestCase>::const_iterator it = functions.begin(), itEnd = functions.end();
            it != itEnd;
            ++it ) {
            std::pair<std::set<TestCase>::const_iterator, bool> prev = seenFunctions.insert( *it );
            if( !prev.second ){
                Catch::cerr()
                << Colour( Colour::Red )
                << "error: TEST_CASE( \"" << it->name << "\" ) already defined.\n"
                << "\tFirst seen at " << prev.first->getTestCaseInfo().lineInfo << "\n"
                << "\tRedefined at " << it->getTestCaseInfo().lineInfo << std::endl;
                exit(1);
            }
        }
    }
```

Note the usage of `exit(1)` above. Instead of that, can you throw an exception? On Windows development, my console window disappears immediately when this happens. Whereas in Visual Studio I can tell my program to stop (auto-breakpoint) when an exception is thrown. I can also set a break point after executing `Catch::Session().run(argc, argv)` if you exit normally.

kirbyfan64:
Why not just set a breakpoint at the `exit` function?

rcdailey:
@kirbyfan64 I think maybe you're misunderstanding. I'm not saying it's impossible to break when duplicates are detected. I've got issues with *how* it's done.

Using `exit()` is unpredictable and difficult to diagnose its point of invocation. I had to grep the code base to find this instance. It's very difficult to maintain a set of breakpoints all over the place to intercept `exit()` calls. Using exceptions here would be far superior. it allows a single point of contact. For example, those that choose to define and control their own `main()` could have an opportunity to handle these exceptions.

This is a design issue that, IMHO, needs to be addressed for the long-term. I'll probably work on this since @philsquared seems pretty swamped right now. Open to feedback on design.

kirbyfan64:
...I meant in the actual exit function, not where it's called (like `break exit` in gdb or [this](http://stackoverflow.com/a/2617724/2097780)) in VS.

I agree with you on the issues of `exit`; I'm just saying that there's a workaround. :)

sldr:
[...]

PR #626, opened 2016-04-01
[#608] Don't use exit() on duplicate test descriptions

rcdailey:
Instead of `exit(1)`, it now throws `std::runtime_error` with the details
of the failure. This exception is handled in `run()` at a higher level where
the log is printed to cerr and the test gracefully exits.

### DIFF
--- a/include/internal/catch_test_case_registry_impl.hpp
+++ b/include/internal/catch_test_case_registry_impl.hpp
@@ -60,13 +60,15 @@ namespace Catch {
             it != itEnd;
             ++it ) {
             std::pair<std::set<TestCase>::const_iterator, bool> prev = seenFunctions.insert( *it );
-            if( !prev.second ){
-                Catch::cerr()
-                << Colour( Colour::Red )
-                << "error: TEST_CASE( \"" << it->name << "\" ) already defined.\n"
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
:+1:

rcdailey:
@philsquared Could you merge this? It's become a real pain to debug duplicate-named test cases and sections, since the application just aborts. This makes it at least possible to break on exception in Visual Studio and the like. It also gives helpful output on command line.

nabijaczleweli:
Phil never was quick to get to issues, was he

PureAbstract:
> Phil never was quick to get to issues, was he

Phil has a rather full life - fixing Catch isn't always right at the top of his priority list.

philsquared:
Guilty as charged!
It's worth the bump as I do try and dip in from time to time even if I'm busy elsewhere (currently preparing to give a course on Swift!). Thanks for the defence, Andy :-)

Anyway, this one looks pretty straightforward and a win-win all round, so I've merged the PR in now: look for build 1.5.3 - and thanks for the PR!

rcdailey:
Thanks @philsquared !! Good luck on your other endeavors!

PR #627, opened 2016-04-01
Include catch header files in CMake script

rcdailey:
This doesn't matter for make systems, but for IDEs such as Visual Studio
on Windows, this allows you to edit Catch header files directly from the
project.

### DIFF
--- a/projects/CMake/CMakeLists.txt
+++ b/projects/CMake/CMakeLists.txt
@@ -5,7 +5,10 @@ project(Catch)
 # define some folders
 get_filename_component(CATCH_DIR "${CMAKE_CURRENT_SOURCE_DIR}" PATH)
 get_filename_component(CATCH_DIR "${CATCH_DIR}" PATH)
+
 set(SELF_TEST_DIR ${CATCH_DIR}/projects/SelfTest)
+set(INCLUDE_DIR ${CATCH_DIR}/include)
+
 if(USE_CPP11)
   ## We can't turn this on by default, since it breaks on travis
   message(STATUS "Enabling C++11")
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
Why would you wanna do that?

rcdailey:
I explained it in my commit message and PR description.

nabijaczleweli:
You explained what it lets one *do*, not *why* anyone would want to do that.

rcdailey:
So if I'm understanding you correctly, you need to know *why* I'd ever want to see files and edit them directly from Visual Studio. I felt the benefits were obvious.

Build systems like make or ninja only function on translation units, because header files do not compile into object files. However, CMake still accepts them in the list of source files for IDEs. They are not explicitly acted on by the toolchain, but CMake will make them appear in the IDE if you specify them. This allows for intellisense, which enables you to find symbols from within the IDE. Without these files included, I had to grep outside of Visual Studio to find code inside them. This is a huge pain.

Linux developers that work with make or ninja directly are not impacted by this change and it does not benefit them unless they use some kind of IDE.

nabijaczleweli:
Hrmmm, alright. In your PR desc and commit msg you mentioned *editing* Catch which sounded suspicious.
:+1:

horenmar:
Since we want to move build to base CMake based (with IDE specific files generated on-demand), I will look into this soon.

horenmar:
Can you check whether the current CMakeLists work for you? I tried generating solution for VS 2015 and the header files are included.

philsquared:
I added all the header files to CMakeLists.txt last night so I think we can close this PR

philsquared:
I've also confirmed it working in both Xcode and VS projects generated from CMake.

PR #628, opened 2016-04-02
removes uneeded parentheses

luxe:


nabijaczleweli:
They are very much needed.

lightmare:
In case anyone wonders *why* they are needed, it's because some systems (win?) define function-like macro `max(a, b)`, and expanding that after `std::` would cause compilation error. Wrapping `std::max` in parentheses prevents that.

luxe:
I see.

PR #630, opened 2016-04-04
Promote interface to parent class instead of using dynamic_cast

eirikba:
This eliminates one use of RTTI. Along with another pull request I will send shortly, this makes it possible to compile Catch with gcc and -fno-rtti.

I also did a clean-up while I was there (in the second commit). That cleanup could be done without eliminating the dynamic_cast, too.

### DIFF
--- a/include/internal/catch_interfaces_reporter.h
+++ b/include/internal/catch_interfaces_reporter.h
@@ -226,6 +226,13 @@ namespace Catch
         // Implementing class must also provide the following static method:
         // static std::string getDescription();
 
+        virtual bool supportsChainedReporters() const {
+            return false;
+        }
+
+        virtual void addChainedReporter( Ptr<IStreamingReporter> const& reporter ) {
+        }
+
         virtual ReporterPreferences getPreferences() const = 0;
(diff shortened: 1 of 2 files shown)


eirikba:
Another solution would be to add asMultipleReporters() to IStreamingReporter and have it return nullptr there, and override it to return this in MultipleReporters. But I think this design is cleaner.

eirikba:
A third solution that might be better would be to implement MultipleReporters as a linked list instead of holding a vector. I will code that up and send a pull request, so you can compare the solutions and see which one you prefer.

eirikba:
Personally I prefer #636 over this one.

philsquared:
Sorry it's taken a while to get back to this. I've now made some changes - more along the lines of your original PR, I think (I wanted to minimise the overhead of the majority case - when there is only one reporter. The cost was just one virtual method on the interface relating to multiple reporters).
Please see v1.5.1

eirikba:
That looks pretty much like my second suggestion, yes :) I believe that solves my issue. Thanks.

PR #631, opened 2016-04-04
Replace two dynamic_casts with static_casts

eirikba:
This eliminates two uses of RTTI in Catch. Along with the pull request I just sent, this makes it possible to compile Catch with gcc and -fno-rtti.

### DIFF
--- a/include/internal/catch_test_case_tracker.hpp
+++ b/include/internal/catch_test_case_tracker.hpp
@@ -234,10 +234,8 @@ namespace TestCaseTracking {
             SectionTracker* section = CATCH_NULL;
 
             ITracker& currentTracker = ctx.currentTracker();
-            if( ITracker* childTracker = currentTracker.findChild( name ) ) {
-                section = dynamic_cast<SectionTracker*>( childTracker );
-                assert( section );
-            }
+            if( ITracker* childTracker = currentTracker.findChild( name ) )
+                section = static_cast<SectionTracker*>( childTracker );
             else {
                 section = new SectionTracker( name, ctx, &currentTracker );
                 currentTracker.addChild( section );
@@ -265,10 +263,8 @@ namespace TestCaseTracking {
             IndexTracker* tracker = CATCH_NULL;
 
             ITracker& currentTracker = ctx.currentTracker();
-            if( ITracker* childTracker = currentTracker.findChild( name ) ) {
-                tracker = dynamic_cast<IndexTracker*>( childTracker );
-                assert( tracker );
-            }
+            if( ITracker* childTracker = currentTracker.findChild( name ) )
+                tracker = static_cast<IndexTracker*>( childTracker );
             else {
                 tracker = new IndexTracker( name, ctx, &currentTracker, size );
                 currentTracker.addChild( tracker );


eirikba:
If you are attached to the assert, you could add "assert( dynamic_cast<SectionTracker*>( childTracker ) );" to each of the changes. That will break compilation in debug mode with -fno-rtti, but that's not very important.

kirbyfan64:
What if you do something like:

```c++
#if defined(__GNUC__) && !defined(__GXX_RTTI)
// No RTTI.
#define CATCH_CAST static_cast
#else
#define CATCH_CAST dynamic_cast
#endif
```

and leave the asserts.

eirikba:
Possible, I guess. But that would require extending that test for each compiler you would want to support. And besides, I don't like making different compilations see different code. I find that's a good place for bugs to hide.

lightmare:
Why is there a "generic tree interface" `ITracker` when all its nodes have to be `SectionTracker`? It looks like the whole tree should've been templated on node type. I'd get rid of `ITracker` and make `TrackerContext` and `TrackerBase` templates, like this:
```c++
template <typename TrackerT> class TrackerContext<TrackerT> { ... };
template <typename TrackerT> class TrackerBase<TrackerT> { ... };

class SectionTracker : public TrackerBase<SectionTracker> { ... };
typedef TrackerContext<SectionTracker> SectionTrackerContext;
```

That should be possible with quite minimal code changes. I can make a pull req if you're interested.

NB: `IndexTracker` looks like a forgotten unused class, btw.

eirikba:
That sounds like a very good question. If you don't make that change, I might try coding up some variant of it and see what it looks like :)

philsquared:
Just jumping in here - I'd have to check the code (which I don't have access to at the moment) but I believe the hierarchy you're looking at (`ITracker`-based) is a pre-requisite for the property-based testing implementation I'm working on (generators need to interact with sections, so they need to be modelled as peer nodes).

That's not to say there are no opportunities for clean-up or optimisation around there - I just wanted to let you know there is a reason for it being that way that is not currently reflected in the code on GitHub.

philsquared:
... on the subject of optimisation, and back towards the spirit of the PR itself, this is similar to something I have implemented in other performance sensitive code-bases. For cases where I don't *expect* a down cast to fail I provide a simple cast template, `safePolyCast<>`, which is used like `static_cast` or `dynamic_cast` and internally forwards to the former in a release build and the latter in a debug build. In debug the result is also checked and asserted on (accounting for null being passed in).
It would be trivial to add something similar to Catch and, I believe, would meet the need being discussed here. I haven't done so before as I had not considered it a performance bottleneck in practice.
If you are implying that you do see it as a performance bottleneck then I'll add it.
Of course it still requires RTTI - although that can also be trivially conditionally compiled - but now any code that needs such a facility is localised to one place.

eirikba:
I'm not (not yet, anyway) concerned about performance at all. My reason for making this patch is only to avoid requiring RTTI. There are plenty of projects that disable RTTI for various reasons, and if Catch requires RTTI that means Catch can not be used with those projects. And that just seems silly in light of how little use Catch actually has for RTTI.

philsquared:
As mentioned in #630, I've resolved the underlying issue

eirikba:
Yes, that should do the trick. Thanks.

PR #634, opened 2016-04-05
Update tutorial.md

### LINKED ISSUE #633: Download link not working
Hello

I'm trying to use Catch for the very first time, and reading the tutorial, I've tried to download from the link provided there (http://builds.catch-lib.net/), but the link is broken.

rafaeleyng:
Closes #633


philsquared:
Thanks - I need to check why that url is not fowarding

PR #635, opened 2016-04-05
XmlEncode::encodeTo(): incorrect escaping of UTF-8 symbols

vzelenkin:
Fixed two issues in the XmlEncode::encodeTo
1. Signed char symbols with code > 127 became negative and were escaped as control chars
2. In any cases were generated incorrect xml escape sequence without ; symbol at the end

Added corresponding test for encodeTo


mrpi:
Your patch solves 2 of 3 problems with this code section. The third case is, that the output has to be utf-8 even if the input strings are not utf-8 encoded (very likely for random byte sequences).
I am escaping the chars > 127, if the hole value is not a valid utf-8 sequence (#588).

philsquared:
It's not clear if this issue has been addressed by my recent acceptance of one of the PRs in #588.
Could you please confirm whether any of your reported issues are still outstanding and, if so, which ones? I'm trying to close off all the XML reporter issues.

philsquared:
I believe the chars > 127 part of this has been covered now by the merging of PR #767.
Have I read that correctly? Is there any more that is still outstanding?

vzelenkin:
Yes, both of mentioned issues have been fixed by PR #588 and PR #767. So I close my PR.

philsquared:
thanks

PR #636, opened 2016-04-05
Eliminate dynamic cast by implementing MultipleReporters as a linked list

eirikba:
This is an alternative to the pull request I made yesterday (https://github.com/philsquared/Catch/pull/630). It eliminates the same use of RTTI as 630, and together with 631 (or some alternative to that one) makes Catch compile successfully with gcc and -fno-rtti.

I think I prefer this solution to 630. It avoids polluting the IStreamingReporter interface with methods that doesn't really have anything to do with reporting. Design-wise, I think I prefer this  to the original as well.

Note that I left MultiReporters::add() in since it is a public method. It is no longer used by addReporter(), though. So if it isn't used elsewhere and isn't considered part of the public API it can probably be dropped.

### DIFF
--- a/include/reporters/catch_reporter_multi.hpp
+++ b/include/reporters/catch_reporter_multi.hpp
@@ -13,132 +13,121 @@
 namespace Catch {
 
 class MultipleReporters : public SharedImpl<IStreamingReporter> {
-    typedef std::vector<Ptr<IStreamingReporter> > Reporters;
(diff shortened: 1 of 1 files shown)


philsquared:
As mentioned in #630, I've resolved the underlying issue.
Thanks for the PRs

eirikba:
Thanks.

PR #643, opened 2016-04-20
Fix warning: conversion from 'int' to 'char', possible loss of data.

xieyubo:
Root cause:
The return type of ::tolower is int. So when we use ::tolower in
std::transform(), it will generate this warning. If user turns on "treat
warning as error", Catch build will fail due to this warning.

### DIFF
--- a/include/external/clara.h
+++ b/include/external/clara.h
@@ -375,6 +375,10 @@ namespace Clara {
             return str.size() >= prefix.size() && str.substr( 0, prefix.size() ) == prefix;
         }
 
+        inline char tolower(char ch) {
+            return (char)::tolower(ch);
+        }
+
         template<typename T> struct RemoveConstRef{ typedef T type; };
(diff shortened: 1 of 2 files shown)


abominous:
Why not just cast return type?
std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), (char (*)(int))tolower );

abominous:
Why not just cast return type?
std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), (char (*)(int))tolower );

wilx:
Re https://github.com/philsquared/Catch/pull/643#issuecomment-231393297:
> Why not just cast return type?
> `std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), (char (*)(int))tolower );`

This is UB.

horenmar:
As far as I can see, #728 was already merged and fixes this. Closing.

PR #646, opened 2016-04-21
Put quote marks around printed characters; also nicely print some escapes

Eric41293:
This change puts quote marks around printed characters in expression expansions. Previously, for instance,
`REQUIRE('\0'=='0');`
would result in the confusing expansion
`0 == 0`
It now results in
`0 == '0'`
(The initial impetus for this change was a case in which a type was an alias for unsigned char, and the two values happened to be `' '` and `'0'`, thus resulting in `== 0`, which confused me for a while. The expansion `' ' == '0'` is much clearer.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ matrix:
       compiler: clang
       addons: &clang38
         apt:
-          sources: ['llvm-toolchain-precise', 'ubuntu-toolchain-r-test']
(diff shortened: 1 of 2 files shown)


nabijaczleweli:
As for the failing Clang3.8 CI build you need to make [this package](https://github.com/philsquared/Catch/blob/master/.travis.yml#L67) `llvm-toolchain-precise-3.8`

horenmar:
I merged this to master in 5121b5b05886c3667fa8ba738b2f55d138d35490.

PR #647, opened 2016-04-22
add missing argument to CATCH_CHECK_THROWS (closes #602)

### LINKED ISSUE #602: incorrect signature used for CATCH_CHECK_THROWS?
Compare `CATCH_CHECK_THROWS` and `CATCH_REQUIRE_THROWS`:

https://github.com/philsquared/Catch/blob/master/single_include/catch.hpp#L10286
https://github.com/philsquared/Catch/blob/master/single_include/catch.hpp#L10275

It looks like the `INTERNAL_CATCH_THROWS` macro takes 4 arguments; however, `CATCH_CHECK_THROWS` only passes on 3. Should it pass along an empty 'matcher' argument?

kevinushey:
`INTERNAL_CATCH_THROWS` now takes a matcher as the third argument, but `CATCH_CHECK_THROWS` was not updated to accommodate that. This PR should resolve that.


horenmar:
Its interesting that this bug got to live for so long.

PR #658, opened 2016-05-12
add support for comparison with strong typedefs

### LINKED ISSUE #652: Add support to Approx for comparison with constructible types
Approx is proving very useful to me. It would be more useful still if comparison between strong typedefs and Approx could be made to work without explicit conversions.

```

class Volatility {
  double underlying_;
 public:  
  explicit Volatility(double u) : underlying_(u) {}
  explicit operator double() const { return underlying_; }
}

auto v = Volatility(0.3);

REQUIRE(v == Approx(0.3)); // does not compile
REQUIRE(static_cast<double>(v) == Approx(0.3)); // compiles but I'd like to avoid verbosity

```

I'd be happy to do the work for this.

jbcoe:
closes #652

### DIFF
--- a/include/internal/catch_approx.hpp
+++ b/include/internal/catch_approx.hpp
@@ -13,6 +13,10 @@
 #include <cmath>
 #include <limits>
 
+#if defined(CATCH_CPP11_OR_GREATER)
+#include <type_traits>
+#endif
+
 namespace Catch {
 namespace Detail {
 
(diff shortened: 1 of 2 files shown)


philsquared:
Thanks for this.
All looks reasonable, except for the `enable_if`s and `std::is_constructible` - since I need to maintain support for C++98/ 03.
I might be able to come up with some raw SFINAE workarounds, although I'm little reluctant to use SFINAE at all as so many pre-C++11 compilers have patchy support for it.
The best option might be to conditionally compiler the whole thing so that only C++11+ compilers see the templated version?

jbcoe:
Sounds good to me. Do you have existing examples I can base an improved PR upon?

philsquared:
I'm not quite sure what you mean by existing examples. Specific to `Approx`? or in general?

Anyway, we happened to meet at lunch today (small world!) and you also mentioned that you wanted an overload of `<` (and `>`) on `Approx` - and this worried me a little. My main concern was that `<` would not always be the inverse of `>`. It also doesn't quite _feel_ right to me to say, effectively: "Is this number less than that one (or just a little bit more - that's ok too)". I can't quite articulate _why_ that feels so wrong - and it's probably just a misleading gut feel. I'm hoping others may be watching this thread who may have an opinion and want to weigh in?

jbcoe:
I've updated the PR so tests pass.

I wonder if you had existing macros to deal with conditional c++ 11 support?

I agree that having < and > both true is daft. All the LP and QP libraries I've seen expose <= and >= instead. If Catch did the same and Approx supported <= and >= would that be ok?

jbcoe:
Are further changes to this PR required?

jbcoe:
Are further changes needed to merge this? 

I've been using my patched version of Catch without any issues for a few months now.

include/internal/catch_approx.hpp
```diff
@@ -40,7 +44,19 @@ namespace Detail {
             approx.scale( m_scale );
             return approx;
         }
+#if defined(__cplusplus) && __cplusplus >= 201103L
+        template <typename T, typename = typename std::enable_if<std::is_constructible<T, double>::value>::type>
```
lightmare: This is enabled when `T` is constructible from `double`. But you actually do the inverse below. Should use `std::is_constructible<double, T>` here.

jbcoe: Well spotted. Fixed.

jbcoe:
test failures are addressed by #716 and are unconnected with this change.

philsquared:
Sorry I've been quiet on this. I'm gradually getting up to speed again now!

From my initial quick scan this looks good (with @lightmare's fix) but I now see what you mean about the conditional C++ macros. Catch `#define`s `CATCH_CPP11_OR_GREATER` for that purpose.

include/internal/catch_approx.hpp
```diff
@@ -44,7 +44,7 @@ namespace Detail {
             approx.scale( m_scale );
             return approx;
         }
-#if defined(__cplusplus) && __cplusplus >= 201103L
+#if CATCH_CPP11_OR_GREATER
```
lightmare: use `#ifdef`, it doesn't have a value

martinmoene: Why not "[Why you should prefer #ifs over #ifdefs](https://www.iar.com/support/resources/articles/advanced-preprocessor-tips-and-tricks/)" (somewhat past halfway)  :gem:

jbcoe:
I'll push support for != too. Seems silly to leave this out.

include/internal/catch_approx.hpp
jbcoe: Amended.

jbcoe:
Squashed into one commit and ready to go.

jbcoe:
I've rebased this again to handle the incorporation of <= and >= operators.
I chose to rebase to keep commit history clean.

horenmar:
This is now in master, I added a new feature toggle so that MSVC can use this as well (it still doesn't trigger the C++11 feature check).

PR #660, opened 2016-05-16
Fix string [] access beyond range. cppcheck complains. 

RossBencina:
Fix string iteration accesses null terminator. cppcheck complains.

Alternatively, if this is intentional, a comment would be appreciated, such as: 

```
for( std::size_t i = 0; i <= arg.size(); ++i ) { // intentional access of terminating-'\0'
```


horenmar:
This has been fixed in Clara (which is theoretically an external dependency) so I will close the PR.

PR #669, opened 2016-06-05
--list-test-names-and-sources

Cyberboss:
I'm working on a Visual Studio unit test adapter for Catch and was looking for a way to view the source file and line of a test without actually running it. So I implemented this

Sample output:

Scenario: Test that threaded operations can be canceled
s:\documents\actual documents\da git\cyberenginemkiii\tests\platform\system\threads.cpp(78)
Scenario: Test basic sleep works
s:\documents\actual documents\da git\cyberenginemkiii\tests\platform\system\threads.cpp(109)
Scenario: Test basic yield works
s:\documents\actual documents\da git\cyberenginemkiii\tests\platform\system\threads.cpp(126)
Scenario: Test thread system errors work
s:\documents\actual documents\da git\cyberenginemkiii\tests\platform\system\threads.cpp(150)

### DIFF
--- a/include/internal/catch_config.hpp
+++ b/include/internal/catch_config.hpp
@@ -33,6 +33,7 @@ namespace Catch {
             listReporters( false ),
             listTestNamesAndSources( false ),
             listTestNamesOnly( false ),
+            listTestSources( false ),
             showSuccessfulTests( false ),
             shouldDebugBreak( false ),
             noThrow( false ),
@@ -52,6 +53,7 @@ namespace Catch {
         bool listTags;
         bool listReporters;
         bool listTestNamesAndSources;
+        bool listTestSources;
         bool listTestNamesOnly;
 
         bool showSuccessfulTests;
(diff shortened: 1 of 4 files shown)


cybnetsurfe3011:
Clang builds failed because http://llvm.org/apt/llvm-snapshot.gpg.key 404'd

nabijaczleweli:
Can confirm, `llvm.org/apt` has been down for the last couple of days

include/internal/catch_config.hpp
```diff
@@ -50,7 +52,9 @@ namespace Catch {
         bool listTests;
         bool listTags;
         bool listReporters;
+        bool listTestNamesAndSources;
         bool listTestNamesOnly;
+		bool listTestSources;
```
martinmoene: Mmm, a Tab character!

cybnetsurfe3011: A) VS is a sneaky mistress
B) Whoops, I've been commiting to master when I have a pull request open

philsquared:
Thanks for this (and apologies for the delay in getting to it). While this looks like a useful PR I'm about to make some changes elsewhere that will make it redundant.
That is, I'm going to make `-l` pass off to the reporter to format its output. That way you can use the xml reporter to provide machine readable output that will have this information.

Cyberboss:
Alright, no problem

PR #670, opened 2016-06-06
Modified team city reporter to report testIgnored if it was skipped

pnemere:
TeamCity was always running tests and reporting them as failed, even though locally the [!mayfail] tag prevented them from failing (and just printed the output).

This change makes such cases report as testIgnored, which in team city still shows the test assertion but does not treat it as an error, mirroring what the local console run does.


pnemere:
Code indentation: Switched \t to spaces

horenmar:
At some point this change made it into the teamcity reporter independently (?), so I am going to close this PR.

PR #699, opened 2016-08-22
Fix Visual Studio analyze warnings

zachstruck:
This fixes warnings reported by Visual Studio's static analysis enabled by the "/analyze" compiler switch.


horenmar:
IIRC both should be fixed in current master, the first one for quite a bit and I just commited fix for the second one -- we now just `windows.h` to get the proper declaration, since the file should not be compiled outside of main file it does not hurt the compile times so much.

PR #701, opened 2016-08-24
Fixes for XML encoding.

razeh:
This commit fixes the following scenario:
  * You have a test that compares strings with embedded control
  characters.
  * The test fails.
  * You are using JUnit tests within TeamCity.

Before this commit, the JUnit report watcher fails on parsing the XML
for two reasons: the control characters are missing a semicolon at the
end, and the XML document doesn't specify that it is XML 1.1.

XML 1.0 --- what we get if we don't specify an XML version --- doesn't support embedding control characters --- see
http://stackoverflow.com/questions/404107/why-are-control-characters-illegal-in-xml
for all of the gory details.

This is based on PR #588 by @mrpi

### DIFF
--- a/include/internal/catch_xmlwriter.hpp
+++ b/include/internal/catch_xmlwriter.hpp
@@ -55,9 +55,10 @@ namespace Catch {
                         break;
 
                     default:
-                        // Escape control chars - based on contribution by @espenalb in PR #465
+                        // Escape control chars - based on contribution by @espenalb in PR #465 and
+                        // by @mrpi PR #588
                         if ( ( c < '\x09' ) || ( c > '\x0D' && c < '\x20') || c=='\x7F' )
-                            os << "&#x" << std::uppercase << std::hex << static_cast<int>( c );
+                            os << "&#x" << std::uppercase << std::hex << std::setfill('0') << std::setw(2) << static_cast<int>( c ) << ';';
                         else
                             os << c;
                 }
@@ -112,13 +113,20 @@ namespace Catch {
         :   m_tagIsOpen( false ),
             m_needsNewline( false ),
             m_os( &Catch::cout() )
-        {}
+        {
+            // We encode control characters, which requires
+            // XML 1.1
+            // see http://stackoverflow.com/questions/404107/why-are-control-characters-illegal-in-xml-1-0
+            *m_os << "<?xml version=\"1.1\" encoding=\"UTF-8\"?>\n";
+        }
 
         XmlWriter( std::ostream& os )
         :   m_tagIsOpen( false ),
             m_needsNewline( false ),
             m_os( &os )
-        {}
+        {
+            *m_os << "<?xml version=\"1.1\" encoding=\"UTF-8\"?>\n";
+        }
 
         ~XmlWriter() {
             while( !m_tags.empty() )
--- a/projects/SelfTest/MiscTests.cpp
+++ b/projects/SelfTest/MiscTests.cpp
@@ -406,27 +406,27 @@ TEST_CASE( "Tabs and newlines show in output", "[.][whitespace][failing]" ) {
 
 
 TEST_CASE( "toString on const wchar_t const pointer returns the string contents", "[toString]" ) {
-	const wchar_t * const s = L"wide load";
-	std::string result = Catch::toString( s );
-	CHECK( result == "\"wide load\"" );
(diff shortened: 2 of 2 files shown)


philsquared:
Thanks for that - sorry it took a while to get to.
I can't believe I missed the final ; on the encoding!
The other bits make sense too.

philsquared:
/insert face palm gif here

I'll have to look at making it optional.
That's going to suck, though. Looks like there's no clear win here?

razeh:
Yeap :-(

An alternative would be to escape the strings ourselves into something
normal text, rather than trying to represent them as XML.

If you'd be willing to go down that line of thought, I can try coming up
with something over the weekend.

Robert

razeh:
Argh.  I hit send too soon.

An alternative would be to escape the strings into normal text --- and not
trying to come up with the right XML representation for binary.

If you'd be willing to go entertain that line of thought, I can try coming
up with something over the weekend.

Robert

philsquared:
That would be great if you could come up with something - thanks!

include/internal/catch_xmlwriter.hpp
```diff
@@ -55,9 +55,10 @@ namespace Catch {
                         break;
 
                     default:
-                        // Escape control chars - based on contribution by @espenalb in PR #465
+                        // Escape control chars - based on contribution by @espenalb in PR #465 and
+                        // by @mrpi PR #588
                         if ( ( c < '\x09' ) || ( c > '\x0D' && c < '\x20') || c=='\x7F' )
-                            os << "&#x" << std::uppercase << std::hex << static_cast<int>( c );
+                            os << "&#x" << std::uppercase << std::hex << std::setfill('0') << std::setw(2) << static_cast<int>( c ) << ';';
```
mrpi: `static_cast<int>( c )` is wrong. Has to be `static_cast<int>( static_cast<unsigned char>(c) )`

mrpi:
@razeh sorry that I did not reply earlier. Your patch seems to address the problems with 7bit ASCII (better than my patch did). But you ignored byte values greater 0x7F.
P.S.: There is a difference between "random" binary and utf8. But if I would be living in the US or UK, I would also try to ignore that ... ;-)

include/internal/catch_xmlwriter.hpp
razeh: Is there a test I can add to catch this for next time?

Bu11etmagnet: Might as well `static_cast<unsigned int>` from `unsigned char`

gerikintercede:
The XML header is always written to the console and not the output file:

> Tests.exe -r xml -o test.xml
> <?xml version="1.1a" encoding="UTF-8"?>

This fix to catch_reporter_xml.hpp fixes it:
     public:
         XmlReporter( ReporterConfig const& _config )
         :   StreamingReporterBase( _config ),
-            m_sectionDepth( 0 )
+            m_sectionDepth( 0 ),
+            m_xml( _config.stream() )
         {
             m_reporterPrefs.shouldRedirectStdOut = true;
         }

Found in 1.5.8

philsquared:
Sorry it's taken me a while to come back to this, but I've been looking at the XML 1.1 issue.
I'd forgotten about what was discussed here, but ended up going down the same route anyway - escaping the control chars instead of encoding them.
See:
https://github.com/philsquared/Catch/tree/Xml1_0

I'd appreciate your thoughts on this before I merge, @razeh, @mrpi, @IlyaFinkelshteyn ?

razeh:
I think escaping the control characters is the right route to go. I don't think the XML 1.1 support is worth the added complexity; XML 1.1 was only added so we could encode the control characters, and I don't think that has a lot of value. Escaping them is good enough.

IlyaFinkelshteyn:
If I remember correctly in our case in [AppVeyor](https://www.appveyor.com/), test result parser rely on .NET so we could not parse catch test results. If results become XML 1.0 again, AppVeyor customers who use Catch will be happy :)

philsquared:
I've now pushed these changes onto master. They should be in the next patch release (technically they count as breaking changes and should at least be reserved for a a minor release, but we just did one of those and I don't think it's worth waiting on this).

philsquared:
I believe that should conclude this issue, but I'll leave it open for a while for further comment, if any.

PR #705, opened 2016-08-30
Do not trim test case names in the XML reporter.

hmich:
SCENARIO does not add leading spaces to the test name (only BDD-style section
names are modified), so the trimming is not necessary. But if the name is
trimmed, it makes it harder to correlate the output of XML reporter with tests
that have leading spaces in their name: e.g. these tests will have the same name
attribute:

```
TEST_CASE("Test") {}
TEST_CASE(" Test") {}
```


philsquared:
Thanks

PR #715, opened 2016-09-24
Add support for <= and >= for Approx

### LINKED ISSUE #651: Add support for <=, >= to Approx
Approx is very useful and I'd make further use of it if support for <=, >= could be added.

I'll be happy to do this work.

jbcoe:
closes #651

### DIFF
--- a/include/internal/catch_approx.hpp
+++ b/include/internal/catch_approx.hpp
@@ -58,6 +58,26 @@ namespace Detail {
             return !operator==( rhs, lhs );
         }
 
+        friend bool operator <= ( double lhs, Approx const& rhs )
+        {
+          return lhs < rhs.m_value || lhs == rhs;
+        }
+
+        friend bool operator <= ( Approx const& lhs, double rhs )
+        {
+          return lhs.m_value < rhs || lhs == rhs;
+        }
+
+        friend bool operator >= ( double lhs, Approx const& rhs )
+        {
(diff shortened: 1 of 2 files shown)


jbcoe:
failing tests are unconnected with this change and addressed by https://github.com/philsquared/Catch/pull/716

onqtam:
looking at this PR - isn't it possible to also implement less-than and greater-than - like this:

```
    friend bool operator< (double lhs, Approx const& rhs) { return lhs < rhs.m_value && lhs != rhs; }
    friend bool operator< (Approx const& lhs, double rhs) { return lhs.m_value < rhs && lhs != rhs; }
    friend bool operator> (double lhs, Approx const& rhs) { return lhs > rhs.m_value && lhs != rhs; }
    friend bool operator> (Approx const& lhs, double rhs) { return lhs.m_value > rhs && lhs != rhs; }
```

???

jbcoe:
@onqtam I think you are right. I'm not sure I'd find `<`, `>` useful for `Approx`. Do you have a use-case in mind?

onqtam:
not really - but just for completeness' sake

jbcoe:
`REQUIRE(x < Approx(2.3));`

seems a bit odd to me. I'd be reluctant to add your functions for completeness' sake without a motivating use case.

horenmar:
I am against adding those.

PR #716, opened 2016-09-24
Quick fix for failing travis builds on OS X

jbcoe:
I'm not happy with this fix but I don't (currently) understand travis-ci well enough to fix the real problem.
For now this gets CI passing.


jbcoe:
Any issues with this?

philsquared:
only the usual ... my time - and missing GitHub notifications :-)
Thanks for the bump. I've accepted the PR as-is (I was looking at doing similar myself - but you got ahead of me!)

Issue #387, opened 2015-03-24
Suppress warnings generated by MSVC /Wall


socantre:
There are a number of warnings that the catch header generates when compiled with MSVC and which should be suppressed. Here's what I found with VS2012 and /Wall

C4820 padding
C4625 copy constructor could not be generated because a base class copy constructor is inaccessible
C4626 assignment operator could not be generated because a base class assignment operator is inaccessible
C4512 assignment operator could not be generated
C4514 unreferenced inline function has been removed
C4350 behavior change from previous version of compiler
C4571 behavior change from previous version of compiler
C4371 class layout change from previous version of compiler
C4710 functions not inlined

The header should also be tested with clang's -Weverything.

Trass3r:
There's a good reason many warnings aren't even enabled on /W4.

philsquared:
/Wall generates loads of warnings even in the std lib that ships with Visual Studio. I know the MS std lib team were doing a lot of work to address that, but AFAICS that's still the case even in VS2017.
I get no warnings in MSVC with /W4. Quite a few warnings have been addressed in the last couple of years (e.g those in the referenced #722).
Is this still a legitimate issue for you?

socantre:
I would still really like third party headers to compile cleanly even with /Wall. This makes it much easier to use the 'subtractive' strategy for enabling warnings.

This doesn't mean that all the warnings need to be "fixed". If the headers are intended to be warning free at some particular warning level then the headers should internally force that warning level, rather than ask users to externally apply warning suppression (which is my current solution).

The changes for VS2017 includes a 'warning overhaul', in which they apply something similar, but a bit more complicated in order to support two internal levels of warnings. They mention that /Wall works for now, but is unsupported. I hope that their mind can be changed and that they will explicitly maintain /Wall support.

https://blogs.msdn.microsoft.com/vcblog/2017/02/06/stl-fixes-in-vs-2017-rtm/

horenmar:
While we could support any warning level just by force disabling it, that is not a good implementation strategy. Our users can do the same, with the advantage of them knowing that they disabled the warnings explicitly, instead of believing that they have the warnings enabled and are silent.

socantre:
You're only supposed to enforce the warning level internally, not on the client's code. So if the client enables a warning, it won't be silent for their code no matter what you do.

[...]

PR #722, opened 2016-10-08
Solved compiler warnings C5026 and C5027 for VS2015.

RLangendam:


RLangendam:
Possibly solves #387

lightmare:
IMO these two warnings are silly. If anything, they should be disabled.

RLangendam:
I agree that a compiler should only warn if something is done wrong; not when something could go wrong when used inappropriately. In general though, one should strive to satisfy the [rule of five](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)#Rule_of_Five).

horenmar:
This was solved a bit differently -- if `CATCH_CONFIG_CPP11_GENERATED_METHODS` is defined, it also `delete`s its move constructor and move assignment -- so I am going to cloes this soon unless someone still has the problem.

PR #723, opened 2016-10-10
Improved support for Oracle Solaris Studio 12.1 toolchain

afoster1:
Catch doesn't currently build with Oracle Solaris Studio 12.1 but the changes required to make it build are minor.  Three changes in total, two are trivial, one, sadly, is a little ugly due to the platforms non-compliant STL implementation for std::distance.

These issues probably affect other versions of the toolchain but i'm unable to verify so have tried constrain the "non-compliant STL" change just to this compiler.


horenmar:
Catch2 is dependent on C++11, so old versions of Solaris Studio are no longer even possible to support.

Issue #531, opened 2015-10-29
CHECK and REQUIRE can't expand a custom type (QString)


vpicaver:
I've implementing operator<< and Catch::toString for printing QString's in CHECK statements but nothing seems to work.  QString is a string class from the Qt framework.

```
#include "catch.hpp"

std::ostream& operator << ( std::ostream& os, QString const& value ) {
    os << value.toStdString();
    return os;
}

namespace Catch {
    std::string toString( QVariant const& value ) {
        return value.toString().toStdString();
    }
}

TEST_CASE("My testcase")
{
    CAPTURE(QString("moo"));
    INFO(QString("Sauce"));

    CHECK(QString("moo") == QString("sauce"));
```

This will prints out:

```
/Users/vpicaver/Documents/Projects/cavewhere/testcases/SurveyChunkTest.cpp:91: FAILED:
  CHECK( QString("moo") == QString("sauce") )
with expansion:
  {?} == {?}
with messages:
  QString("moo") := moo
  Sauce
```

CAPTURE and INFO work okay, but the expansion still returns `{?} == {?}`

I also try to add `#define CATCH_CONFIG_SFINAE` it didn't seem to help.

Is this a bug? Or am I doing something wrong?

nabijaczleweli:
```
#define CATCH_CONFIG_MAIN
#include "catch.hpp"

struct QString {
    std::string _;
    const std::string & toStdString() const { return _; }
    
    bool operator==(const QString & rhs) { return _ == rhs._; }
};

std::ostream& operator <<( std::ostream& os, QString const& value ) {
    os << value.toStdString();
    return os;
}

TEST_CASE("My testcase") {
  CAPTURE(QString{"moo"});
  INFO(QString{"Sauce"});

  CHECK(QString{"moo"} == QString{"sauce"});
}
```

Works for me

vpicaver:
This does work with your mock QString, but doesn't work with the real QString from the Qt library. QString implements a bunch of operator<< and I'm not sure if it's effecting catch.

nabijaczleweli:
Provide an [SSCCE](http://sscce.org/), because you're doing something quite fishy by the looks of it

vpicaver:
Sure no problem. What platform are you on? So I can provide instructions. Are you ever used the Qt framework?

nabijaczleweli:
The example should be platform-independent (I'm on Windows) Also, try *not* to use Qt to achieve the minimal code sample to reproduce the bug

[...]

PR #735, opened 2016-10-24
Fixed function signature in the code snippet

vosel:
Since the 'toString' is a template function specialization, it should
have a 'template<>' statement in it's declaration.

### DIFF
--- a/docs/tostring.md
+++ b/docs/tostring.md
@@ -33,7 +33,7 @@ If you don't want to provide an ```operator <<``` overload, or you want to conve
 
 ```
 namespace Catch {
-	std::string toString( T const& value ) {
+	template<> std::string toString( T const& value ) {
 		return convertMyTypeToString( value );
 	}
 }


lightmare:
The header says "overload" because it means overload, not "specialization". [One should not specialize function templates](http://www.gotw.ca/publications/mill17.htm).

It's not meant to suggest template, it's just documentation pseudo-code where `T` means "your type" (notice how the two `operator<<` examples also don't mention template). Of course if you have a class template, it's perfectly correct to define `template <typename U> std::string toString(YourType<U> const& x)`, which is a *set of overloads*; in this case the `T` in the doc substitutes `YourType<U>` in your actual code.

vosel:
Well, maybe then I misunderstood the situation. I created this pull request because the code from the documentation was not working correctly for me. But when I added the 'template<>' to the signature, it started to work correctly.

Here is the test case for which the behaviour is incorrect:

```c++
#include <string>
#define CATCH_CONFIG_MAIN
#include "catch.hpp"
struct MyClass {
	bool operator == (MyClass const & other) const {
		return false;
	}
};
namespace Catch {
	std::string toString(MyClass const & toDump) {
		return "custom presentation";
	}
}
TEST_CASE("test", "test") {
	MyClass c1;
	MyClass c2;
	REQUIRE(c1 == c2);
}
```

When compiled and executed (I used gcc 4.9.2 with `-std=c++14`, Catch v1.5.7.), the output is the following:

```
testcase.cpp:17: FAILED:
  REQUIRE( c1 == c2 )
with expansion:
  {?} == {?}
```

But when I replace the `std::string toString(MyClass const & toDump)` with `template<> std::string toString(MyClass const & toDump)`, the test output changes to this:
```
testcase.cpp:17: FAILED:
  REQUIRE( c1 == c2 )
with expansion:
  custom presentation == custom presentation
```

Notice the change in the expansion: from `{?} == {?}` to `custom presentation == custom presentation`.

So, I assumed, that there is a simple typo in the code snippet in documentation and created this fix.

lightmare:
Oh I see where you're coming from. This issue has been reported in #531 and the culprit is that a conforming compiler only considers overloads declared prior to the point of call. Just try moving the `#include "catch.hpp"` line right before the `TEST_CASE` line and you'll see your overload will be picked.

Now that I checked I found that even Catch itself is guilty of specializing `toString` for `Catch::Detail::Approx`, probably because the late-overload approach doesn't work.

I believe that the first thing that should be fixed is the implementation (i.e. call `toString` through ADL, unqualified) and then the documentation (i.e. put `toString(MyClass)` overload in the same namespace as MyClass).

horenmar:
Some time ago I added the fact that the overload has to happen before Catch is included to documentation, and because `Catch::toString` is currently deprecated and will be going away in favour of specializing `Catch::StringMaker` in Catch 2.0 (or providing `operator<<`), I am going to close this.

PR #736, opened 2016-10-28
Print out full section stacks on failure.

### LINKED ISSUE #734: Failure reporting should print out previous SECTIONs for data-driven testing
If a testcase differs only in the amount of input data, it would be nice to use `SECTION` to pass these bits into the object under test, like the following example. However, because the `REQUIRE` stanza is outside of the nested `SECTION`s, error messages do not contain references to the failing sections.

``` cpp
TEST_CASE("check that Foo detect invalid input") {
  std::string input;
  SECTION("no data") {
  }
  SECTION("garbage") {
    input = "xxx";
  }
  SECTION("just two words") {
    input = "foo bar";
  }
  Foo foo;
  foo.setLogLevel(666);
  REQUIRE(!foo.parse(input));
}
```

This can be worked around by adding additional macros, but it quickly gets ugly when the test involves more than few statements.

``` cpp
TEST_CASE("check that Foo detect invalid input") {
  #define DOIT(...) \
    Foo foo;\
    foo.setLogLevel(666); \
    REQUIRE(!foo.parse(__VA_ARGS__));

  SECTION("garbage") {
    DOIT("xxx")
  }
  SECTION("just two words") {
    DOIT("foo bar")
  }
}
```

I would like Catch to keep track of those `SECTION`s which have been already closed, and to add their names into the test output if expectations fail.

jruffin:
The Console reporter now "traces" the full section stack of its
current test execution and prints it out instead of the section
stack of the failing assertion itself. Likewise, the JUnit reporter
now groups assertions together into the testcase element of the test
execution where they happened.

This fixes #734 and allows post-condition tests after the deepest
SECTION in a test execution.

### DIFF
--- a/include/reporters/catch_reporter_bases.hpp
+++ b/include/reporters/catch_reporter_bases.hpp
@@ -43,10 +43,14 @@ namespace Catch {
         }
         virtual void sectionStarting( SectionInfo const& _sectionInfo ) CATCH_OVERRIDE {
             m_sectionStack.push_back( _sectionInfo );
+            m_deepestSectionStack.push_back( _sectionInfo );
         }
 
         virtual void sectionEnded( SectionStats const& /* _sectionStats */ ) CATCH_OVERRIDE {
             m_sectionStack.pop_back();
+            if (m_sectionStack.empty()) {
+                m_deepestSectionStack.clear();
+            }
         }
         virtual void testCaseEnded( TestCaseStats const& /* _testCaseStats */ ) CATCH_OVERRIDE {
             currentTestCaseInfo.reset();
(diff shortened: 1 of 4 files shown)


jktjkt:
This looks like what I've been looking for. Thanks!

ChrisThrasher:
Can we revisit this feature? In the case that a segfault happens within a given SECTION, it's useful to know what deepest SECTION was reached to get an idea for where to go looking for the problem. In such a case, I'm not sure how using generators as recommended [here](https://github.com/catchorg/Catch2/issues/734#issuecomment-485232118) would be helpful.

PR #743, opened 2016-11-12
Gato rat patch 1

GatoRat:
Adds parenthesis around expr in INTERNAL_CATCH_TEST to eliminate warnings in GCC 4.4.7


martinmoene:
This parenthesis 'issue' comes up regularly, but not having parenthsis there is at the basis of CATCH's expression decomposition machinery, see e.g. my comment [here](https://github.com/philsquared/Catch/issues/607#issuecomment-194259627).

GatoRat:
My mistake. I didn't look close enough and didn't test with failing tests.

PR #753, opened 2016-11-26
Fix for sigsegv stack overflow behavior

horenmar:
Also stops Catch from assuming its the only signal user in the binary,
and makes it restore the signal handlers it has replaced. Same goes for
the signal stack.

The signal stack itself probably shouldn't be always reallocated for
fragmentation reasons, but that can be fixed later on.

### DIFF
--- a/include/internal/catch_fatal_condition.hpp
+++ b/include/internal/catch_fatal_condition.hpp
@@ -37,8 +37,6 @@ namespace Catch {
 #else // Not Windows - assumed to be POSIX compatible //////////////////////////
 
 #include <signal.h>
-#include <memory>
-#include <array>
 
 namespace Catch {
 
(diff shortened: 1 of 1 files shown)


horenmar:
This takes unconditional dependency on C++11 (in the form of including `<array>` and `<memory>`), while catch tries for C++03 conformance. Going to fix this later tonight.

krvajalmiguelangel:
@philsquared any updates on this??

philsquared:
@krvajalmiguelangel - I started work on my own signals fixes before I saw @horenmar's PR - and have now covered half of it - will look at the stack overflow part soon.

horenmar:
I cherry picked this into `dev-signals` branch, so the PR is obsolete.

PR #756, opened 2016-11-29
fix implicit narrowing warnings in Catch

Rastaban:
Catch passes ::tolower into std::transform with string iterators.
::tolower has the signature int(int), which triggers a stealth narrowing
warning inside std::transform, because transform calls
*_Dest = _Fn(*_First), which implicitly narrows an int to a char.

For this particular application the narrowing is fine, so explicitly
narrow in an inline function.

This is identical to the issue that we (the MSVC C++ team) fixed in pull request #728,
just in a different place.

### DIFF
--- a/include/external/clara.h
+++ b/include/external/clara.h
@@ -401,7 +401,7 @@ namespace Clara {
         }
         inline void convertInto( std::string const& _source, bool& _dest ) {
             std::string sourceLC = _source;
-            std::transform( sourceLC.begin(), sourceLC.end(), sourceLC.begin(), tolowerCh );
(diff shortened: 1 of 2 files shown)


Rastaban:
The current travis test failures are not related.  After #754 is checked in I expect this to pass.

philsquared:
Thanks for reporting this - and for the PR.
I've actually addressed it slightly differently (made it self contained) - the reason being that this fix was in the code for Clara - which is a separate library (embedded, here, in Catch). Since I want the fix to work, as is, in the vanilla Clara code it can't rely on code elsewhere in Catch.

PR #759, opened 2016-12-02
Please pull compiler support for SunStudio 11 and SunStudio 12  

wiscy:
Dear philsquared,

I am writing tests using Catch for a legacy project. The necessary build Environment includes Sun OS Workstation and is using the compiler from SunStudio 12 and SunStudio 11. 

I already verified the changes using gcc at Linux workstations. Mac or Windows are not yet tested.

Please integrate the patch for SunStudio support into catch.

Best regards,
Wiscy


horenmar:
Catch2 requires at least c++11, which these versions of SunStudio don't support.

PR #761, opened 2016-12-16
Added signal handling under Windows.

horenmar:
Only some "signals" are handled under Windows, because Windows does not
use signals per-se and the mechanics are different. For now, we handle
sigsegv, stack overflow, div-by-zero and sigill. We can also meaningfully
add various floating point errors, but not sigterm and family, because
sigterm is not a structured exception under Windows.

There is also no catch-all, because that would also catch various
debugger-related exceptions, like EXCEPTION_BREAKPOINT.


horenmar:
I tried correcting an oversight I made and... yeah, I am just going to create new one.

Issue #762, opened 2016-12-16
Catch includes windows.h without disabling worst parts


horenmar:
Catch includes `windows.h` on Windows without defining `NOMINMAX`, meaning there is `min` and `max` macro in file where Catch defines main. This probably does not matter much in the usual use-case (where there is one file with
```cpp
#define CATCH_CONFIG_MAIN
#include "catch.hpp"
```
but for user supplied main this can be very annoying.

Catch probably should also define `WIN32_LEAN_AND_MEAN`, before including `windows.h`, so it doesn't bring in symbols it does not need, speeding up main compilation and again, not polluting source file with user supplied main needlessly.

horenmar:
This is in.

PR #763, opened 2016-12-16
Added NOMINMAX and WIN32_LEAN_AND_MEAN defines before including windows.sh

horenmar:
This stops the `windows.h` header from defining `min` and `max` macros
and including lot of Windows APIs that are not needed by Catch.

Addresses #762.


philsquared:
Seems like a no-brainer - thanks.
However I have one concern: If `NOMINMAX` or `WIN32_LEAN_AND_MEAN` are already `#define`d then this will fail or warn - then (if compilation continues) leave those identifiers always undefined

horenmar:
So guarding the define like this
```cpp
#if !defined(NOMINMAX)
#define NOMINMAX
#define CATCH_DEFINED_NOMINMAX
#endif
... same for WIN32_LEAN_AND_MEAN
#include <windows.h>
#ifdef CATCH_DEFINED_NOMINMAX
#undef NOMINMAX
#endif
```
would be prefered? (Obviously at that point the logic would be factored out to `catch_compiler_capabilities.h`)

If we go this way, we could also provide a configuration option to stop Catch from including `<windows.h>` with these macros defined, but I think that is quite far past the point of diminishing returns and user who wants to include `<windows.h>` with all symbols in the same file as Catch main/runner, should just include it before including catch.

philsquared:
That's one way to do it. I did wonder if it worked out easier to do the cross product of possibilities (so you don't need the intermediate identifiers) - but probably not - and certainly not scalable.
So, yes, if you go with your suggestion I think that should cover things sufficiently - not need to overthink it further.

horenmar:
I rebased and changed the PR, but the results are not pretty.

    #  ifdef CATCH_DEFINES_NOMINMAX
    #    define NOMINMAX
    #  endif
    #  ifdef CATCH_DEFINES_WIN32_LEAN_AND_MEAN
    #    define WIN32_LEAN_AND_MEAN
    #  endif
    #include <windows.h>
    #  ifdef CATCH_DEFINES_NOMINMAX
    #    undef NOMINMAX
    #  endif
    #  ifdef CATCH_DEFINES_WIN32_LEAN_AND_MEAN
    #    undef WIN32_LEAN_AND_MEAN
    #  endif
    
To the best of my knowledge, only way to fix this would be to create a proxy header, that hides away the preprocessor directives.

PR #764, opened 2016-12-16
Added signal handling under Windows.

horenmar:
Only some "signals" are handled under Windows, because Windows does not
use signals per-se and the mechanics are different. For now, we handle
sigsegv, stack overflow, div-by-zero and sigill. We can also
meaningfully
add various floating point errors, but not sigterm and family, because
sigterm is not a structured exception under Windows.

There is also no catch-all, because that would also catch various
debugger-related exceptions, like `EXCEPTION_BREAKPOINT`.

### DIFF
--- a/include/internal/catch_fatal_condition.hpp
+++ b/include/internal/catch_fatal_condition.hpp
@@ -28,9 +28,65 @@ namespace Catch {
 
 namespace Catch {
 
+#define NOMINMAX
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#undef WIN32_LEAN_AND_MEAN
+#undef NOMINMAX
+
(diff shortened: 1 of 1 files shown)


tufak:
There are includes inside a namespace -- looks like a merge mishap to me.

horenmar:
Fixed. It was an artifact of me developing in single-file include and then merging the changes back, badly.

I wanted to see if a Windows CI could be set up and as far as I can see, Travis-CI still has no Windows support.  However, AppVeyor does support Windows and is free for Open Source projects, so that might be interesting to look into.

horenmar:
I cherry picked this into `dev-signals` branch, so the PR is obsolete.

PR #767, opened 2016-12-26
Do not encode extended ASCII characters in XML reporter

hmich:
At the moment Catch encodes (assuming signed char) characters from extended ASCII encodings in the XML reporter. This was probably not intended (since only control characters should be escaped), and is also a problem for XML parsers which do not understand XML v1.1 yet (like .NET's XmlReader).


philsquared:
Thanks @hmich

PR #772, opened 2017-01-09
stringify expressions even more lazily

lightmare:
Alternative to #562 

Using the test at the top of #556, with only 1 million items and a `vector::reserve` to avoid reallocations I get the following timings:
before: 4.2s
after: 1.3s

### DIFF
--- a/include/internal/catch_assertionresult.h
+++ b/include/internal/catch_assertionresult.h
@@ -13,6 +13,27 @@
 
 namespace Catch {
 
+    struct STATIC_ASSERT_Expression_Too_Complex_Please_Rewrite_As_Binary_Comparison;
+
+    struct DecomposedExpression
+    {
+        virtual ~DecomposedExpression() {}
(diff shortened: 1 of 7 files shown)


horenmar:
Just a quick note, this isn't being ignored, I am just insanely busy until weekend.

horenmar:
I started work at performance related changes in a [dev-performance branch](philsquared/Catch/tree/dev-performance). If you have some representative benchmarks, please contribute.

As far as this PR, it seems like a large win for passing tests and extremely small penalty for non-passing tests, so its probably going in. (I will have to run rest of tests first)

lightmare:
I haven't done any other benchmarks beyond the one in #556. I haven't even thought about failure perf, that's a good addition you made.

horenmar:
This is now in, but as far as I can tell there is no way to mark a PR as merged, if I cherry-picked it into a branch and then merged to master.

Issue #582, opened 2016-01-30
non-throwing CHECK


lightmare:
I'm looking for a variant of `CHECK(expr)` that doesn't abort if `expr` throws an exception.

For example:

``` c++
TEST_CASE("string access")
{
    std::string s("foo");
    CHECK(s.at(3) == 'o'); // throws
    CHECK(s.at(2) == 'o'); // ok
    CHECK(s.at(1) == 'f'); // fails
}
```

With `CHECK` only the first (throwing) one is reported, because `Catch::ResultDisposition::Normal` is used [here](https://github.com/philsquared/Catch/blob/7424b23bfb4423fd1213141633844f09faaff8bc/include/internal/catch_capture.hpp#L39). What I'd like to see in the output is:

```
s.at(3) == 'o' FAILED with exception
s.at(1) == 'f' FAILED with 'o' == 'f'
```

I implemented that by copy&pasting `INTERNAL_CATCH_TEST` and using `resultDisposition` in the catch clause ([ref](https://github.com/lightmare/mapnik/blob/1d34d7fd201ce65bd53fa935cadfcec1d6998f99/test/catch_ext.hpp#L15)). I'm not comfortable with the name, should be something like `NORETHROW_CHECK` or `NOABORT_CHECK` or perhaps default behaviour of `CHECK`?

horenmar:
Honestly, this seems like a bug in the internal macro, I can't see a reason not to continue on exception when `CHECK`ing.

horenmar:
This is now in master, from 184865358c80ed37f6d9dadb6f80207482c95ab4.

PR #782, opened 2017-01-15
CHECK no longer stops running SECTION on exception.

horenmar:
This seems to be much closer to desired behaviour than the current one, where
CHECK(false) lets the execution continue, but CHECK(throws) does not.


horenmar:
This change is related to #582

horenmar:
In master now, as 184865358c80ed37f6d9dadb6f80207482c95ab4.

PR #784, opened 2017-01-19
simplify output filtering in approvalTests.py

lightmare:
This just makes the filtering function more concise by using `re.sub` instead of checking for matches and gluing the pieces afterwards. Results should be identical.


horenmar:
I have some work in progress on porting approvalTests script to Windows, so I'll pull this into a branch and merge all the changes first.

Issue #574, opened 2016-01-20
Expression is evaluated twice when operator&&(bool, decltype(expr)) is defined


hpesoj:
```
} while( Catch::isTrue( false && (expr) ) )
```

This line assumes the built-in `&&` is used. If a user-defined `operator&&` is used, the expression is not short-circuited, and `expr` is evaluated here (and twice overall). I know that overloading `&&` is not generally recommended, but I think this corner-case can be fixed by explicitly casting `expr` to `bool`:

```
} while( Catch::isTrue( false && static_cast<bool>(expr) ) )
```

philsquared:
I never got back to you about this.
I implemented your suggested fix - but it turns out that has an unwanted side-effect (see #657).
I'm considering using `!!(expr)` instead. That seems to do the trick for me but I'd appreciate it you were able to confirm that it has the desired effect for you, too?

philsquared:
Actually I'm fairly confident that this should work for you. I've pushed a commit with the change in.
Please do let me know if it causes any problems.

hpesoj:
Hi Phil. I suggested the original fix for compatibility with a custom
boolean type which overloaded the && operator. Said type also overloaded
the ! operator, so your proposed change would reintroduce the
incompatibility for such types. I believe keeping both the !! and the
static cast would fix both issues, though I am not 100% sure if other
issues wouldn't arise.

However, I have since removed the && operator overload for my type as I
feel short circuiting evaluation is too useful a feature to sacrifice.
Thus, your proposed change would no longer affect me personally. Perhaps
you feel that it is not worth attempting to support such a corner case?

philsquared:
The problem with the `&&` overload was that short circuiting was disabled - and the whole point of this expression was for it to be seen by the compiler (so you get natural warnings etc) without being evaluated at runtime.
Using `!!` puts us back into the realm of short-circuiting, so the original expression should not be evaluated at runtime.
At that point as long as the `!` operator compiles for your type (and, ideally, doesn't generate any warnings) then it shouldn't matter what it actually does - it cannot influence short-circuiting here so should never be executed.

I put together a little test case for this which reproduced your original issue when I didn't use the cast or the `!!` - but had no issues (with `&&` or `!`) with the `!!` approach.

So I'm inclined to leave things as they are here. If anyone else hits problems with it I'll consider reverting back to the original, though.

Thanks for taking the time to respond - especially given that the original issue no longer replies to you!

hpesoj:
[...]

PR #785, opened 2017-01-19
use sizeof(expr) for unevaluated syntax check

lightmare:
There were numerous different ideas and scattered code snippets in #574. This PR gathers the pieces to implement the `sizeof` approach.


horenmar:
I like this better than the current solution.

philsquared:
me too ;-)

PR #788, opened 2017-01-21
Added INSTALL commands

samaursa:
Even though CATCH is a header only library, it makes sense to allow it to INSTALL the includes in a standard manner. This is especially useful when adding dependencies using the CMake ExternalProject_Add command as an example.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,3 +237,6 @@ set_tests_properties(ListTests PROPERTIES PASS_REGULAR_EXPRESSION "[0-9]+ test c
 
 add_test(NAME ListTags COMMAND SelfTest --list-tags)
 set_tests_properties(ListTags PROPERTIES PASS_REGULAR_EXPRESSION "[0-9]+ tags")
+
+install(DIRECTORY "include/" DESTINATION "include/catch")
+install(DIRECTORY "single_include/" DESTINATION "include/catch/single_include/")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,5 +238,4 @@ set_tests_properties(ListTests PROPERTIES PASS_REGULAR_EXPRESSION "[0-9]+ test c
 add_test(NAME ListTags COMMAND SelfTest --list-tags)
 set_tests_properties(ListTags PROPERTIES PASS_REGULAR_EXPRESSION "[0-9]+ tags")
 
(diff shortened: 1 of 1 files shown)


knedlsepp:
I also find this is a good idea, but I'd rather not install both the single include version and the full one. Also I'd prefer to install into `include` instead of `include/catch/`.

samaursa:
The reason I wrote `include/catch` is to stay consistent with all the other libraries out there (currently, I am depending on `bgfx`, `eigen`, `glfw` and `spdlog` and all of them are using this scheme).

knedlsepp:
Well prefixing it with `include/catch` would mean, that the end user would have to use it via:
```
#include <catch/catch.hpp>
```

samaursa:
That's how it should be. 

```
#include <bgfx/bgfx.h>
#include <catch/catch.hpp>
```

If a user doesn't like that, they have the choice of adding the `catch` folder in their include directories. The opposite is more difficult.

knedlsepp:
Well. I disagree on the "should be" part. For lots of small libraries often times they opt for: 
```C++
#include <catch.hpp>
```
This is certainly a matter of taste, which we should let the owner decide.

samaursa:
It gives me a chance, as a user,  to resolve ambiguities. This is like putting your functions/classes in a namespace. If I have a header `catch.hpp` (which is entirely possible, `catch` is not a unique name) then there is an ambiguity that I simply cannot resolve without moving around files.

On the other hand, if I have `#include <catch/catch.hpp>` but say you don't like the extra folder, you can easily ask `CMake` to `include_directories` the `catch` dir as well. The opposite is not so easy.

Can we work around it? Yes, of course :) C++ libraries are notorious for being inconsistent, at least we have CMake now!

Having said that, I agree - decision is with @philsquared .

horenmar:
This is going to need some discussion, both whether we should install the `single_include` version or the normal one (I say `single_include`, because it corresponds to last tagged release and not just master), and whether we should install as `catch/catch.hpp` or just `catch.hpp`. Personally I am for keeping it as `catch.hpp` but it's not a strong preference.

horenmar:
After discussion with Phil, we decided to have the command install the single-include header (because that remains stable between releases) and to have it namespaced as `catch`, so it installs as `catch/catch.hpp`. The second is subject to change given feedback.

PR #789, opened 2017-01-23
Enable breaking into debugger on Mac

knedlsepp:
The integrated assembler segment was missing an underscore:
`"_asm__"`. This regression was accidentally introduced in: https://github.com/philsquared/Catch/commit/25d017763b73e09d3e9adf7f4db9ce71992f0471#diff-96e75ad10f6edace03f2cb7d3607de00L34
Also we remove the "DEBUG" macro check, so we are consistent
with the linux and windows variant.
Now breaking into gdb on failure should work via:
`gdb --args test_executable --break`

I tested this to work on Mac OS 10.11.6 using GNU gdb (GDB) 7.12.

### DIFF
--- a/include/internal/catch_debugger.h
+++ b/include/internal/catch_debugger.h
@@ -23,14 +23,12 @@ namespace Catch{
 
     // The following code snippet based on:
     // http://cocoawithlove.com/2008/03/break-into-debugger.html
-    #ifdef DEBUG
-        #if defined(__ppc64__) || defined(__ppc__)
-            #define CATCH_TRAP() \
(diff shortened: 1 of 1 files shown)


horenmar:
This looks good, but as it is on mac I am unable to test it.

philsquared:
DEBUG macro was there because it was in the snippet I based the original implementation off (http://www.cocoawithlove.com/2008/03/break-into-debugger.html).
I agree with the arguments made, elsewhere, that it shouldn't be there, though - so I've accepted this PR in it's entirety. Thanks for that.

Looks like the missing underscore crept in during a recent merge. I've verified that it all works again now on my machine.

PR #791, opened 2017-01-26
Compilation warnings fix

ravstrix:


horenmar:
The one in `contains` is not just a compilation warning, its an actual bug that needs fixing (although I don't think Catch uses `contains` anywhere, I only added it because it felt weird adding single char overloads for `startsWith` and `endsWith` and not for `contains` as well).

PR #797, opened 2017-01-30
Cap main exit code to 255

fcelda:
## Description

According to POSIX, only lower 8-bits of a program exit code are available through `wait()` and `waitpid()` calls. Although there is a newer `waitid()` interface, I think most of the programs still use the old one. As a result, multiples of 256 wrap to 0 for these programs.

This pull request sets cap to 255 for the default main exit code for better compatiblity. Also the documentation for own-main() is updated.

## Reproducer

Tested on macOS Sierra with zsh and Fedora 25 with bash.

```c++
#define CATCH_CONFIG_MAIN
#include <catch.hpp>

TEST_CASE("256 failures")
{
	for (int i = 0; i < 256; i++) {
		CHECK(false);
	}
}
```

```
$ ./before; echo "exit_code: $?"
...
test cases:   1 |   1 failed
assertions: 256 | 256 failed

exit_code: 0
```

```
$ ./after; echo "exit_code: $?"
...
test cases:   1 |   1 failed
assertions: 256 | 256 failed

exit_code: 255
```

### DIFF
--- a/docs/own-main.md
+++ b/docs/own-main.md
@@ -45,13 +45,16 @@ int main( int argc, char* argv[] )
 
   int returnCode = session.applyCommandLine( argc, argv );
   if( returnCode != 0 ) // Indicates a command line error
-  	return ( returnCode < 0xff ? returnCode : 0xff );
+  	return returnCode;
 
   // writing to session.configData() or session.Config() here 
   // overrides command line args
   // only do this if you know you need to
 
   int numFailed = session.run();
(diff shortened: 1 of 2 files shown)


horenmar:
Ok, this is interesting and why we can't have nice things.

I am going to make some more documentation changes before accepting though, the clamp on return codes from `applyCommandLine` is not needed -- it returns either 0 or INT_MAX.

fcelda:
POSIX is one huge trap. :) Thank you, Martin!

PR #799, opened 2017-02-02
Fix single-header download link in README

mmcshane:
I think you may have made a typo in creating the git tag for for the 1.7.0 release - there is an extra dot between the v and the 1.  This fixes the download link to incorporate the extra dot but you may want to consider re-tagging the release.


horenmar:
I am going to retag the release instead. Thanks for the notice though.

PR #826, opened 2017-02-20
Add reporter for Automake

jrw972:
## Description

Added a reporter to generate the :test-result: lines expected by Automake's default log compiler.  This will allow projects using Automake to integrate Catch tests.  Users of Automake will need to use an appropriate LOG_DRIVER.  A prototype of the  log driver would be something like
[catch-driver.sh.txt](https://github.com/philsquared/Catch/files/786503/catch-driver.sh.txt)


horenmar:
We don't want to add more reporters to the single-include header, so I am going to fix-up the PR a bit.

Otherwise looks pretty good.

PR #828, opened 2017-02-21
Add “slim” reporter

ronen:
## Description

This is a simple modification to the `compact` reporter, to include the test name (and sections) on each line, i.e.:

```
<filename>:##: <Status>: <TestName & Sections>: <Assertion>    // slim reporter (this PR)
<filename>:##: <Status>: <Assertion>                           // compact reporter
```

I run tests on the command line and visually scan the result, so I like having just one line per test as the `compact` reporter provides.   But, at least in my case, a test file often has a collection of tests with different setups but similar assertions, so including the test name helps me know what's what.

It would probably be better to have less repeated code between `slim` and `compact`.  One way would be to have just one reporter and add a mechanism to pass parameters to control the output.  But that would be a bigger change.   Simpler would be to make a shared base class, or even have `CompactReporter` inherit from `SlimReporter` but disable printing the name.  But at least for this initial PR I just wanted to add new code and not change (i.e. potentially break) any existing code.

FYI this PR also adds an additional intentional name to the colors:  `Colour::TestName = Blue`


horenmar:
Not sure if this change is worth a separate reporter, and it shouldn't be included in the single-include by default.

Maybe there is interest in changing the compact reporter to work like this instead?

ronen:
> Not sure if this change is worth a separate reporter, and it shouldn't be included in the single-include by default.

Not my call :)

> Maybe there is interest in changing the compact reporter to work like this instead?

That would be fine by me.   I didn't submit this PR as a change to the compact reporter because I don't know whether others depend on its current output format.  In particular the compact reporter describes itself as intended for use with IDEs, and so changing the output format might break any existing IDE integration?

ronen:
> Maybe there is interest in changing the compact reporter to work like this instead?

I went ahead and submitted #829 which the changes compact reporter.

ronen:
FYI I've merged the latest master into this branch, fixing the (trivial) merge conflicts that had arisen.    Would still love it if this PR or #829 could be merged.   Thanks!

PR #829, opened 2017-02-23
Add test name & sections to compact report’s output

ronen:
## Description

Modifies the compact reporter so that its output includes the test name and sections:

```
<filename>:##: <Status>: <TestName & Sections>: <Assertion>    // after this PR
<filename>:##: <Status>: <Assertion>                           // before this PR
```

I run tests on the command line and visually scan the result, so I like having just one line per test as the compact reporter provides. But, at least in my case, a test file often has a collection of tests with different setups but similar assertions, so including the test name helps me know what's what.

This PR also adds an additional intentional name to the colors: `Colour::TestName = Blue`

**NOTE**
Submitting this PR as an alternative to #828 which provides this output via a separate reporter.  This PR is simpler, but may be problematic if it's objectionable to change the output of the compact reporter (if, e.g., some IDE integrations depend on the current output format).


ronen:
FYI I've merged the latest master into this branch, fixing the (trivial) merge conflicts that had arisen.    Would still love it if this PR or #828 could be merged.   Thanks!

PR #832, opened 2017-02-27
Make INFO() use the same output customization as CHECK()

rmartinho:
## Description

This changes MessageBuilder so that `INFO(x << y)` has output similar to `CHECK(x == y)`. I'm not sure if this was an intentional omission, but since I make heavy use of `StringMaker`, I'm really happy with this change :)

### DIFF
--- a/include/internal/catch_message.h
+++ b/include/internal/catch_message.h
@@ -11,6 +11,7 @@
 #include <string>
 #include "catch_result_type.h"
 #include "catch_common.h"
+#include "catch_tostring.h"
 
 namespace Catch {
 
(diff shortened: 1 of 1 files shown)


horenmar:
I remember talking about this with Phil, and IIRC it is intentional. 

That does not mean it can't change, but it will need some discussion first.

horenmar:
Well, this has been a while, but I finally got around to this.

The result is that this won't be merged, because making the change in this way would lead to undesirable cosmetic changes.

Consider `INFO` usage from MiscTests.cpp:
```cpp
INFO( "Testing if fib[" << i << "] (" << fib[i] << ") is even" );
```
the current output is `Testing if fib[7] (21) is even`, the changed output would be `"Testing if fib["7"] ("21") is even"`.

It would also change how `CAPTURE` is stringified
```cpp
int i = 42;
CAPTURE(i);
```
would result in `"i := "42`

PR #833, opened 2017-02-27
Add missing typename

rmartinho:
## Description

I found a couple of missing `typename` while compiling a test within a template. This adds them.

### DIFF
--- a/include/internal/catch_capture.hpp
+++ b/include/internal/catch_capture.hpp
@@ -104,7 +104,7 @@
                 static_cast<void>(expr); \
                 __catchResult.captureResult( Catch::ResultWas::DidntThrowException ); \
             } \
-            catch( Catch::add_const<Catch::add_lvalue_reference<exceptionType>::type>::type ) { \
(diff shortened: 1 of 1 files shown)


horenmar:
This also breaks old gcc

lightmare:
It also doesn't do what you think it does. `add_const` is kinda pointless anywhere, as it's just a tedious way of writing `T const`. And in this snippet, it doesn't do anything, because `add_const<T&>` returns `T&`.

horenmar:
@lightmare Thats not actually completely true, having duplicate const specifier usually triggers a warning, which `add_const` prevents.

In general, tests in templates are going to need some work -- as an example, `_Pragma` based warning supression does not work inside templates, so pretty much every non-trivial `REQUIRE` triggers `Wparentheses`.

horenmar:
@rmartinho I slept on this a bit and maybe we could revert the `catch (...)` block back, so that it uses the given type exactly once again, because using `typename` outside of template is C++11 extension and Catch Classic is still C++98.

horenmar:
I've reverted the change, `*_THROWS_AS` macros now use the exception type argument as given.

Issue #815, opened 2017-02-07
Add SET_UP and TEAR_DOWN or similar macros to the Catch framework


mKlus:
Would it be possible to add a global set up and tear down macros to the Catch?

Today this is possible to do in the custom main, but it would be much more useful to do it in a similar way as we write TEST_CASES so I can do CHECKS and REQUIRE in the SET_UP. 
The advantage is that if SET_UP fails then 1. you get a report of that in the reporter output, 2. test cases would not run in that case.

philsquared:
You can also do this using a listener object (which is basically a type of reporter that you can use to hook the same events - but runs in addition to whatever reporter you're using).

The downside (other than that it's not yet documented) is that it still doesn't give you the ability to report on failures during set-up. It sounds like that may be your motivating feature?

I'll consider adding that.

mKlus:
When I wrote the SET_UP and TEAR_DOWN macros in our version of catch my motivation was to:
- make it easier for developers to setup their tests (no need to write their own main)
- report failures during setup and don't runt any test if setup fails.

Thanks for considering this. I could share my code or send you PR but it's not really great design so I have a feeling that you'll write it in completely different way.

textshell:
I think allowing event listeners to abort test execution would be great some use-cases.
This can be useful for setup / teardown or for checking test environment invariants etc.

PR #837, opened 2017-03-01
Added SET_UP and TEAR_DOWN macros.

mKlus:
This is PR to add support for SET_UP and TEAR_DOWN macros requested in #815.
Probably needs more work and may need to be changed completely but just wanted to share my quick implementation of it.


horenmar:
Catch 1.X is meant to support old (some might say obsolete) compilers compiling C++98, so for things like `nullptr`, you should be using `CATCH_NULL`. Adding tests to the SelfTest project would also be appreciated.

The above isn't sufficient for merge, I'll leave that up to Phil, but at least compilation needs to be fixed.

PR #838, opened 2017-03-01
TAP Reporter: count success tests even if not printed

gahr:
## Description

This PR fixes an issue within the TAP reporter that causes it not to count success tests in absence of the `-s` switch, resulting in it always printing `1..0`.


horenmar:
I admit I don't actually __use__ TAP protocol anywhere, so I wasn't sure whether it should or should not report all of the tests in summary.

However, my reading of the specs, such as they are, implies that if we report `1..10` tests, we should also print out the `ok <num>` lines (basically, TAP reporter should behave as if `-s` was always specified). Can you confirm/deny?

gahr:
Yes, I think that is correct. I'm not a TAP user myself, I was just trying out the feature since I saw that on the release notes..

PR #842, opened 2017-03-04
Update catch.hpp

Tocic:
add const ref to function parameters


horenmar:
> Please do not submit pull requests changing the `version.hpp`
or the single-include `catch.hpp` file, these are changed
only when a new release is made.

...

There are actually more places where reporters take the string by value instead of by-reference, I'll take a look and fix them up tomorrow.

PR #845, opened 2017-03-06
teamcity reporter should time durations explicitly

awglyde:
## Description
I've modified the teamcity reporter to time individual test cases. TeamCity can time tests without reporting duration explicitly however the timing can be subject to network latency.

See the teamcity [documentation](https://confluence.jetbrains.com/display/TCD10/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-Supportedtestservicemessages) for the duration attribute on the testFinished service message.

### DIFF
--- a/include/reporters/catch_reporter_teamcity.hpp
+++ b/include/reporters/catch_reporter_teamcity.hpp
@@ -148,7 +148,7 @@ namespace Catch {
         }
 
         virtual void testCaseStarting( TestCaseInfo const& testInfo ) CATCH_OVERRIDE {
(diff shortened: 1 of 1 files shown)


horenmar:
First look: spaces, not tabs please.

As to whether this gets merged I'll leave to @philsquared who uses TeamCity and understands it much better.

SomeWritesReserved:
This would be helpful to me as well. I think TeamCity might keep track of timing itself but I can't be sure. Either way, having Catch do the timing itself (which it already does) would be better.

Issue #565, opened 2015-12-28
Warnings about REQUIRE when used in templates


rr-:
Consider this:

```
template<class T> inline void test_reading_multiple_bytes(const TestType type)
{
    SECTION("...")
    {
        T reader("\x8F\x8F"_b); // 10001111
        if (type == TestType::Msb)
        {
            REQUIRE((reader.get(7) == (0x8F >> 1)));
            REQUIRE(reader.get(1));
            REQUIRE(reader.get(1));
            REQUIRE(!reader.get(1));
            REQUIRE((reader.get(4) == 3));
            REQUIRE((reader.get(2) == 3));
        }
        else
        {
            REQUIRE(reader.get(1));
            REQUIRE((reader.get(7) == (0x8F >> 1)));
            REQUIRE(reader.get(1));
            REQUIRE(reader.get(1));
            REQUIRE(reader.get(2) == 3);
            REQUIRE(reader.get(4) == 8);
        }
    }
}

TEST_CASE("...")
{
    test_reading_multiple_bytes<MsbBitReader>(TestType::Msb);
    test_reading_multiple_bytes<LsbBitReader>(TestType::Lsb);
}
```

This produces warnings with gcc 5.3 on GNU/Linux:

```
/home/rr-/src/arc_unpacker/tests/io/bit_reader_test.cc: In instantiation of ‘void test_reading_multiple_bytes({anonymous}::TestType) [with T = au::io::OldBitReader]’:
/home/rr-/src/arc_unpacker/tests/io/bit_reader_test.cc:277:64:   required from here
/home/rr-/src/arc_unpacker/tests/io/bit_reader_test.cc:89:39: warning: suggest parentheses around comparison in operand of ‘==’ [-Wparentheses]
             REQUIRE(reader.get(2) == 3);
                                   ^
/home/rr-/src/arc_unpacker/tests/io/bit_reader_test.cc:90:39: warning: suggest parentheses around comparison in operand of ‘==’ [-Wparentheses]
             REQUIRE(reader.get(4) == 8);
```

When I wrap with `(` `)` like gcc asks me to, it stops complaining, although on top of code becoming ugly, I lose information on the reason behind failed tests (i.e. rather than being shown useful `3 == 8`, I am shown `false` only - not useful much).

rr-:
...and in case you wonder - the reason behind templated tests is basically rolling my own solution to #357.

horenmar:
[...]

PR #848, opened 2017-03-09
Add parentheses around macro argument "expr"

Viatorus:
Without paranteses arround the macro argument "expr" the used overloaded operator, the execution order and the resulting type is not clear for all macro substitutions:

Example:

```
expr = a || b:

// Bad:
(__catchResult <= expr ) -> (__catchResult <= a || b )

// Good:
(__catchResult <= (expr) ) -> (__catchResult <= (a || b) )
```

On GCC I get the warning without parentheses: 
warning: suggest parentheses around comparison in operand of ‘==’ [-Wparentheses]

### DIFF
--- a/include/internal/catch_capture.hpp
+++ b/include/internal/catch_capture.hpp
@@ -44,7 +44,7 @@
         Catch::ResultBuilder __catchResult( macroName, CATCH_INTERNAL_LINEINFO, #expr, resultDisposition ); \
         try { \
             CATCH_INTERNAL_SUPPRESS_PARENTHESES_WARNINGS \
-            ( __catchResult <= expr ).endExpression(); \
+            ( __catchResult <= (expr) ).endExpression(); \
             CATCH_INTERNAL_UNSUPPRESS_PARENTHESES_WARNINGS \
         } \
         catch( ... ) { \


philsquared:
Could you give an example of what `a` and `b` might be?
Catch doesn't support composing expressions that like in `REQUIRE` or `CHECK` assertions (Matchers can be composed, though).
Adding parentheses around the expression would defeat the expression decomposition logic that allows lhs and rhs values to be reported independently.
You can, of course, add them in yourself at point of use, e.g.:

`REQUIRE( (a || b) );`

(But, again, you will lose the fine grained reporting).

Viatorus:
I am not quite sure what you mean with:
> you will lose the fine grained reporting

What do you lose exactly when you put a parentheses around `expr`?

I will leave an example asap.

philsquared:
```c++
int a = 6*9;
REQUIRE( a == theAnswer() );
```
The failure output will tell you that the expression expanded to `54 == 42`.
```c++
REQUIRE( ( a == theAnswer() ) );
```
With the extra parentheses the failure output will tell you that the expression expanded to `false`.

Viatorus:
Oh, sorry - you are absolutely right, I didn't see it. So my fix is totally wrong. Sorry for this!

But the warning still exists:

```
#define CATCH_CONFIG_MAIN
#include "catch.hpp"

template<typename T>
void test1() {
    CHECK(T{} == false);
}

template<typename T>
void test2() {
    T v{};
    CHECK(v == false);
}

template<typename T>
void test3() {
    T v{};
    CHECK(v == v);
}

void test4() {
    using T = bool;
    T v{};
    CHECK(v == v);
}

void test5() {
    bool v{};
    CHECK(v == v);
}

TEST_CASE("Test leaf with scalar types") {
    test1<bool>();
    test2<bool>();
    test3<bool>();
    test4();
    test5();
};
```
test1, test2 and test3 raise a warning. And this is very annoying because warnings are errors for us...

Please see:
http://melpon.org/wandbox/permlink/VR75fS8EC905uloQ

Seems to be a gcc bug... or?

horenmar:
Yeah, there is a problem with warning suppression inside templated tests, that I am not quite sure how to fix.

Basically, if you are on a new-enough gcc (4.8+), then we use suppression based on `_Pragma`, which lets us supress this warning for a single line, leaving it intact for the rest of the test file. For some reason, this does not work inside templates and I haven't gotten around investigating whether this is a gcc or standard bug (weirdness).

A solution is to do what we do for old gccs, just use pragma to disable the warning for the whole file.

philsquared:
I'm going to close this PR as we've established it is not applicable.
The remaining issue is being tracked in #565 (although I'm not sure what, if anything, we can do about that).

PR #855, opened 2017-03-14
Fix Catch::toString() linker error when main() is compiled as C++

### LINKED ISSUE #278: Catch build 41 link error for ObjC++
I have a very small .mm file that does a few tests with CATCH() macros.

I just updated to build 41 and now my test binary fails to link with:

Undefined symbols for architecture x86_64:
  "Catch::toString(NSString\* const&)", referenced from:
      Catch::ExpressionResultBuilder& Catch::ExpressionLhs<long const&>::captureExpression<(Catch::Internal::Operator)1, NSString_>(NSString_ const&)

I was using build 33 previously.
Xcode 5.1.1

I checked the header and it seems the toString() on line 7165 should be the implementation needed. So I'm not sure what's going on.

The only tests that I have where an NSString\* is a direct arg are like the following:

NSString *string = ...
CATCH(nil != string);

kainjow:
Closes #278

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -63,7 +63,7 @@ std::string toString( std::nullptr_t );
 
 #ifdef __OBJC__
     std::string toString( NSString const * const& nsstring );
-    std::string toString( NSString * CATCH_ARC_STRONG const& nsstring );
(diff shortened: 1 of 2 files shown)


philsquared:
Thanks - sorry for the delay in getting to this.

PR #856, opened 2017-03-15
Set execute permission on runnable python scripts

lightmare:
## Description

Python scripts got a shebang some time ago, but some were missing execute permission.


philsquared:
Thanks - been meaning to get to that :-)

PR #857, opened 2017-03-15
Use bulk vector::insert rather than back_inserter

lightmare:
## Description

This replaces 3 instances of
```c++
std::copy(first, last, std::back_inserter(dest_vec));
```
with
```
dest_vec.insert(dest_vec.end(), first, last);
```

It's not that I measured some difference in either compilation or run-time speed, I just prefer using a member function (which is almost guaranteed to run faster for more than 1 item) over `*inserter`. I also think having the modified thing on the left helps readability.

As for `#includes`, `<iterator>` provided the `back_inserter` so it's not needed anymore.
`catch_test_case_tracker.hpp` included `<map>` for no reason, so I removed that, and on the other hand didn't include `<algorithm>` while using `std::find_if`, so I added that.

### DIFF
--- a/include/internal/catch_generators.hpp
+++ b/include/internal/catch_generators.hpp
@@ -10,7 +10,6 @@
 
 #include "catch_context.h"
 
-#include <iterator>
 #include <vector>
 #include <string>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


philsquared:
Merged via 4dc06bdb7034d36071caa2219f14c4b2ce52c80a.
Thanks for that!

PR #876, opened 2017-03-30
print messages when unexpected exceptions are thrown

### LINKED ISSUE #748: captures with unexpected exceptions
Is it possible to make CAPTURE print its value when an unexpected exception occurs?
Currently, for the following code

```c++
#define CATCH_CONFIG_MAIN
#include <catch.hpp>
#include <stdexcept>

void foo()
{
    throw std::logic_error("error");
}

TEST_CASE("test")
{
    int i = 2;
    CAPTURE(i);
    foo();
    REQUIRE(false);
}
```

I get 
```cmd
Test.exe is a Catch v1.1 b3 host application.
Run with -? for options

test
\giant_nas\dviry\my documents\visual studio 2015\projects\test\test\source.cpp(1
0)
...............................................................................

\giant_nas\dviry\my documents\visual studio 2015\projects\test\test\source.cpp(1
0): FAILED:
due to unexpected exception with message:
  error

test cases: 1 | 1 failed
assertions: 1 | 1 failed

Press any key to continue . . .
```

While without calling `foo()` it's
```cmd

Test.exe is a Catch v1.1 b3 host application.
Run with -? for options

test
\giant_nas\dviry\my documents\visual studio 2015\projects\test\test\source.cpp(1
0)
...............................................................................

\giant_nas\dviry\my documents\visual studio 2015\projects\test\test\source.cpp(1
5): FAILED:
  REQUIRE( false )
with message:
  i := 2

test cases: 1 | 1 failed
assertions: 1 | 1 failed

Press any key to continue . . .
```

dvirtz:
Closes #748

### DIFF
--- a/include/internal/catch_message.hpp
+++ b/include/internal/catch_message.hpp
@@ -38,7 +38,10 @@ namespace Catch {
     {}
 
     ScopedMessage::~ScopedMessage() {
-        getResultCapture().popScopedMessage( m_info );
+        if (!std::uncaught_exception())
+        {
+            getResultCapture().popScopedMessage(m_info);
+        }
     }
 
 
--- a/include/reporters/catch_reporter_console.hpp
+++ b/include/reporters/catch_reporter_console.hpp
@@ -138,7 +138,11 @@ namespace Catch {
                     case ResultWas::ThrewException:
                         colour = Colour::Error;
                         passOrFail = "FAILED";
-                        messageLabel = "due to unexpected exception with message";
+                        messageLabel = "due to unexpected exception with ";
+                        if (_stats.infoMessages.size() == 1)
+                            messageLabel += "message";
+                        if (_stats.infoMessages.size() > 1)
+                            messageLabel += "messages";
                         break;
                     case ResultWas::FatalErrorCondition:
                         colour = Colour::Error;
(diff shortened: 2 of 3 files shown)


horenmar:
Looks good to me, I want to make minor changes before merging it.

I also have some minor concerns about using `std::uncaught_exception` causing unexpected behaviour (there is a reason it is deprecated in C++17), but it is unlikely to actually trigger, and even if it did, the difference in behaviour would be bad output message, nothing functional.

dvirtz:
Thanks. I believe it is deprecated only because 17 has a version which
tells the number of exceptions, not because there's something wrong with it.

On 3 Apr 2017 3:35 p.m., "Martin Hořeňovský" <notifications@github.com>
wrote:

Looks good to me, I want to make minor changes before merging it.

I also have some minor concerns about using std::uncaught_exception causing
unexpected behaviour (there is a reason it is deprecated in C++17), but it
is unlikely to actually trigger, and even if it did, the difference in
behaviour would be bad output message, nothing functional.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/philsquared/Catch/pull/876#issuecomment-291129733>, or mute
the thread

horenmar:
@dvirtz There are actually pretty good reasons why the new version exists, see [n4152](http://wg21.link/n4152.

philsquared:
actually, in addition to the limitations of `std::uncaught_exception` (that the plural version addresses) there are also problems with the implementation in several compiler versions. I forget which ones, now, but at least a few versions of GCC that were common a year or two ago (would need to look back through some, possibly close, issues to see which).
IIRC it was something like, once the flag had been set once it was not being reset - and so was unreliable!

horenmar:
@philsquared Oh yeah, we actually have that in [known limitations](https://github.com/philsquared/Catch/blob/master/docs/limitations.md#clangg----skipping-leaf-sections-after-an-exception).

Issue #757, opened 2016-12-01
compilation error : a type of global namespace


faithandbrave:
Hi, I am using `__int128` type of GCC/Clang. The type defined in global namespace. When I define `operator<<` in global namespace, happen compilation error.

```cpp
#include <iostream>

//namespace std { // workaround for Catch bug
// http://stackoverflow.com/questions/25114597/how-to-print-int128-in-g
std::ostream&
operator<<( std::ostream& dest, __int128 value )
{
    std::ostream::sentry s( dest );
    if ( s ) {
        __int128 tmp = value < 0 ? -value : value;
        char buffer[ 128 ];
        char* d = std::end( buffer );
        do
        {
            -- d;
            *d = "0123456789"[ tmp % 10 ];
            tmp /= 10;
        } while ( tmp != 0 );
        if ( value < 0 ) {
            -- d;
            *d = '-';
        }
        int len = std::end( buffer ) - d;
        if ( dest.rdbuf()->sputn( d, len ) != len ) {
            dest.setstate( std::ios_base::badbit );
        }
    }
    return dest;
}
//}

#define CATCH_CONFIG_MAIN
#include "catch.hpp"

TEST_CASE("compare of 128-bits integer") {
    __int128 a = 1, b = 1;
    CHECK(a == b);
}
```

```
$ clang++ main.cpp -std=c++14 -I /repository/GitHub/Catch/include
In file included from main.cpp:33:
In file included from /repository/GitHub/Catch/include/catch.hpp:34:
In file included from /repository/GitHub/Catch/include/internal/../reporters/../internal/catch_capture.hpp:11:
In file included from /repository/GitHub/Catch/include/internal/catch_result_builder.h:96:
In file included from /repository/GitHub/Catch/include/internal/catch_expression_lhs.hpp:13:
/repository/GitHub/Catch/include/internal/../reporters/../internal/catch_tostring.h:91:49: error:
      use of overloaded operator '<<' is ambiguous (with operand types 'std::ostream'
      (aka 'basic_ostream<char>') and 'const __int128')
        enum { value = sizeof( testStreamable(s << t) ) == sizeof( TrueType ) };
                                              ~ ^  ~
/repository/GitHub/Catch/include/internal/../reporters/../internal/catch_tostring.h:149:37: note:
      in instantiation of template class 'Catch::Detail::IsStreamInsertable<__int128>' requested here
    Detail::StringMakerBase<Detail::IsStreamInsertable<T>::value> {};
```

[...]

PR #877, opened 2017-04-02
Add opt-in c++11 stream insertable check.

### LINKED ISSUE #872: Compilation error when private base of class of tested variable is "ostreamable" but class itself isn't
## Description
When private base of class of tested variable is "ostreamable" but class itself isn't test code fails to compile.

### Steps to reproduce
```
#define CATCH_CONFIG_MAIN
#include <catch.hpp>
struct A {};
std::ostream &operator<< (std::ostream &o, const A &v) { return o << 0;}

struct B : private A
{
    bool operator==(int){ return true;}
};

B f ();
std::ostream g ();

TEST_CASE ("asdas")
{
    B x;
    REQUIRE (x == 4);
}

```

>main.cpp:1579:49: error: 'A' is an inaccessible base of 'B'
>         enum { value = sizeof( testStreamable(s << t) ) == sizeof( TrueType ) };

So the problem is in `IsStreamInsertable` check which features converting type to some bogus stremable type named `Borgtype` which everything converts to but it's pretty obvious that converting to your base class is still easier and thus error happens. It seems like replacing this check to the one featured in [this stackoverflow answer](http://stackoverflow.com/a/22759544/1269661) fixes this problem. I can send pull-request if that's needed and proposed solution is fine.

### Extra information
* Catch version: 1.8.2
* Operating System: Windows + online compilers
* Compiler+version: g++ 6.3.0, clang 3.8, MSVC 2017

Predelnik:
It adds opt-in define `CATCH_CONFIG_CPP_11_STREAM_INSERTABLE_CHECK` for enabling C++11 stream insertable check which should be better in some cases as suggested by @philsquared in #872.
All current tests seem to pass fine with it on my system.

Should I also add it to documentation somewhere?

PR should resolve #872 and #757 for those who opt-in.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -63,13 +63,35 @@ This can be useful on certain platforms that do not provide ```std::cout``` and 
     CATCH_CONFIG_CPP11_UNIQUE_PTR              // Use std::unique_ptr instead of std::auto_ptr
     CATCH_CONFIG_CPP11_SHUFFLE                 // Use std::shuffle instead of std::random_shuffle
     CATCH_CONFIG_CPP11_TYPE_TRAITS             // Use std::enable_if and <type_traits>
-    CATCH_CONFIG_CPP11_STREAM_INSERTABLE_CHECK // Perform C++11 expression SFINAE based check if classes could be inserted to std::ostream
+    CATCH_CONFIG_CPP11_STREAM_INSERTABLE_CHECK // Use C++11 expression SFINAE to check if class can be inserted to std::ostream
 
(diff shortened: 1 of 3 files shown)


horenmar:
You should explain the toggle in `docs/configuration.md`, but generally looks it looks good.

I'll fix-up whitespace and fmt later.

Predelnik:
I added explanation and removed weird underscore between `CPP` and `11`

horenmar:
It's in and I have to say, I honestly can't wait till Catch Classic is put into legacy mode, there is way too many C++11 toggles to test properly.

Issue #719, opened 2016-10-03
CTest integration


amigo421:
This is a feature request.
I see , in docs, piece of cmake script for the getting this from github.
however would be nice to have the script (cmake) for automatically adding CATCH scripts to CTest.
to make a single fully workable script which clones the Catch, scan the project for the tests and adds CTest target. 
There is a few scripts , somewhere in github which do this, I'm using this ability from time to time , 
would be perfect to have such module in Catch distribution, your opinion?

horenmar:
Not against it, but I admit I don't want to do this myself.

amigo421:
okay, I'll review which I'm using one, 
if this will look okay for a publishing , I'll create PR

PR #879, opened 2017-04-03
cmake script to generate ctest definitions from catch tests

amigo421:
#719 

this is a cmake script which can be used for automatic CTest generating by parsing c++ sources for CATCH tests. 
When you'd like to add CTest test you are usually doing this:

```
enable_testing()

add_test(...)
add_test(...)

```
Instead of the adding each test manually, I suggest to try using of this script which parses and adds found CATCH tests into cmake project

```
enable_testing()

ParseAndAddCatchTests(${PROJECT_NAME})

```

Notice 1: this script is a simplified version of  [Fraser999's scripts](https://github.com/Fraser999/MaidSafe/blob/846ce37c23bd86ee261e439919dfd7000a8f9372/cmake_modules/add_catch_tests.cmake)

Notice 2: this script doesn't replace and doesn't contain cmake commands for the adding CATCH lib into the project


horenmar:
I tested this out and I am probably going to merge this in, but I am not sure why it requires tags, even though it does not use them.

amigo421:
Give me a little bit time more, to review a tags functionality, probably it was required in initial version with a specific of the project, if yes, I'll remove this

philsquared:
Would it not be easier to run the Catch executable with `--list-just-test-names` and read the output (which will be each test name on its own line)?

amigo421:
> Would it not be easier to run the Catch executable with --list-just-test-names and read the output (which will be each test name on its own line)?

if I understand you and problem right, the configuration of cmake/ctest testing should be at cmake configuration step , even before project build, so no executable on this step

amigo421:
> but I am not sure why it requires tags, even though it does not use them.

if I understand your question right, the tags (in CATCH terminology)  = LABELS in terms of CTest, you can look at the output, the tags are added as LABELS property and can be used for the running a group of the tests, 
but really there is unused variable: CamelCaseProjectName, 
I'm going to remove this and re-submit

horenmar:
@amigo421 My problem with requiring tags (as opposed to optionally using them) is that for example in my current project we have almost no tags and instead we use Catch's `-#` option to run tests in individual test files.

If the script used them when they are present, but did not require them, I would be much happier with merging it.

amigo421:
you arer right, script currently requires a tags, I'll let you know here, when I'll have fixed this and make tags optionally

amigo421:
do you need something else from my side? current version should work with optional tags

horenmar:
No, I just needed to find some free time to test the changes. It is now in, in `contrib/`, because `scripts/` is meant for development scripts.

amigo421:
okay, got it. let me know if case of any steps required from my side

benjaminjack:
This script does not recognize tests when the opening curly bracket is on a new line after a the Catch macro. For example, the script will fail to recognize this tests:

```TEST_CASE("Feature construction", "[Feature]")
{
// tests here
}```

amigo421:
got this, going to fix
thank you

amigo421:
sorry for so long delay in one symbol fix :)
please review

horenmar:
@amigo421 You should open up a new PR for the changes. Otherwise it looks reasonable.

Issue #415, opened 2015-05-13
Use of INFO


mintyc:
IMO INFO currently is limited because it is scoped. It can only be used in the TEST_CASE body itself, not in mocks etc.

I would like to be able to log supplementary information that via the command line I can force to be output irrespective of whether an assertion fails.

Say I have some mock classes that I use in tests, I'd like them to be able to output INFO but they are scoped such that the INFO will never get out.
e.g. I have mocks of element types, Copyable, Movable etc and as well as asserting on numbers of constructions, move assignments etc in the test case itself, I sometimes want individual calls into the various methods to log a copy construction etc.

The existing behaviour of WARN can do this, but tags messages with 'warning' and is always on.

I'd propose existing INFO is renamed SCOPED to convey its restricted (but useful) purpose and INFO is repurposed to become more like WARN with the exception that it is only logged when explicitly requested on the command line.

philsquared:
Hi @mintyc. Sorry it's been a long time getting this (I'm just seeing it for the first time during an old tickets review).
Is this still something that's relevant to you?

Ironically what you proposed used to be the case: `INFO` worked unscoped, and there was a separate `SCOPED_INFO` that had the behaviour that `INFO` now does. But I kept getting reports from people who found the behaviour of `INFO` surprising so eventually changed it to the scoped behaviour.
So perhaps the solution would be a separate macro for the unscoped case? `UNSCOPED_INFO` doesn't sound like a great name, but you get the idea?

Anyway I'm going to put this back on the queue as a feature request.

mintyc:
Sounds  good 👍

horenmar:
`UNSCOPED_INFO` is a thing now.

PR #884, opened 2017-04-07
Add a note about when logging macros don't work

davidshepherd7:
## Description

The logging macros don't work quite the way I would expect when used inside if statements or loops. I think it's useful to have this documented.

## GitHub Issues

Related to  #415

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -2,6 +2,8 @@
 
 Additional messages can be logged during a test case.
 
+Note that these macros do not log anything unless they are *directly* inside a catch test scope macro (i.e. TEST_CASE, SECTION, GIVEN/WHEN/THEN/...). In particular they cannot be used inside loops or ```if``` statements.
+
 ## Streaming macros
 
 All these macros allow heterogenous sequences of values to be streaming using the insertion operator (```<<```) in the same way that std::ostream, std::cout, etc support it.


horenmar:
While I am not against making the documentation more explicit about how the logging macros are scoped, the proposed change is not actually true. Consider this naive test
```cpp
TEST_CASE("test1") {
    for (int i = 0; i < 10; ++i) {
        INFO("The number is " << i);
        REQUIRE(i == 0);
    }
}
```
the `INFO` macro is used in a loop, but the message is printed just as expected, because the macro is still in scope when an assertion fails.

davidshepherd7:
Ah right I think I misunderstood then. Is the requirement that they are in the same scope as the assertion?

horenmar:
Still not quite. They are scoped, which means that the assertion has to be in same or deeper scope, see these:
```cpp
TEST_CASE("Foo") {
    INFO("Test case start");
    for (int i = 0; i < 2; ++i) {
        INFO("The number is " << i);
        CHECK(i == 0);
    }
}

TEST_CASE("Bar") {
    INFO("Test case start");
    for (int i = 0; i < 2; ++i) {
        INFO("The number is " << i);
        CHECK(i == i);
    }
    CHECK(false);
}
```
In the first case, when the `CHECK` fails, two messages will be printed:
```
  Test case start
  The number is 1
```

In the second case, when the `CHECK(false)` fails, `INFO("The number is " << i);` is no longer in scope and won't be printed, so the output will be 
```
  Test case start
```

horenmar:
I am going to close this PR, but add the above example to the documentation to explain it better.

davidshepherd7:
Thanks, I was going to write a correct version but kept putting it off. :)

PR #892, opened 2017-04-22
cmake script to generate ctest definitions from catch tests

amigo421:
#719

this is a cmake script which can be used for automatic CTest generating by parsing c++ sources for CATCH tests.
When you'd like to add CTest test you are usually doing this:

enable_testing()

add_test(...)
add_test(...)

Instead of the adding each test manually, I suggest to try using of this script which parses and adds found CATCH tests into cmake project

enable_testing()

ParseAndAddCatchTests(${PROJECT_NAME})

Notice 1: this script is a simplified version of Fraser999's scripts

Notice 2: this script doesn't replace and doesn't contain cmake commands for the adding CATCH lib into the project

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -56,7 +56,7 @@ function(ParseFile SourceFile TestTarget)
     RemoveComments(Contents)
 
     # Find definition of test names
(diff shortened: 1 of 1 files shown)


horenmar:
I am going to test and merge it tomorrow.

In the meantime, if you don't mind, I am going to give you a bit of advice for better PRs. Seeing how this is a follow-up on the previous PR, you should note the reason for this one, but there is no need to copy verbatim the body of the previous one. IE, this one should be something along the lines of:


> This is a followup for #879, where I added a CMake script that can parse out Catch `TEST_CASE`s and register them in ctest. However, it does not recognize `TEST_CASE` macros where the opening brace is on a separate line.
>
> This PR fixes this.

and that is it.

amigo421:
> In the meantime, if you don't mind, I am going to give you a bit of advice for better PRs.

thank you veyr much! I have a huge experience in enterprise development (with a tens or even hundreds of the branches but no PR technique)  but absolutely don't know the rules  of open source game. I'm glad an any advice here

PR #900, opened 2017-05-01
Added warning if source file cannot be found

benjaminjack:
If source files are defined using relative paths, CMake will compile the tests, but this script will (sometimes) fail to find and parse the tests from the source files.  I have added an explicit warning when ParseAndAddCatchTests fails to find a source file.

## Description
ParseAndAddCatchTests will fail without warning when it cannot find source files to parse. There are instances where CMake will find sources to compile the executable, but ParseAndAddCatchTests fails to find those same source files when using relative paths. This can lead to confusion for the end user when CMake and make compile code without errors, yet running `make test` or `ctest` reports that there are no tests to run. An explicit warning allows users to resolve potential path problems.

## GitHub Issues
I decided to fix this problem myself instead of officially opening an issue.

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -48,6 +48,7 @@ endfunction()
 # Worker function
 function(ParseFile SourceFile TestTarget)
     if(NOT EXISTS ${SourceFile})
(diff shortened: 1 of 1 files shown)


horenmar:
This is likely to be helpful to someone, thanks.

PR #903, opened 2017-05-06
Added Win32 UNICODE wmain support

becrux:
## Description
Added alternate main definition, available on Win32 platforms, when UNICODE
support is enabled. argv command line is a wchar_t, so Catch run interface should
be adapted in order to manage it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ elseif(USE_CPP14)
     set(CMAKE_CXX_FLAGS "-std=c++14 ${CMAKE_CXX_FLAGS}")
 endif()
 
-if($ENV{USE_WMAIN} STREQUAL "1")
+if(USE_WMAIN)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /ENTRY:wmainCRTStartup")
 endif()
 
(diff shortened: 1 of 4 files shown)


becrux:
In case it is accepted, I guess a specific build on AppVeyor should be added, with "/SUBSYSTEM:CONSOLE /ENTRY:wmainCRTStartup" as parameters. I could do that, if I'm allowed.

horenmar:
I not against the change, but I have to ask what is the intended benefits of these changes? Catch will not be able to work with utf-8 encoded arguments anyway (well, not at this point) and tests should be able to have `UNICODE` defined regardless of the entry point.

Is there anything else that also changes when `wmain` is used as entry point?

becrux:
The problem occurs if someone uses wmainCRTStartup in is project, or MinGW-w64 GCC with -municode flag.

The are no benefits in terms of performance, or for the UNICODE support in the strict meaning (as you said), but it will save a developer from writing the same code that I wrote in order to adapt the wmain interface to the Catch::Session run method.

Short answer: it's just a convenience to have it working even when the above switches are ON, and you don't have to care about adapting the code.

becrux:
I've added proper configurations in appveyor.yml.

becrux:
Thanks for fixing appveyor.yml!

PR #908, opened 2017-05-14
Typo: Added namespace in method testCaseEnded

pJunger:
Hey,
I added the Catch namespace in the example, as it did not compile as is.


horenmar:
Thanks. I think the original idea was to have listeners in Catch's namespace, and then not all parts were changed.

PR #911, opened 2017-05-18
CTest integration script enhancements

dvirtz:
## Description
Made some improvements to ParseAndAddCatchTests.cmake script:

- use absolute path to test files - accroding to CMake docs EXISTS behavior is well-defined only for full paths
- added option to print debug messages
- added target name to test name and labels - to differentiate tests with similar names from different targets and to be able to run all tests coming from one target.


horenmar:
I wonder why using relative paths is not well defined for EXISTS (I mean its obvious why they could bug out horribly, not why it is not well defined).

Anyway, the changes look good.

PR #914, opened 2017-05-23
ExpressionLhs reconstruction based on value, not truthiness

twhittock:
Types which are truthy, but have more information than the truthiness in their string conversion were showing up as 'true' or 'false' instead of showing the underlying type's string value.


horenmar:
Thanks.

Issue #922, opened 2017-05-30
JUnit reporter improvements (for use with Jenkins)


tsondergaard:
Consider the following test cases:

```c++
TEST_CASE("no-explit-sections")
{
    REQUIRE(true);
}

TEST_CASE("one-explicit-section")
{
    REQUIRE(true);
    SECTION("section")
    {
        REQUIRE(true);
    }
}
```

Using the "junit" reporter produces the following output:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="evbase-test2" errors="0" failures="0" tests="3" hostname="tbd" time="0.111674" timestamp="2017-05-30T13:25:03Z">
    <testcase classname="global" name="no-explit-sections" time="0.000005"/>
    <testcase classname="one-explicit-section" name="root" time="0.000008"/>
    <testcase classname="one-explicit-section" name="section" time="0.000001"/>
  </testsuite>
</testsuites>
```

This is inconsistent and will not result in nice display results in Jenkins. The classname element should be constructed in a consistent manner. Based on how jenkins uses the classname attribute, the best fit is to construct the classname attribute from the suite-name and the [#filename] tag. The reason is that Jenkins doesn't care about the <testsuite> hiearchy in the junit xml reports. Instead it displays the results based on the content of the classname attributed loosely parsed as a fully qualified java class name, ie using "." to separate packages and class names.

See also my comment here: https://github.com/philsquared/Catch/issues/320#issuecomment-304874747

PR #923, opened 2017-05-30
Improve junit xml for jenkins

tsondergaard:
This pull request improves the junit xml output in the following ways

* The output isn't significantly different for TEST_CASEs with and without SECTIONs
* With "-#" test results will be grouped by [#filename] in jenkins test results UI
* The test suite name will be used for grouping too

Addressing these three issues ensures that the test results are logically organized in the jenkins test results UI.

This adresses #922 and maybe #320.

I've split the PR up in three tiny commits to ease review.

### DIFF
--- a/include/reporters/catch_reporter_junit.hpp
+++ b/include/reporters/catch_reporter_junit.hpp
@@ -139,8 +139,7 @@ namespace Catch {
             std::string className = stats.testInfo.className;
 
             if( className.empty() ) {
-                if( rootSection.childSections.empty() )
-                    className = "global";
+                className = "global";
             }
             writeSection( className, "", rootSection );
         }
--- a/include/reporters/catch_reporter_junit.hpp
+++ b/include/reporters/catch_reporter_junit.hpp
@@ -150,6 +150,10 @@ namespace Catch {
                 if ( className.empty() )
                     className = "global";
             }
+
+            if ( !m_config->name().empty() )
+                className = m_config->name() + "." + className;
+
             writeSection( className, "", rootSection );
         }
 
--- a/include/reporters/catch_reporter_junit.hpp
+++ b/include/reporters/catch_reporter_junit.hpp
@@ -128,6 +128,13 @@ namespace Catch {
             xml.scopedElement( "system-err" ).writeText( trim( stdErrForSuite.str() ), false );
         }
 
+        static std::string fileNameTag( const std::set<std::string> &tags ) {
+            std::set<std::string>::const_iterator it = tags.lower_bound("#");
+            if( it != tags.end() && !it->empty() && it->front() == '#' )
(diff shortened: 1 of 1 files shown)


horenmar:
I will have to think about this.

In general I agree with the changes, but it is not a change that can be done trivially. Both xml and junit reporters are meant to be consumable by outside tools, so changing their output is potentially breaking change.

include/reporters/catch_reporter_junit.hpp
```diff
@@ -128,6 +128,18 @@ namespace Catch {
             xml.scopedElement( "system-err" ).writeText( trim( stdErrForSuite.str() ), false );
         }
 
+        static std::string fileNameTag( const std::set<std::string> &tags ) {
+            for( std::set<std::string>::const_iterator
+                 it = tags.begin(),
+                 itEnd = tags.end();
+                 it != itEnd;
+                 ++it ) {
+                if( !it->empty() && (*it)[0] == '#' )
```
horenmar: The performance difference is unlikely to matter, but I would prefer this to be done via `std::set::lower_bound`.

tsondergaard:
This could indeed break things. The best option is perhaps to make a "junit2" reporter with the new behaviour. That is what I have done in my own project. It could also be called "junit-jenkins". The change that encodes the test suite name in the classname attribute as a java package justify this name, I think.

I can alter the pull request to do this.

include/reporters/catch_reporter_junit.hpp
tsondergaard: Okay, done.

include/reporters/catch_reporter_junit.hpp
```diff
@@ -128,6 +128,13 @@ namespace Catch {
             xml.scopedElement( "system-err" ).writeText( trim( stdErrForSuite.str() ), false );
         }
 
+        static std::string fileNameTag( const std::set<std::string> &tags ) {
+            std::set<std::string>::const_iterator it = tags.lower_bound("#");
+            if( it != tags.end() && !it->empty() && it->front() == '#' )
```
horenmar: `std::string::front()` is C++11 and forward

horenmar:
I thought about it a bit more, and decided to do two things:

1) I am going to merge the changes to the default junit reporter in Catch 2 branch. Since it is a major release, breaking changes are expected and this should be an improvements for most people.

2) I am going to link this PR from reporter documentation, so other people using Jenkins can make the changes for themselves, until Catch 2 is released.

I definitely do not want to add another reporter that is included in Catch by default, and single, relatively small, change is probably not worth a stand-alone reporter if it is going to be merged in (hopefully close) future.

tsondergaard:
I see the changes have been merged to branch dev-modernize, so closing pull request.

Issue #930, opened 2017-06-18
Feature Request: Make more info available from command line


BMBurstein:
## Description
Option to get more information about the tests without running them.
For example, get file names and line numbers for each test and get sections list for a test.

This will be useful when trying to connect to an external runner (I am currently trying to connect to Visual Studio's Test Explorer, but this information would probably be relevant for other runners, too).

horenmar:
This is in both Catch 1 and 2, so I am going to close it.

PR #934, opened 2017-06-21
add --extra-info flag

BMBurstein:
this will add line info to test lists, and test descriptions to the long
form of the test list

Related to issue #930

### DIFF
--- a/include/internal/catch_commandline.hpp
+++ b/include/internal/catch_commandline.hpp
@@ -187,6 +187,10 @@ namespace Catch {
             .describe( "list all/matching test cases names only" )
             .bind( &ConfigData::listTestNamesOnly );
 
+        cli["--extra-info"]
+            .describe( "list more info" )
+            .bind( &ConfigData::extraInfo );
+
         cli["--list-reporters"]
             .describe( "list all reporters" )
             .bind( &ConfigData::listReporters );
--- a/include/internal/catch_commandline.hpp
+++ b/include/internal/catch_commandline.hpp
@@ -187,9 +187,9 @@ namespace Catch {
             .describe( "list all/matching test cases names only" )
             .bind( &ConfigData::listTestNamesOnly );
 
-        cli["--extra-info"]
-            .describe( "list more info" )
-            .bind( &ConfigData::extraInfo );
+        cli["--list-extra-info"]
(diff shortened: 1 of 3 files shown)


horenmar:
I tested this quickly and I am willing to merge it*, but I found what I would consider a bug in the implementation of `--extra-info`:

If I pass only `--extra-info` option to Catch binary, it behaves as if no arguments were passed (all tests are run). This is different from all other list X commands, and even though it isn't named `list-*`, I think user would expect it to behave like one (especially since its description says "list more info").


* I didn't have time to look through the code yet, so standard disclaimers apply.

horenmar:
I decided to make some changes, it is now `--list-extra-info` and behaves like other `--list-*` flags, meaning that it lists tests on its own and prevents execution of tests.

Also improved the CLI description a bit.

BMBurstein:
Sorry, I saw your comment but hadn't had a chance to fix it myself.
I am wondering if I am going about this the wrong way. My intention was to make more info available for external runners to use, but for that maybe instead of adding more info to the existing lists, would it make more sense to create a new list command `--list-parseable` or something similar, which will list tests in a documented, parser-friendly format instead of a screen-friendly one? The format would probably include all the info available on a test (name, description, line info, tags, etc.) on a single line, with the different fields separated by some delimiter.
Would that be a better idea than the current one? Or maybe leave this already and add also the parser-friendly list command?

horenmar:
It is quite likely the wrong solution, but we had a discussion with Phil and we are willing to merge this in at least temporarily. Routing list outputs through reporters, so they can format the output accordingly, and introducing verbosity levels would be preferred, but I have no idea when that would make it in.

BMBurstein:
> Routing list outputs through reporters, so they can format the output accordingly, and introducing verbosity levels would be preferred, but I have no idea when that would make it in.

I would like to try working on this, but am unsure where to do this. It would probably involve some slight breaking changes, at least to the command line. Should I base this on the `master` branch or on `dev-modernize`? I am assuming that the later is the work on a Catch v2? How close is that to becoming stable?

horenmar:
@BMBurstein A preview of Catch v2 should be released relatively soon, but it being preview means that breaking changes might still get in afterwards.

If you want to try your hand at making this change, I think you should base your work on `dev-modernize`.

Issue #781, opened 2017-01-14
[Meta] Catch's performance


horenmar:
As well know, Catch's performance, at both runtime and compile time, could be better.

Some changes are on their way to master, but more could be always done. To that end, benchmarks (self-contained test suites) that show pathological usage patterns, representative usage patterns or just unexpected slowness, are welcome.

Obviously, so are PR's against [`dev-performance` branch](/philsquared/Catch/tree/dev-performance).

horenmar:
There is a request for special macros that only check for equality, without having expression decomposition, etc in #668.

horenmar:
After merging in `dev-performance-round2`, I did some benchmarking, comparing current master to last release (1.6.1).

Using 3 benchmarks
1. A long loop (2^20 iterations) with extremely simple `REQUIRE` condition.
2. A long loop (2^20 iterations) with simple `CHECK` condition, that fails.
3. Full benchmark suite of the [Json for Modern C++ project](https://github.com/nlohmann/json) (version 2.1.0, contains ~11M assertions).

I ended with these numbers

| Benchmark | 1.6.1 | master |
| ------------ | ------ | ------- |
| Passing tests | 3.4s | 0.7s |
| Failing tests | 15.5s | 15.1s |
| JSON for modern C++ | 6m 25s | 4m 14s |

philsquared:
Not bad! And nice use of a real world project

GatoRat:
FYI, the complaint I was given by one build engineer last year was that the compile time was becoming too long. I suggested precompiled headers, but he wasn't sure how to do it right on Linux.

horenmar:
@GatoRat Yeah, there definitely has been some compile time regression (I added compile time information to the post above), as the framework gains more features there is more code that the compiler has to chug through. If you have some specific code that has gotten significantly worse, please share.

Otherwise, the compile times should get a bit better soon, Phil has some simplification in store.

horenmar:
The new compile time flag (`CATCH_CONFIG_FAST_COMPILE`) speeds up test compilation on linux by ~10%. Windows benchmark pending.

horenmar:
Another ~15% speedup for test compilation using `CATCH_CONFIG_FAST_COMPILE` was achieved in 7a8a0205b46a3770cc8f0d26a24a903a4c03e82a.

PR #935, opened 2017-06-21
Performance improvement in AssertionInfo.

coombez:
The What:

By using char const * instead of std::string we avoid significant
copying per assertion.  In a simple loop with 10000000 CHECKS on
my system, this reduces the run time from 9.8s to 5.8s.

This change is at the expense of no longer capturing the second arg,
which no currently existing test notices.

The Why:

For some large scale matrix testing catch is proving too slow.  This change gets it about half way to what would be reasonable for this testing.  I'm hoping the other half may be found in the ostream usage and conversion to string in build via endExpression, but that looks much harder to tackle than this.

I have no desire to stop capturing the second arg, but it didn't break any tests, and I'm not yet familiar enough to realize if there is another easy way to keep it.  I'm hoping this PR will inspire you to help me out with ideas and your deeper knowledge of catch.

Ideas on how to tackle the other low hanging fruit would be great too, since I still have another 4s or so to shave off.

Thanks,
Neal

### DIFF
--- a/include/internal/catch_assertionresult.h
+++ b/include/internal/catch_assertionresult.h
@@ -40,14 +40,14 @@ namespace Catch {
     struct AssertionInfo
     {
         AssertionInfo() {}
-        AssertionInfo(  std::string const& _macroName,
+        AssertionInfo(  char const * _macroName,
                         SourceLineInfo const& _lineInfo,
-                        std::string const& _capturedExpression,
+                        char const * _capturedExpression,
                         ResultDisposition::Flags _resultDisposition );
 
-        std::string macroName;
+        char const * macroName;
         SourceLineInfo lineInfo;
-        std::string capturedExpression;
+        char const * capturedExpression;
         ResultDisposition::Flags resultDisposition;
     };
 
--- a/include/internal/catch_assertionresult.hpp
+++ b/include/internal/catch_assertionresult.hpp
@@ -13,9 +13,9 @@
 namespace Catch {
 
 
-    AssertionInfo::AssertionInfo(   std::string const& _macroName,
+    AssertionInfo::AssertionInfo(   char const * _macroName,
                                     SourceLineInfo const& _lineInfo,
-                                    std::string const& _capturedExpression,
+                                    char const * _capturedExpression,
                                     ResultDisposition::Flags _resultDisposition )
     :   macroName( _macroName ),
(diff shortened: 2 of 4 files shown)


horenmar:
I like the direction this takes (I did some perf work on Catch in the past, see #781), but a quick test and look at the diff says that you broke reporting for some macros, ie 
`REQUIRE_THROWS_WITH( thisThrows(), "expected exception" )` is now shown as `REQUIRE_THROWS_WITH( thisThrows() )`.

Proper way of testing Catch is to build the `SelfTest` binary and then run `scripts/approvalTests.py` with the binary, ie `...$ scripts/approvalTests.py build/SelfTest`. This will compare output of the new binary against known (hopefully) good output and tell you what has changed.

However, this should be fixable with some extra work.

coombez:
Thanks.  I had run SelfTest, but hadn't figured out how to run approvalTest.py (probably just didn't try hard enough :), I can now get the output, and see the result of my secondArg being removed.  I'll see if I can figure out how to wedge it back in without regressing the performance.

horenmar:
@coombez Our documentation for contributors is pretty bad, so I don't blame you. Just wanted to note that running the test binary by itself is not enough to verify proper output.

I actually wanted to remove the second argument at some point in the past, but in the end I decided it was not worth the time at that point. You will want to go through the macros in `catch_capture.hpp` and change them accordingly.

coombez:
Apparently github doesn't like the way I work :)  I'm closing this PR, a new one with a corrected commit will appear momentarily.

horenmar:
@philsquared The second argument is kind of holdover from some assumption about there being either 1 or 2 arguments to a macro, that is then used to either insert ',' when stringifying the macro, or not.

Not all macros actually use it, some do the ',' insertion manually, some have more arguments, etc, so it is kinda pointless.

philsquared:
Yeah, I didn't look close enough to see if it was really needed. The point was we can work around it even if it is

coombez:
PR #937 passes all tests, keeps the second arg with no change (seems like this is a separate issue to the performance), and gets the performance from 9.8s -> 6s, almost as good an improvement.  Hopefully something you can merge without much thought :)

coombez:
Noticed I never answered your question about needing c++11.  I currently use c++14 exclusively so that should be fine for me.

PR #937, opened 2017-06-22
Performance improvement in AssertionInfo.

coombez:
By using char const * instead of std::string we avoid significant
copying per assertion.  In a simple loop with 10000000 CHECKS on
my system, this reduces the run time from 9.8s to 6s.

### DIFF
--- a/include/internal/catch_assertionresult.h
+++ b/include/internal/catch_assertionresult.h
@@ -40,15 +40,17 @@ namespace Catch {
     struct AssertionInfo
     {
         AssertionInfo() {}
-        AssertionInfo(  std::string const& _macroName,
+        AssertionInfo(  char const * _macroName,
                         SourceLineInfo const& _lineInfo,
-                        std::string const& _capturedExpression,
-                        ResultDisposition::Flags _resultDisposition );
+                        char const * _capturedExpression,
+                        ResultDisposition::Flags _resultDisposition,
+                        char const * _secondArg = "");
 
-        std::string macroName;
+        char const * macroName;
         SourceLineInfo lineInfo;
-        std::string capturedExpression;
+        char const * capturedExpression;
         ResultDisposition::Flags resultDisposition;
+        char const * secondArg;
     };
 
     struct AssertionResultData
--- a/include/internal/catch_assertionresult.hpp
+++ b/include/internal/catch_assertionresult.hpp
@@ -13,14 +13,16 @@
 namespace Catch {
 
 
-    AssertionInfo::AssertionInfo(   std::string const& _macroName,
+    AssertionInfo::AssertionInfo(   char const * _macroName,
                                     SourceLineInfo const& _lineInfo,
-                                    std::string const& _capturedExpression,
-                                    ResultDisposition::Flags _resultDisposition )
+                                    char const * _capturedExpression,
+                                    ResultDisposition::Flags _resultDisposition,
+                                    char const * _secondArg)
     :   macroName( _macroName ),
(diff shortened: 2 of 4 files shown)


horenmar:
Ill take a look at this on the train, but first one more question: what exactly was your benchmark and is it measured with any optimizations?

coombez:
I can't provide the code, but basically -O3 optimized on gcc.  Pseudo code for the test with an rdtsc time source.

Time start = now();
for(int i = 0; i < 10000000; ++i)
    CHECK(now() - start < 10_seconds);

The time in here is just used as work to do that can't be completely optimized away, not for benchmarking.  For the runtime of the test, my build system tells me how long it ran, /usr/bin/time would work just fine for this purpose.  Or you could just output your equivalent of now() - start after the loop.

horenmar:
Ok, we have something similarly simple as Benchmark binary...

I will merge this and open a new branch you should check out, it removes the second argument and thus quite a lot of the extra concatenation. It doesn't quite pass approval tests, but it would be fairly simple to fix that and the perf gain might be worthwhile.

horenmar:
Ok, performance gains from `dev-second-string-argument-removal` should be final for those changes.

coombez:
I tested out dev-second-string-argument-removal, it neither improved or regressed performance, so seems fine to me.

I've also discovered that the single most expensive operation remaining is the construction/destruction of the std::ostringstream.  removing this drops a test with 100M instead of 10M iterations from 60s->5.5s.  A further drop from 5.5s->0.2s is achievable by avoiding building the assertion results and 'handling' them in the case of success.  That said, I have not achieved either of these changes while maintaining your tests of course.  But it's still useful information for where to focus future changes.

philsquared:
Yeah, I started a proof-of-concept rewrite of they core that did no eager conversions (and didn't use stringstreams at all.
I'm going to be taking a serious look at incorporating some of the ideas once I get v2.0 released.

coombez:
PR #939 eliminates the construction/destruction of the std::ostringstream while passing the tests.  Hopefully another easy win.

coombez:
For Catch v2.0, if you'll accept the suggestion, it may be worth benchmarking the assertions compared to the cost of an if-statement.  If that seems too restrictive, I wouldn't throw away that benchmark, but perhaps add a couple, one that include incrementing a count, not sure what the third should be but I'm sure you could come up with one.

Fundamentally, the real value Catch will add will be in the exceptional case of failure (or the user wanting full results), and that's where all the cost should lie.  Successful tests that are never reported should be as close to free as possible.

Once my last PR is merged, Catch assertions will be 40x faster! :)  But still 50x slower than a simple if.  Most of this is in the ostringstream.  But more than I expected is in the evaluation of whether successful tests should be included, in the increment of passed tests, and in the deferred checking of success or failure.

In Catch, I will reach a point of diminishing returns in continuing optimization soon, but if you're writing something from scratch, there are often ways to implement the features that will give you the benefits you want without costing in performance.  It's much harder to perfectly match results when optimizing after the fact, at least for large structural optimizations.

philsquared:
For Catch 2 the plan is to defer (almost) all the cost to the reporters - by passing just references to the original, typed, objects that are still on the stack in the assertion. It's up the reporters whether they will obtain a stringified representation or not (depending on whether the assertion passed or not, whether the `-s` option has been passed, or whether the reporter always logs all results, regardless.

I did it this way in my proof-of-concept spike to Catch2 and it worked very well - so I'm going to try to steer the current implementation that way.

PR #939, opened 2017-06-23
Cache std::ostringstream between assertions.

coombez:
This is not thread safe, but I think that was already true of Catch.
The construction/destruction of the std::ostringstream is where the
vast majority of time is spent per assertion.  A simple test of
100000000 CHECK()s is reduced from around 60s to 7.4s


horenmar:
Yeah, Catch Classic (1.x) is thread unsafe, because C++98 does not provide cross-platform facilities for dealing with threads...

Anyway, thanks I'll take a look later

PR #940, opened 2017-06-26
Eliminate some work when results won't be reported.

coombez:
I need some help with this one.  I've eliminated the vast majority of work in the cases that the results won't be reported (in general success).  It passes all but one tiny portion of the tests, which I'm hoping you'll know what I've missed.  This will get us from 7.6s->2.7s if I can figure out the last little piece.

After this half the work remaining is the m_stream.oss.str("") in the constructor, and the other half may be the m_stream.oss.str() in build().

### DIFF
--- a/include/internal/catch_capture.hpp
+++ b/include/internal/catch_capture.hpp
@@ -83,12 +83,12 @@
 ///////////////////////////////////////////////////////////////////////////////
 #define INTERNAL_CATCH_IF( macroName, resultDisposition, expr ) \
     INTERNAL_CATCH_TEST( macroName, resultDisposition, expr ); \
-    if( Catch::getResultCapture().getLastResult()->succeeded() )
+    if( Catch::getResultCapture().lastAssertionPassed() )
 
 ///////////////////////////////////////////////////////////////////////////////
 #define INTERNAL_CATCH_ELSE( macroName, resultDisposition, expr ) \
     INTERNAL_CATCH_TEST( macroName, resultDisposition, expr ); \
-    if( !Catch::getResultCapture().getLastResult()->succeeded() )
+    if( !Catch::getResultCapture().lastAssertionPassed() )
 
 ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


horenmar:
I see you figured out the IF issue -- didn't get around replying yesterday after I found it.

Gonna take a look at the last difference.

coombez:
This now passes all the tests while capturing the performance improvement.  If you can't squash these, let me know I'll just make another PR with a single commit.

coombez:
Still passes tests with this last commit, doubles performance again.

horenmar:
You can always squash and force push yourself; it will work out.

Anyway, I am going to say that I like the idea of these changes, but not necessarily the implementation. I don't like how much data and "smarts" was pushed back into the interface, so I am going to have to think this over.

coombez:
I don't believe there are any alternatives to making some of this data visible in the hotpath without simply avoiding manipulating it at all, which of course dramatically changes the test results.

coombez:
Accidentally pressed that button thinking it was a cancel button :)  didn't even intend to comment, but oh well.

coombez:
This should address your concerns.  Hopefully, it's more tempting to merge now :)

coombez:
Are there any remaining issues with this that I should be trying to address?

horenmar:
Nothing I can see quickly. I want to split out the changes into two commits and refactor it a bit, but it will be couple of days until I have the time.

horenmar:
I slept on this a bit and decided to merge it as is.

coombez:
Thanks.  When are you planning to release 1.9.7?  I am happy to just upgrade to HEAD, but will delay upgrading if you happen to be planning a release sometime soon anyhow.

horenmar:
Go with HEAD for now. I am enjoying trying to somewhat work off mobile internet throttled back to dial-up speeds (and not even the nice, 56k era) and Phil is also busy for at least a week (and the priority is on Catch 2 anyway).

coombez:
Heh, first world problems :)  Sounds good, thanks!

PR #947, opened 2017-07-05
Dev better verbosity

BMBurstein:
This change will allow different reporters to support different verbosity levels, including ones that will be added later, gracefully. If the reporter does not recognize the verbosity level, it will say so and exit.


horenmar:
As to the question in your first commit, IIRC the text formatting library is [textflowcpp](https://github.com/philsquared/textflowcpp).

Also, I don't think that Catch should throw (and thus usually abort) when a reporter doesn't support requested verbosity level, but rather I'd expect it to take "closest" (for given values of closest).

BMBurstein:
Not having the reporter try to guess the closest supported level allows adding verbosity levels in the future more easily (I was thinking of a "parsable" verbosity level), or adding support for existing verbosity levels to existing reporters without breaking anything that expects a certain output format.

horenmar:
Fair, but the other side of coin is that any time a new verbosity level is added, either the existing reporters (at least the ones bundled by default) need to be modified to support it as well, or we could end up with weird patchwork of supported verbosities.

Anyway, I am not saying that it has to be that way, just wanted to note that there are arguments against as well.

horenmar:
This should now be properly merged (including fixing the merge after myself)

PR #964, opened 2017-07-23
Prefer if ( MSVC ) in CMakeLists.txt.

johannesmoene:
Prefer to use `if(MSVC)` in [`CMakeLists.txt`](https://github.com/philsquared/Catch/blob/46bf7605f41a97535f63abfb2406ec408573b441/CMakeLists.txt#L293).

See e.g. Stackoverflow [What's the CMake syntax to set and use variables?](https://stackoverflow.com/a/31044116/437272), section *Recursive substitutions*.


horenmar:
Why is it preferred? I took a quick look at the linked answer, but as far as I can see, it just says that it works.

martinmoene:
As long as one doesn't rewrite it as `if ( CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" )` and use CMake < 3.1. 

I do like the repetition though of `if ( CMAKE_CXX_COMPILER_ID MATCHES "pattern" )` as it is now.

If you think `if ( MSVC )` doesn't improve anything, I'm completely happy to have this PR go unused.

horenmar:
I think I will pass then

PR #971, opened 2017-07-26
Adding more flexibility into the cmake catch parsing script

ddanila:
## Description

Just adding more flexibility for the cmake catch parsing script. No functional changes, mostly control over existing ones. For example, currently it always adds all tests with [!hide] tag which is not what I want. Meanwhile, I still want those tests to be compilable for a reason. Same for naming the tests -- for me, it's overkill to add fixtures and project name to the test name. For others it may be fine. I feel like configuration here is nice to have.

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -36,6 +36,9 @@
 cmake_minimum_required(VERSION 2.8.8)
 
 option(PARSE_CATCH_TESTS_VERBOSE "Print Catch to CTest parser debug messages" OFF)
+option(PARSE_CATCH_TESTS_NO_HIDDEN_TESTS "Exclude tests with [!hide] tag" OFF)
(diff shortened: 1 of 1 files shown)


ddanila:
Failed build is not because of me apparently, anyway this file is not involved in any tests or build chain.
`/usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/cstdio:120:11: error: no member named 'gets' in the global namespace`

horenmar:
Travis recently changed something that broke Clang in C++14 mode.

horenmar:
I got around to looking through this and there are 2 problems that should be fixed before it is merged

1) Only `[!hide]` is supported when hiding tests, but as far as I know, using it is rather rare. Much more common is using the short form of `[.]` and combined variation `[.foo]`.
2) There is a summary of the function at the top, it should be updated with new options.

ddanila:
Made an update to address the review. Thanks for it, btw.

horenmar:
A quick test says it works correctly now, so I am going to merge it. Thanks for the PR.

PR #976, opened 2017-07-30
Dev ibmi

zeromem88:
## Description
We want to make IBM ILE C++ unit tests for our IBM AS400 Mainframe software and run them on the mainframe side.
There is absolutelty no solution for IBM i ILE C++ Compiler for now.
This is the first attempt to comfigure environment to compile and run unit tests on IBM i (AS400) natively which succeeded.

Tested single header version on IBM i v7r2 and tutorial examples run perfectly.

Still need some improvements to make POSIX signals compile and work correctly (i'll try to implement slightly different FatalConditionHandler to leverage IBM i specific exception handling)

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -89,6 +89,10 @@
 
 #endif
 
+#ifdef __OS400__
+#       define CATCH_INTERNAL_CONFIG_NO_POSIX_SIGNALS
+#       define CATCH_CONFIG_COLOUR_NONE
+#endif
 
 ////////////////////////////////////////////////////////////////////////////////
 // Cygwin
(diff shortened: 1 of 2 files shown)


horenmar:
Looks good enough to me.

Out of curiosity, does your compiler support C++11?

zeromem88:
Partially yes.
IBM states that their latest operating system IBM i 7.3 has full c++11 support, but it should be tested first. I have some doubt about that because, for example, there is nothing about variadic templates in ibm manuals. Multi threading under a big question too.

Our current IBM I v 7.2 has partial support for c+0x TR0. Upgrade to the latest release usually happens when IBM finally announce end of life date. Mainframe maintenance best practice - never upgrade until you really need it. 
I suppose this will happen in a couple of years. Until that time no c++11 and IBM I 7.3

PR #982, opened 2017-08-03
Update matchers docs to reflect namespace usage

### LINKED ISSUE #972: [Documentation suggestion] matcher namespace requirement
The [matchers documentation](https://github.com/philsquared/Catch/blob/master/docs/matchers.md) currently does not mention anywhere that calls to the built-in matchers (StartsWith, Equals, EndsWith, etc) need to be qualified.

## Expected
```cpp
// file test.cpp
#define CATCH_CONFIG_MAIN
#include "catch.hpp"
TEST_CASE("matcher namespace")
{
    CHECK_THAT("foo", Equals("foo"));
}
```
should compile and work, as per the documentation.

## Actual
Need to qualify `Equals` as `Catch::Equals` or `Catch::Matchers::Equals` in order for `test.cpp` to compile. This is OK, but having it explicitly mentioned in the documentation would be nice. I can submit a pull request if you'd like.

codyhan94:
## Description

Update matchers documentation as per #972. Added a line in the first example making it clear that the built-in matchers are not in the global namespace.

## GitHub Issues

Closes #972

### DIFF
--- a/docs/matchers.md
+++ b/docs/matchers.md
@@ -13,6 +13,7 @@ which consists of either a single matcher or one or more matchers combined using
 For example, to assert that a string ends with a certain substring:
  
  ```c++
+using Catch::Matchers::EndsWith; // or Catch::EndsWith
 std::string str = getStringFromSomewhere();
 REQUIRE_THAT( str, EndsWith( "as a service" ) ); 
  ```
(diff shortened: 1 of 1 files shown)


horenmar:
Looks good enough

Issue #969, opened 2017-07-25
[Documentation suggestion] link to top of content for more pleasant ux


martinmoene:
Just an idea that crossed my mind, for example:

---
## How to use it
This documentation comprises these three parts:

* [Why do we need yet another C++ Test Framework?](https://github.com/philsquared/Catch/blob/1e59ccee41f269440310979a0b6ab7e4dbb1e453/docs/why-catch.md#why-do-we-need-yet-another-c-test-framework)
* [Tutorial](https://github.com/philsquared/Catch/blob/1e59ccee41f269440310979a0b6ab7e4dbb1e453/docs/tutorial.md#getting-catch) - getting started
* [Reference section](docs/Readme.md) - all the details **[no usable anchor]**
---
Note 1: full URL used here, in reality `docs/....` is used).
Note 2: perhaps provide a generic anchor at top of each page, like `<a id="top"></a>`.

horenmar:
Took me a while to see the difference.

Anyway, this is something that I think GitHub should provide out of the box, but since that is not the case at this moment, feel free to submit a PR. However, I would prefer a script that does this automatically, instead of having it remain manual work.

horenmar:
Since the PR was merged, I am going to close this.

PR #998, opened 2017-08-24
[Documentation] link to top of content

johannesmoene:
As per issue #969.

Preview result at [johannesmoene/Catch](https://github.com/johannesmoene/Catch/tree/doc-top-anchor).

It appears to me that adding some titles could improve the reading experience. For example: *Tutorial* in the [Tutorial](https://github.com/johannesmoene/Catch/blob/doc-top-anchor/docs/tutorial.md#top) and *Reference* in the [Reference](https://github.com/johannesmoene/Catch/blob/doc-top-anchor/docs/Readme.md#top).

Regards,
Martin (Joahannes) Moene


horenmar:
That does look somewhat better

Issue #999, opened 2017-08-24
[Documentation] Add several page titles


martinmoene:
Several documentation pages are missing a proper page title, notably:

- Tutorial (now start with *Getting Catch*)
- Reference
- Fixtures
- Configuration
- Command line

horenmar:
Since the PR has been merged, I am going to close this.

PR #1000, opened 2017-08-24
Add page titles

martinmoene:
As per issue #999


martinmoene:
Replaced by PR #1001.

PR #1003, opened 2017-08-26
Minor - added header-only flag in conan

solvingj:
See header-only guidelines: 
http://conanio.readthedocs.io/en/latest/howtos/header_only.html?highlight=header%20only
Its borderline cosmetic, but it does have a purpose.


horenmar:
LGTM

PR #1006, opened 2017-08-28
Remove duplicate CLI option

sam-bristow:
The "use-colour" option was accidentally duplicated as part of commit
feaf355 (Implemented libidentify support).


horenmar:
Well, that is kinda embarassing but at least it is harmless. Thanks.

Issue #1013, opened 2017-08-31
Copy initialization of explicit ctor prevents building on clang


offa:
Compiling the following minimal Test Main on Travis CI fails with Clang 3.8 - 4.0 (libc++ enabled):

```cpp
#define CATCH_CONFIG_MAIN
#include <catch.hpp> // Line 3
```

The error message:

```
In file included from […]/test/TestMain.cpp:3:
[…]/test/catch/catch.hpp:11064:16: error: 
      chosen constructor is explicit in copy-initialization
        return { };
               ^~~
/usr/include/c++/v1/set:428:14: note: constructor declared here
    explicit set(const value_compare& __comp = value_compare())
             ^
1 error generated.
```

Curiously this works on Travis using GCC 5 to 7 and also locally (GCC 7 and Clang4).

### Steps to reproduce

1. Use Catch2 (Single Header)
1. Add a Travis CI Build with the Test Main listed above and Clang with libc++ enabled
1. Build fails (Clang only)

### Extra information

* Catch version: **v2.0.0-develop.3** and **catch2** branch
* Operating System: **Travis (trusty)**
* Compiler+version: **clang 3.8 - 4.0 with libc++ enabled**

Works on the same CI Builds with any GCC from 5 to 7. Also works locally on Linux using GCC 7 and Clang 4.

offa:
I'll open a PR for this.

philsquared:
Thanks for this.
My reading is that this is a C++11 specific thing. In C++11 `set` had an explicit constructor with a defaulted argument. From C++14 that changes to having a separate default constructor, which is not explicit (it doesn't really make sense for a default constructor to be explicit!).
So looks like that's why this one slipped though the net.

offa:
This is also what [cppreference says](http://en.cppreference.com/w/cpp/container/set/set) (ctor 1). Curiously I'm on C++14 …

PR #1014, opened 2017-08-31
Explicit Ctor fixed

offa:
This is a fixes the build failure caused by an explicit ctor not being called explicit. The failure occurres only under some circumstances as described in #1013.

I accidentally fixed two whitespaces too.

## GitHub Issues

**Ref.** #1013


offa:
Appveyor failure is not related to this PR and already exists in *catch2*, from which this is branched.

philsquared:
Accepted via be067bce37f669d022bc6843a2fbe00a4446462e

PR #1017, opened 2017-09-04
Update handling of __JETBRAINS_IDE__ macro

fastasturtle:
1. Use it to conditionally define `CATCH_INTERNAL_CONFIG_COUNTER`, not
   `CATCH_CONFIG_COUNTER`, as `__JETBRAINS_IDE__` is similar to
   compiler-provided macros, not to user-provided ones.

2. Since `__COUNTER__` will work starting with CLion 2017.3, use it
   when possible (and hopefully remove this check altogether
   at some point).

## Description

CLion will support `__COUNTER__` macro properly soon (https://youtrack.jetbrains.com/issue/CPP-820)


horenmar:
Looks reasonable enough, but I am going to blame you if CLion 2017.3 comes out and it doesn't support `__COUNTER__` 😉

PR #1044, opened 2017-10-11
added PARSE_CATCH_TESTS_ADD_TO_CONFIGURE_DEPENDS option

dvirtz:
## Description
Adds option to ParseAndAddCatchTests.cmake which causes CMake to rerun when test files change so that new tests will be discovered.

## GitHub Issues
Resolves https://github.com/philsquared/Catch/issues/1035

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -36,6 +36,8 @@
 #    -- adds fixture class name to the test name                                                   #
 #    PARSE_CATCH_TESTS_ADD_TARGET_IN_TEST_NAME (Default ON)                                        #
 #    -- adds cmake target name to the test name                                                    #
+#    PARSE_CATCH_TESTS_ADD_TO_CONFIGURE_DEPENDS (Default OFF)                                      #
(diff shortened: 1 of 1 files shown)


horenmar:
I looked at the diff and `set_property` documentation and I wonder if it wouldn't be cleaner to add the dependency to the tests, rather than current directory.

Admittedly, this could cause CMake's dependency tracking to slow down significantly given enough tests, so it is more of a nitpick.

dvirtz:
What do you mean 'to the tests'?
The property means that the cmake configuration step depends on the given
files.

horenmar:
I am going to preface this with saying that I know just enough CMake to be dangerous.

Anyway, I thought that we could do
```cmake
set_property(
  TEST ${TestName}
  APPEND
  PROPERTY CMAKE_CONFIGURE_DEPENDS ${SourceFile}
)
```
to express that it is the test that depend on the source file, rather than a directory, but I tested it and it doesn't work.

dvirtz:
That's because it is a DIRECTORY property:
https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#properties-on-directories

horenmar:
Thanks for the PR btw

PR #1050, opened 2017-10-13
`RandomNumberGenerator::result_type` should be unsigned

sgrottel:
## Description
`result_type` must be unsigned:
http://en.cppreference.com/w/cpp/concept/UniformRandomBitGenerator

Using a signed type causes an infinite loop working with MS Visual Studio 2017, targetting: v140, WindowsTargetPlatformVersion 10.0.15063.0, Debug, x64

Simply using uint32_t seems sufficient, since the value of `max()` is small enough.

### DIFF
--- a/include/internal/catch_test_case_registry_impl.hpp
+++ b/include/internal/catch_test_case_registry_impl.hpp
@@ -22,7 +22,7 @@
 namespace Catch {
 
     struct RandomNumberGenerator {
-        typedef std::ptrdiff_t result_type;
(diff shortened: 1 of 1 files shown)


horenmar:
That change won't work for Catch master either, as it still has to be backwards compatible with C++98 (which doesn't have `<cstdint>`).

sgrottel:
Ah. I see. I wasn't aware of that. So, it's `unsigned int` then, I guess.

horenmar:
The last version will work nicely, thanks.

PR #1052, opened 2017-10-13
Fix URL for cross-link

LegalizeAdulthood:


horenmar:
That link won't work either but thank you for reminding me I haven't fixed that yet. (The proper link is https://github.com/philsquared/Catch/blob/master/docs/limitations.md#visual-studio-2017----raw-string-literal-in-assert-fails-to-compile)

Issue #1037, opened 2017-10-06
[Documentation suggestion] Provide executable examples and link to them


martinmoene:
Readers of Catch' documentation may benefit from readily available, complete and executable examples that show a limited and coherent set of features and steps needed for compilation. The tutorial and reference documentation can then link to relevant examples that the reader can [try, perhaps even online](https://wandbox.org/permlink/QzBURjkOxcFH99BF). 

Compilation of these examples must be integrated with the continuous integration process. As such the examples become [executable documentation](https://accu.org/index.php/journals/1871) geared towards the user.

For example:
```
// Assertions

// Compile:
// - g++ -I$(CATCH_SINGLE_INCLUDE) -o assertions assertions.cpp && assertions --success
// - cl -EHsc -I%CATCH_SINGLE_INCLUDE% assertions.cpp && assertions --success

// Let Catch provide a main():
#define CATCH_CONFIG_MAIN

#include "catch.hpp"

int one() {
    return 1;
}

TEST_CASE( "Assert that something is true (pass)", "[require]" ) {
    REQUIRE( one() == 1 );
}

TEST_CASE( "Assert that something is true (fail)", "[require]" ) {
    REQUIRE( one() == 2 );
}

TEST_CASE( "Assert that something is true (stop at first failure)", "[require]" ) {
    REQUIRE( one() == 2 );
    REQUIRE( one() == 1 );
}

TEST_CASE( "Assert that something is true (continue after failure)", "[check]" ) {
    CHECK(   one() == 2 );
    REQUIRE( one() == 1 );
}
```

Subjects that come to mind:
- test case (test_case)
- assertions (require, check)
- exceptions (require_nothrow, require_throws, require_throws_as, require_throws_with, require_throws_matches)
- matchers (require_that)
- floating point (approx)
- logging (capture, info, warn, fail, fail_check)
- fixtures (test_case_method)
- sections
- bdd (scenario, given, when, then, and_when, and_then)
- string conversions
- reporters
- event listeners
- supplying your own main()
- compile-time configuration

horenmar:
Basically, you want parts of SelfTest to become self-contained and hosted on Wandbox or similar, correct?

martinmoene:
Indeed, with the difference that SelfTest may be geared (too much) towards implementers, complete and overwhelming where mere users may need another approach to support their understanding.

Key:
- relatively simple
- complete program
- that compiles and executes and shows what's necessary for that
- and is at the tip if reader's fingers from the documentation

[...]

PR #1055, opened 2017-10-14
[Please review, don't merge] Add examples

martinmoene:
Hi,

Per issue #1037 I've added folder examples, docs/list-of-example.md and a small subset of the example mentioned there.

I'd like to know if this PR represents a sane direction, or if there are any other remarks you'd like to make.

Thanks in advance.

### DIFF
--- a/docs/event-listeners.md
+++ b/docs/event-listeners.md
@@ -19,7 +19,7 @@ file that defines `CATCH_CONFIG_EXTERNAL_INTERFACES`.
 
 Then register it using `CATCH_REGISTER_LISTENER`.
 
-For example:
+For example ([complete source code](../examples/210-Evt-EventListeners.cpp)):
 
 ```c++
 #define CATCH_CONFIG_MAIN
--- a/docs/list-of-examples.md
+++ b/docs/list-of-examples.md
@@ -1,0 +1,31 @@
+<a id="top"></a>
+# List of examples
+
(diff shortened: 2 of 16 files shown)


horenmar:
Looks reasonable enough so far, but I also remember you wanting to make compiling and running these part of CI. Are you still planning to do so?

martinmoene:
Yeah, I'll have a look at compiling as part of CI shortly; I'm not so sure about running though.

philsquared:
Argh, sorry! I always forget that deleting a branch closes all associated issues and PRs - with no way to re-open :-(

philsquared:
How much of this is still outstanding? I know you've been working through it.

martinmoene:
I think it's perfectly fine to have this one closed.

I had started to look at integrating compilation of the examples with CI, but got stuck on how to approach that in face of what's already there (and got sidetracked by other things).

Further, (many) more examples must be created and locations found to link to them.

A question is: should all examples be added at one time, or can it be done incrementally (once building them is integrated)?

horenmar:
@martinmoene It is ok to add them incrementally. As to building them, I suggest creating a standalone CMakeLists that defines all targets for the examples and then (conditionally) including it from the main CMakeLists.

Also, note that they should not be built as part of normal development build, but should be hidden behind a CI specific option (we already do that for C++14, so you can just copy that).

martinmoene:
@horenmar Thanks for the pointers.

PR #1057, opened 2017-10-16
adds flushes to the output stream of teamcity reporter,…

sgrottel:
… making the test output more responsive.

## Description
There is the issue with the teamcity reporter described here:
https://github.com/philsquared/Catch/blob/master/docs/build-systems.md#teamcity-reporter
Tests are only reported after the whole suite is completed.
This can be fixed by explicitly flushing the output stream of the reporter.

I added flush commands to the output functions of the reporter.

Now (long) running tests are reported in Teamcity more quickly, almost as soon as the tests are completed.

If these flushes would create a performance issue, additional code could be added to only flush every some seconds. But in my setup it did not seem necessary.

### DIFF
--- a/include/reporters/catch_reporter_teamcity.hpp
+++ b/include/reporters/catch_reporter_teamcity.hpp
@@ -141,6 +141,7 @@ namespace Catch {
                            << "]\n";
                 }
             }
+            stream.flush();
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


horenmar:
This is a potential issue with streaming reporters in general and we have had to fix it for the XML reporter earlier in the year.

Sadly without NVI-based rework of the reporters, it has to be fixed in each reporter individually.

PR #1063, opened 2017-10-23
created destructor for LeakDetector

bigboynaruto:
## Description

This code generates memory leaks.
```#define CATCH_CONFIG_RUNNER
#include "catch.hpp"

int main(int argc, char **argv) {
    if (false) { // do not run tests
        return Catch::Session().run(argc, argv);
    }
}
```

### DIFF
--- a/include/catch.hpp
+++ b/include/catch.hpp
@@ -69,7 +69,12 @@ public:
     }
 };
 #else
-class LeakDetector {};
+class LeakDetector {
+    public:
+        ~LeakDetector() {
+            Catch::cleanUp();
+        }
+};
 #endif
 
 LeakDetector leakDetector;


philsquared:
It does.
Is there a reason this is a problem?
One of `Catch::Session`'s roles is to clean up singletons in it's destructor.
If you want to do it without `Catch::Session` for some reason you can explicitly call `Catch::cleanUp()` instead.

philsquared:
- sorry didn't notice before that your PR was to put `Catch::cleanUp()` in `LeakDetector`s destructor.
That won't work for two reasons:
1. cleanup need to have run before leak detection
2. cleanup should only be called once (we could put a flag in, but IIRC that's not how it works now) - so if you _do_ have a `Session` object too it will break.

bigboynaruto:
So is it ok If I create this flag?

philsquared:
Actually, looking at the code it might work ok. It deletes objects then sets them to `null`. I think it used to be more problematic.
Also the leak detection happens later than the `LeakDetector`'s destructor.
So maybe your PR is ok as-is

bigboynaruto:
Ok.

horenmar:
Is there a reason why you add it only in non-Windows `LeakDetector`?

bigboynaruto:
No, I just don't have Windows, so I can't test it.

JoeyGrajciar:
@bigboynaruto Just checked on Windows and adding it also for Windows LeakDetector would help :)

Issue #955, opened 2017-07-14
Change to capture exception by (const) &


martinmoene:
Catch' `INTERNAL_CATCH_THROWS_AS()` catches its exception **by value**. The [Cpp Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) (Scott's More Effective C++ Item 13) advices to capture [exceptions by `(const) &`](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-exception-ref).

See [`catch_capture.hpp`](https://github.com/philsquared/Catch/blob/fae0fa4ec158438b9f6c6f8b4d73bdf631c08ed6/include/internal/catch_capture.hpp#L133) line 133.

Noted [here at doctest](https://github.com/onqtam/doctest/issues/81).

philsquared:
This changed at some point. I think there was some complaint that by doing it by reference within the macro you didn't get the opportunity to choose different approaches yourself.
With it not being in the macro you can still add the reference to the type you pass as the argument to the macro.

I'm still not sure how I feel about this one.

I'm inclined to change it back, but that would now be a breaking change (especially if you've added the reference externally). So I may revisit for Catch2.

horenmar:
We catch by whatever type we are given, user can choose to have Catch capture by `const&` just by writing it that way.

We actually had it using TMP to ensure the type is properly `const&`-ed no matter what, but that broke people who used the macros inside a templated function, and couldn't be fixed in C++98 (IIRC it was because `typename` cannot be used in plain old function and without it, it didn't work in templated functions).

This rule was changed in C++11, so we can do it without breaking people who used it correctly, but that would mean using type traits in the common include path, which is... unfortunate.

martinmoene:
Tes it is: [Expects that an exception of the specified type is thrown during evaluation of the expression. Note that the exception type is used verbatim and you should include (const) reference.](https://github.com/philsquared/Catch/blob/master/docs/assertions.md#exceptions).

philsquared:
Yeah. Could be clearer, though.
Personally I would be happy to just unconditionally add the `const&` - and provide an alternate way (or let user do it the long way round) for cases where that was not wanted (which must be a small edge case).
As I say, though, that would be quite a breaking change - so will think about it for Catch2.

martinmoene:
Thanks for the feedback.

philsquared:
Thanks for bringing this up, @martinmoene

[...]

PR #1066, opened 2017-10-25
Fix catch exception by reference

tiagomacarios:
When compiling our codebase with clang we were getting the following
error:

GSL/tests/algorithm_tests.cpp:198:58: warning: catch handler catches by
value; should catch by reference instead
[misc-throw-by-value-catch-by-reference]
    CHECK_THROWS_AS(copy(src_span_dyn, dst_span_static), fail_fast);
                                                             ^
Looking at the catch source code exceptions were being caught by value.
One could have it designed so that users might say:

CHECK_THROWS_AS(copy(src_span_dyn, dst_span_static), fail_fast&);

But I don't think this is the intent and looking at the Catch tests
itself looks like this macro does not expect the reference:

    REQUIRE_THROWS_AS( thisThrows(), std::domain_error );
    CHECK_THROWS_AS( thisDoesntThrow(), std::domain_error );

This commit makes Catch catch exceptions by reference instead of by
value.

### DIFF
--- a/include/internal/catch_capture.hpp
+++ b/include/internal/catch_capture.hpp
@@ -136,7 +136,7 @@
                 static_cast<void>(expr); \
                 __catchResult.captureResult( Catch::ResultWas::DidntThrowException ); \
             } \
-            catch( exceptionType ) { \
+            catch( exceptionType& ) { \
                 __catchResult.captureResult( Catch::ResultWas::Ok ); \
             } \
             catch( ... ) { \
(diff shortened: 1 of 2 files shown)


horenmar:
This was already discussed in #955.

The result is that Catch Classic is going to keep the behaviour for backward compatibility reasons, with Catch 2 changing the default to capture by `const&`.

tiagomacarios:
Seems a bit suboptimal to have people to run the compiler suppressing an error for backward compatibility. Is Catch2 ready for production?

horenmar:
@tiagomacarios They can also append a `const&` to the exception type, or use a stand-in macro that converts the type, to have a single point of change later on.

Generally, we don't want to commit breaking changes in minor releases and next major is Catch 2, so this gets to go there either way.

horenmar:
Catch 2 is out, so I am going to close this as already fixed

Issue #980, opened 2017-08-02
[Overarching] Approx issues


horenmar:
There are multiple currently open issues against Approx, that need fixing. Because floating point is hard, it will take a while to fix them properly. In the meantime, this is an overarching issue to keep track of the progress.

### Related issues
* #952 - Margin is non-inclusive (and does not quite work properly)
* #953, #962 - Default approx scale leads to weird results for small numbers

TBBle:
For the record, [cppreference](http://en.cppreference.com/w/cpp/types/numeric_limits/epsilon) gives a reasonable example of floating-point comparison with a relative precision.
```cpp
template<class T>
typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type
    almost_equal(T x, T y, int ulp)
{
    // the machine epsilon has to be scaled to the magnitude of the values used
    // and multiplied by the desired precision in ULPs (units in the last place)
    return std::abs(x-y) < std::numeric_limits<T>::epsilon() * std::abs(x+y) * ulp
    // unless the result is subnormal
           || std::abs(x-y) < std::numeric_limits<T>::min();
}
```

The current API documentation is quite confusing, it's not clear if `episilon` is a percentage with an amazingly low default, or is a value from which a percentage is derived through _magic_, since the docs don't show any relationship between the value set, and the percentage.

`margin` itself is clear, except as noted in #952 the documentation implies it's an inclusive range for success, but the code itself is exclusive range for success.

I don't see why you'd want `scale` distinctly, but it might be due to confusion over `epsilon`'s behaviour, per #962. (The code confused me _further_...)

I also can't see how or why you'd want to combine _both_ relative and absolute tolerances, but I can't really give a reason that you wouldn't.

So it would make sense for me (and avoid surprising existing satisfied users) to introduce two new things like `Approx`:
* `input == Within(value, inclusiveAbsoluteRangeValue)` implying `fabs(input-value) <= inclusiveAbsoluteRangeValue `; and
* `input == Roughly(value, someRelativeMeasure)` implying `fabs(input-value) <= fabs(input-value) * std::numeric_limits<decltype(input-value)>::epsilon() * mutationOf(someRelativeMeasure)`.

That second `fabs(input-value)` might be the wrong multiplier for `std::numeric_limits<>::epsilon`, I see Catch currently takes the machine epsilon of the max of the two sides, rather than the machine epsilon of the difference.

[...]

PR #1068, opened 2017-10-26
more robust Approx behaving as usual(?) expected

Pfiffikus:
What about to check for a well-defined newEpsilon (cf. http://realtimecollisiondetection.net/blog/?p=89)?

Apart from computational sciences view (cf. https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/ or https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html) I would prefer two slight modifications referring to https://en.wikipedia.org/wiki/Approximation_error#Formal_Definition to get the common(?) behaviour:
- The given _epsilon should refer to the target value_, otherwise the result would be unexpected, e.g. 101.02 == Approx(100).epsilon(0.01) gets true. 
- The _default scale should be invisible_, thus, e.g. 101.02 == Approx(100).epsilon(0.01) gets false. Finally (both modifications accepted) even 101 == Approx(100).epsilon(0.01) would get false, e.g.

### DIFF
--- a/include/internal/catch_approx.hpp
+++ b/include/internal/catch_approx.hpp
@@ -25,7 +25,7 @@ namespace Detail {
         explicit Approx ( double value )
         :   m_epsilon( std::numeric_limits<float>::epsilon()*100 ),
             m_margin( 0.0 ),
-            m_scale( 1.0 ),
+            m_scale( 0.0 ),
             m_value( value )
         {}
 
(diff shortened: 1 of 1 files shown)


horenmar:
We have an outstanding issue about reworking Approx, #980, so Approx will definitely receive some changes. There is, however, a good chance that the changes will target Catch 2, instead of Catch 1, as they end up a breaking change.

It will be couple of days before I get enough time to look through this.

horenmar:
I ended up cherry picking and cleaning this up in 00af677577973758b3b35e1c94ab4142c45c3f67 for the Catch 2 branch, where breaking changes are OK.

GitHub doesn't pick up cherry picks properly, so I am going to close this manually.

PR #1080, opened 2017-11-04
Replace include_directories() with target_include_directories()

martinmoene:
to prevent inheritance of include directories that possibly lead to a clash.

A clash occurs when a folder is included, e.g. examples, that wants to use the single-include directory instead of the normal include directory as used by the SelfTest in the next higher level.

CMake: [include_directories()](https://cmake.org/cmake/help/v3.0/command/include_directories.html)
CMake: [target_include_directories()](https://cmake.org/cmake/help/v3.0/command/target_include_directories.html)

Aside: I am a little surprised that the CMake description for SelfTest is in the toplevel `CMakeFiles.txt` and not in e.g. `projects/SelfTest/CMakeFiles.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,12 +281,12 @@ SOURCE_GROUP("Tests" FILES ${TEST_SOURCES})
 SOURCE_GROUP("Surrogates" FILES ${SURROGATE_SOURCES})
 
 # configure the executable
-include_directories(${HEADER_DIR})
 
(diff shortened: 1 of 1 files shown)


horenmar:
Why add the same invocation separately for each compiler?

Otherwise LGTM. There are couple more dumb things we do with global build flags, but so far nobody needed them fixed.

martinmoene:
Indeed, that did surface in my mind after sending... I'll update the PR

martinmoene:
PR updated.

PR #1082, opened 2017-11-05
Add example code linked to from documentation, part 1

martinmoene:
PR linked to issue #1037.

I've integrated building of the examples into CMake.

**`USE_EXAMPLES`**&emsp;The examples are built if `USE_EXAMPLES` is `true`. This name follows the existing pattern, but may not properly reflect the choice. `NO_EXAMPLES` is the wrong way round. Plain `EXAMPLES` is weird.
Do you have suggestions?

**Travis, AppVeyor**&emsp; I've not touched CI yet. Do you have suggestions here; I wouldn't mind to leave this to experts.

Other remarks?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
           packages: ['valgrind', 'clang-3.6']
       env: COMPILER='clang++-3.6' VALGRIND=1
 
-# Travis's containers do not seem to have Clang 3.7 in apt, no matter what sources I add. 
+# Travis's containers do not seem to have Clang 3.7 in apt, no matter what sources I add.
 #    - os: linux
 #      compiler: clang
 #      addons:
@@ -110,6 +110,10 @@ matrix:
           packages: ['valgrind', 'g++-7']
       env: COMPILER='g++-7' VALGRIND=1
 
+    - os: linux
+      compiler: gcc
+      addons: *gcc7
+      env: COMPILER='g++-7' ENV_NO_SELFTEST=1 ENV_BUILD_EXAMPLES=1
 
     # 3b/ Linux C++14 Clang builds
     - os: linux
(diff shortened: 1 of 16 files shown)


martinmoene:
Should we perhaps use `NO_SELFTEST` on the build with examples, as the selftest has already been created for that confiuration?

horenmar:
> Do you have suggestions?

`BUILD_EXAMPLES`?

> Do you have suggestions here; I wouldn't mind to leave this to experts.

Not sure where we are going to get them 😄 , but for Travis, we already specify all elements of the build matrix explicitly, so we can just add new one:
```yaml
    - os: linux
      compiler: gcc
      addons: *gcc6
      env: COMPILER='g++-6' BUILD_TYPE='Debug' CPP14=1 EXAMPLES=1
```
and change the build line:
```yaml
  - cmake -H. -BBuild -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -Wdev -DUSE_CPP11=${CPP11} -DUSE_CPP14=${CPP14} -DBUILD_EXAMPLES=${EXAMPLES}
```
I don't think we need to build the examples with all compilers and configurations, as SelfTest should catch bugs, we just want the examples to not get out of date.


AppVeyor would be a bit harder, but I think we can specify explicit inclusion for the build matrix similarly:
```yaml
matrix:
    include:
        -
            Examples: 1
            os: Visual Studio 2015
        -
            Examples: 1
            os: Visual Studio 2013
```
and change the `before_build` lines similarly.

martinmoene:
To promote the desired use of Catch with separate compilation of Catch itself and the tests in one or more other source files, I'm considering to apply that idiom to all tests with a common main (e.g. 000-CatchMain.cpp) except for  examples 010-TestCase.cpp (single file usage) and 020-TestCase,-[12].cpp (multiple files usage).

Would that be a good idea?

martinmoene:
>Looking through or travis config, we should do some cleanup there. CPP11 toggle is no longer relevant, as that is the base assumption.

Perhaps not slip this in this subject's PR.

horenmar:
@martinmoene The cleanup is done, but you will likely need to rebase.

horenmar:
Keeping the number of separately compiled mains down is a good idea, they take quite a long time to compile. The fact that it teaches good practices is a nice bonus.

martinmoene:
@horenmar File `210-Evt-EventListeners.cpp` fails to compile without `CATCH_CONFIG_MAIN`:
```
../examples/210-Evt-EventListeners.cpp:70:81: error: 'Catch::TestRunInfo' has not been declared
```
Same holds for other listener-related types.

Did I miss something?

horenmar:
@MartinMoene Listeners and reporters require `CATCH_CONFIG_EXTERNAL_INTERFACES` to be defined. It is implied by both `RUNNER` and `MAIN`, which is why it works. It should be in the docs :smile:

martinmoene:
It _is_ in the docs (blush).

martinmoene:
@horenmar 

Up to now I've got:
- several examples that are linked to from the documentation
- examples/CMakeLists.txt mentioned in ./CMakeLists.txt
- integration with Travis
- initial integration with AppVeyor

I could use some help with the latter.

I tried to compile examples only for a single configuration, but have not yet succeeded. Your suggestion does not work: `include` is not a valid word to use there.

horenmar:
@martinmoene Yeah, I've noticed so. It is in Travis, so I just assumed it would work.

I'll take a look at AppVeyor (hopefully soon)

martinmoene:
@horenmar I've been able to limit building of examples to Win32 and x64, Release only. I've (silently) removed the exclusion phrase for VS2013.

What's next: offer a squashed PR for examples, part1?

horenmar:
Pretty much

martinmoene:
Squashed, ready to be merged.

horenmar:
Looks good to me. I might rejigger the CI matrix afterwards, I have some further changes planned :smile:

martinmoene:
Thanks!

PR #1091, opened 2017-11-09
Stop GCC warning suppression leaking out of catch

### LINKED ISSUE #1090: Catch disables GCC unused variable warnings in test files.
## Description
When compiling a test file with an unused variable no warning was generated under GCC.
After removing the catch #include the compiler warning came back.

Discovered this when debugging a unit test and found an unused variable and expected that my compiler should be throwing a warning.

### Steps to reproduce
Given:
```cpp
#include "catch.hpp"
static int test;
```
Compile with
```bash
g++ test.cpp -Wall
```
Expecting a warning but none given.

After removing the catch include we get:
```
test.cpp:3:12: warning: 'test' defined but not used [-Wunused-variable]
```

### Extra information
* Catch version: **v2.0.1**
* Operating System: **Windows 7**
* Compiler+version: **GCC v7.1.0 (MinGW-W64)**

Seems to a from this section in catch_suppress_warnings.h:
```c
#elif defined __GNUC__
#    pragma GCC diagnostic ignored "-Wvariadic-macros"
#    pragma GCC diagnostic ignored "-Wunused-variable"
#    pragma GCC diagnostic ignored "-Wparentheses"

#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wpadded"
#endif
```
Looks like the three warnings suppression are leaked to user code. Should they be moved to after the `push`?

Moving the push to be the first GCC pragma seems to fix it for me and no other warnings are reported.

mike-rob:
Fixes an issue where catch silences some GCC compiler warnings in the unit test.

## Description

`#pragma GCC diagnostic push` should come before any warnings are ignored.
Moving the push to the before any other GCC pragma means any warnings ignored in catch will be restored at end of the header by `#pragma GCC diagnostic pop`

Clang also does not have the push pragma first. Should that also be updated?
Is there a good reason that I have missed for allowing some warning pragma to leak?

## GitHub Issues

Closes #1090.

### DIFF
--- a/include/internal/catch_suppress_warnings.h
+++ b/include/internal/catch_suppress_warnings.h
@@ -17,8 +17,8 @@
 #       pragma clang diagnostic ignored "-Wcovered-switch-default"
 #    endif
 #elif defined __GNUC__
-#    pragma GCC diagnostic ignored "-Wunused-variable"
 #    pragma GCC diagnostic ignored "-Wparentheses"
 #    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wunused-variable"
 #    pragma GCC diagnostic ignored "-Wpadded"
 #endif


horenmar:
The last version was looking good.

I rebased it to keep the history simpler but you should still have the commit attribution.

PR #1102, opened 2017-11-15
Fix AssertionPrinter name conflict in console and compact reporters

coombez:


horenmar:
Doh, thanks.

We really need to set up CI tests to work with the single include version as well as with the development version.

Issue #1086, opened 2017-11-08
Catch2 Performance


coombez:
Took a quick look at Catch2 today.  I'll need to spend some time on its performance before I'll be able to adopt Catch2 for our code base.  It looks like it has some significant performance regressions over Catch1.

Here is the test case for Catch1 and Catch2:

```c++
for(uint64_t i = initial ; i < initial + 1000000000; ++i)
    CHECK(i != reinterpret_cast<uint64_t>(&f));
```

Here's a reasonable benchmark for how long an assertion should take:

```cpp
for(uint64_t i = initial ; i < initial + 1000000000; ++i)
    if(!(i != reinterpret_cast<uint64_t>(&f)))
    {
        ++failed;
        throw "Do Reporting";
    }
    else
    {
        ++passed;
        if(report_non_failure)
            throw "Do Reporting";
    }
```

f is some function, this just makes sure the check doesn't get completely optimized away.  initial is just a value used if you happen to hit the value of the function address.  report_non_failure is a bool that indicates whether even successful checks should be reported.  passed and failed are counters.  The throw's are just substitutes for all the beautiful reporting Catch does :)

Catch2 test: 2m51.072s
Catch1 test: 0m28.460s
reasonable benchmark: 0m00.453s

coombez:
Forgot to mention I ran the Catch2 test above through prof as well.  Despite being compiled with full optimization it looks like most of the functions were not being inlined.  Trivial stuff like Catch::isTrue, and the StringRef constructors and destructors consumed 10s of percent of runtime.  Experimentally I was able to fix a few by adding the inline keyword and moving them closer to the beginning of the single include file.

horenmar:
That is definitely a pretty bad regression.

Anyway, I did some quick profiling under MSVC and so far
 * MSVC says that `std::string` construction takes about 5% of total time spent in `Catch::AssertionHandler::handle`. A quick look says that we pointlessly copy `std::string` because of `AssertionResultData`.
```cpp
    struct AssertionResultData
    {
        AssertionResultData() = delete;

        AssertionResultData( ResultWas::OfType _resultType, LazyExpression const& _lazyExpression );

        std::string message;
        mutable std::string reconstructedExpression;
        LazyExpression lazyExpression;
        ResultWas::OfType resultType;

        std::string reconstructExpression() const;
    };

...
    void AssertionHandler::handle( AssertionResultData const& resultData, ITransientExpression const* expr ) {

        getResultCapture().assertionRun();

[...]

PR #1104, opened 2017-11-15
Performance improvement

coombez:
Begin to address #1086
Brings test from 2m51.072s to 1m15.661s

### DIFF
--- a/include/internal/catch_assertionhandler.cpp
+++ b/include/internal/catch_assertionhandler.cpp
@@ -72,6 +72,13 @@ namespace Catch {
         bool negated = isFalseTest( m_assertionInfo.resultDisposition );
         bool result = expr.getResult() != negated;
 
+        if(result && !getCurrentContext().getConfig()->includeSuccessfulResults())
+        {
+            getCurrentContext().getResultCapture()->assertionRun();
+            getCurrentContext().getResultCapture()->assertionPassed();
+            return;
+        }
+
         handle( result ? ResultWas::Ok : ResultWas::ExpressionFailed, &expr, negated );
     }
     void AssertionHandler::handle( ResultWas::OfType resultType ) {


philsquared:
Possibly - and that might explain some of the regression we've seen - but not all of it.
But on that note - thanks for this @coombez coombez - but I have a local branch here with some more far reaching changes (that include something like your patch) addressing runtime perf. I'm looking to get it finalised and pushed soon.

coombez:
@horenmar Yes, this is almost the same as a change I previously made to Catch 1.x.  Unless it's a really hairy conflict with what @philsquared is doing, it seems like it should be merged back in.  I can't imagine a 4 line change being a hairy conflict though is it?

I have identified a half dozen other areas for optimization as well, and am confident that a billion assertions could be dropped to single digit seconds or even sub seconds.  How long until you believe your changes will be finalized?  Perhaps I should stop looking at this until they're in?

philsquared:
yeah, hold off for now - no sense in us both doing conflicting work.
I'm currently juggling compile time/ runtime trade-offs (I got it down to faster than Catch 1.x runtime speed, but at too great a compile time cost).
At the moment I have something that's probably closer to yours, but with some other organisational changes that may open up further opportunities. I'm probably going to get that in this week.

coombez:
Interesting, I hadn't yet narrowed in on any optimizations that should dramatically impact compile time.  What kind of changes have you been considering in that regard?  It would be useful to know when I do come back to looking at this.  I have a lot of tests so compile time is also very important to me.

philsquared:
I've been looking at reducing the number of entities involved in each assertion. I've not made much progress - but my first attempt actually had a 300% _overhead_ (at least with clang) as I attempted to use lambdas!
There are other areas I'll be looking at too, but I was concentrating on the assertions as I wanted to look at compile/ run time together there.

philsquared:
In fact, after exploring a few dead ends my exploratory branch has got so messy I'm going to abandon it and start selectively applying aspects of it to an up-to-date master.
So, in the meantime, I'm going to take your PR and work back out from there. Thanks!

philsquared:
I've also pushed the first of my other changes, which just inlines `StringRef` constructors. That get's another chunk of the way there.

coombez:
Now down to 0m44.942s with your StringRef inlining.

philsquared:
Thanks, yeah - that's similar to the ratio I see.

I've seen your other PR, too - thanks. I'm actually doing something with `StringReg=` too, at the moment - although slightly different - so I'm not sure if I'll take yours yet (although the other part, with context, looks interesting. I haven't started looking at context yet, but I've been pursing some other opportunities to re-inline dependencies - while balancing the compile-time angle, too).

philsquared:
In fact, on closer inspection, your `StringRef` is mostly orthogonal to what I'm doing, so I'm going accept that, too.

PR #1111, opened 2017-11-20
performance improvement - isTrue, isFalseTest, SourceLineInfo ctor

coombez:
inlined isTrue, isFalseTest and SourceLineInfo ctor.

Brings test from 0m25.584s to 0m20.635s

inlined ~AssertionHandler(), and changed getConfig to return
shared_ptr const reference instead of by value (very expensive for
shared_ptr's).

Brings test from 0m20.635s to 0m17.186s

### DIFF
--- a/include/internal/catch_common.h
+++ b/include/internal/catch_common.h
@@ -43,7 +43,8 @@ namespace Catch {
     struct SourceLineInfo {
 
         SourceLineInfo() = delete;
-        SourceLineInfo( char const* _file, std::size_t _line ) noexcept;
+        SourceLineInfo( char const* _file, std::size_t _line ) noexcept
+        : file( _file ), line( _line ) {}
 
         SourceLineInfo( SourceLineInfo const& other )        = default;
         SourceLineInfo( SourceLineInfo && )                  = default;
@@ -61,7 +62,7 @@ namespace Catch {
     std::ostream& operator << ( std::ostream& os, SourceLineInfo const& info );
 
     // This is just here to avoid compiler warnings with macro constants and boolean literals
-    bool isTrue( bool value );
+    inline bool isTrue( bool value ) { return value; }
     bool alwaysTrue();
     bool alwaysFalse();
 
@@ -82,4 +83,3 @@ namespace Catch {
     ::Catch::SourceLineInfo( __FILE__, static_cast<std::size_t>( __LINE__ ) )
 
 #endif // TWOBLUECUBES_CATCH_COMMON_H_INCLUDED
-
--- a/include/internal/catch_assertionhandler.cpp
+++ b/include/internal/catch_assertionhandler.cpp
@@ -8,10 +8,8 @@
 
 #include "catch_assertionhandler.h"
 #include "catch_assertionresult.h"
-#include "catch_interfaces_capture.h"
 #include "catch_interfaces_runner.h"
 #include "catch_interfaces_config.h"
-#include "catch_context.h"
 #include "catch_debugger.h"
 #include "catch_interfaces_registry_hub.h"
 #include "catch_capture_matchers.h"
@@ -59,12 +57,6 @@ namespace Catch {
     :   m_assertionInfo{ macroName, lineInfo, capturedExpression, resultDisposition }
     {
         getCurrentContext().getResultCapture()->assertionStarting( m_assertionInfo );
-    }
-    AssertionHandler::~AssertionHandler() {
-        if ( m_inExceptionGuard ) {
-            handle( ResultWas::ThrewException, "Exception translation was disabled by CATCH_CONFIG_FAST_COMPILE" );
-            getCurrentContext().getResultCapture()->exceptionEarlyReported();
-        }
     }
 
     void AssertionHandler::handle( ITransientExpression const& expr ) {
--- a/include/internal/catch_assertionhandler.h
+++ b/include/internal/catch_assertionhandler.h
@@ -10,6 +10,8 @@
 
 #include "catch_decomposer.h"
 #include "catch_assertioninfo.h"
+#include "catch_context.h"
+#include "catch_interfaces_capture.h"
 
 namespace Catch {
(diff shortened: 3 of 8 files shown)


horenmar:
Are you also checking what happens with compilation times?

coombez:
I do not have a good benchmark for the compile times as I do for run time, but I am paying attention to make sure the compile time doesn't explode.  Trivial changes like these wouldn't typically affect the compile time, but measuring is always better than intuition :)

Have you guys thought about adding benchmarks for the stuff you care about to your test suite?

include/internal/catch_common.h
```diff
@@ -61,7 +62,7 @@ namespace Catch {
     std::ostream& operator << ( std::ostream& os, SourceLineInfo const& info );
 
     // This is just here to avoid compiler warnings with macro constants and boolean literals
-    bool isTrue( bool value );
+    inline bool isTrue( bool value ) { return value; }
```
janisozaur: Shouldn't that also be marked `static`?

coombez: definitely not, this is not at file scope, and using static for that purpose has been deprecated by the c++ standard.

coombez:
ha!  Just noticed the benchmarkCompile.py.  I'll check it out and add it to my workflow

coombez:
Hmm.  I'm not sure how to use benchmarkCompile.py out of the box...  I'll dig deeper when I have time.
OSError: [Errno 2] No such file or directory: 'benchmark-dir'

philsquared:
This is a tricky one. IIRC `isTrue()` was non-inlined so the compiler wouldn't complain about unreachable code. However that was always a bit of an arms race and may be best left to `#pragma`s - in which case it may not be needed at all in most places. Inlining it is solving the wrong problem, though, I think

horenmar:
@coombez Yeah, `benchmarkCompily.py` is a bit idiosyncratic, I mostly created that one for myself.

Anyway, rough usage guide:
* You need to generate the test files first. This means running it with `-g` or `--generate-files` and `-I <path-to-single-include>`.
* You should now see `benchmark-dir` folder where you ran the script, it should contain `catch.hpp` and `.cpp` files to test compilation
* Run `benchmarkCompile.py` again, this time with an argument specifying which type of compilation you want to check (obviously link requires files to be compiled first). You will also likely need to specify `-f=-std=c++11`.

coombez:
@philsquared All the existing checks pass... does this mean the issue with isTrue() is not covered? or that isTrue() is not needed? or something else all together?  Obviously I don't fundamentally care about it being inlined, it was just consuming 10% of the runtime so it seems like a simple win.

Just as an added data point, in Catch 1.9.6, isTrue is in fact inlined with the same compiler warning comment that it still has.  Perhaps this is still good enough?

martinmoene:
Note: where Catch uses
```
while( Catch::isTrue( false && static_cast<bool>( !!(__VA_ARGS__) ) ) )
```
doctest uses 
```
while((void)0, 0)
```

horenmar:
@martinmoene That formulation suppresses warnings from tested code. Consider these three files

`a.cpp`
```cpp
#include <iostream>

int main() {
    int i = 1;
    if (i == NULL) {
        std::cout << "i = " << i << std::endl;
    }
}
```
`b.cpp`
```cpp
#define CATCH_CONFIG_MAIN
#include "catch.hpp"

TEST_CASE("foo") {
    int i = 1;
    REQUIRE(i == NULL);
}
```
`c.cpp`
```cpp
#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
#include "doctest.h"

TEST_CASE("foo") {
    int i = 1;
    CHECK(i == NULL);
}
```

when compiled with `-Wall -Wextra`, both `a.cpp` and `b.cpp` warn about comparing `int` with `NULL`. `c.cpp` does not.

horenmar:
We could probably throw out the `IsTrue`, but g++'s handling of `_Pragma`s is buggy as hell, so it might mean having to suppress tautological compare in the whole TU instead.

martinmoene:
@horenmar a squint owl that still keeps a sharp look:
```
while( (void)0, 0 && static_cast<bool>( !!(__VA_ARGS__) ) )
```

philsquared:
I get warnings with `(void)0`. However I tested a re-inlined `isTrue()` on a few GCCs, clang and MSVC (2017) with high(ish) warnings (-wall/ -W4) and get no warnings from it - so I've now committed that.

philsquared:
will look at the others when I have a bit more time to run the compiler benchmarks as there can be an interplay.

martinmoene:
Did you really try `while( (void)0,0 && ...` (note the `0,0` ) ?

I've not seen (undesired) warnings with vs2015/2017 -W4, or various gcc/clang -Wall.

philsquared:
@martinmoene hmmm, odd. I just tried it again and this time get no warning with `,` (but I do with `&&`).
I wonder if I left the `&&` in there after previously trying both ways with the other compilers (all in Compiler Explorer).
Maybe that *is* better, then...

coombez:
Results from benchmarkCompile.py:

Before:
Time needed for ...
        ... compiling main, mean: 4.34, median: 4.36 s
        ... compiling test files, mean: 18.68, median: 18.66 s
        ... linking everything, mean: 0.22, median: 0.21 s

After:
Time needed for ...
        ... compiling main, mean: 5.78, median: 5.75 s
        ... compiling test files, mean: 18.63, median: 18.62 s
        ... linking everything, mean: 0.28, median: 0.28 s

martinmoene:
'interesting' ;)

coombez:
I'll have to re-run the benchmarks, that was from the checked in generated header to the generated header after these changes.  I should run the benchmark against the generated header just before these changes instead.

horenmar:
I took the liberty of manually picking parts of this that are still relevant. This means no credit for you (sorry), but it seemed like the least painful option.

PR #1119, opened 2017-12-02
Trivial typo fix.

daniel347x:


horenmar:
Thanks, I rolled in a couple of typos in the vicinity, as single `.` is a bit too trivial.

PR #1150, opened 2018-01-14
CMake fixes

SoapGentoo:
1. Do not build tests with `-Werror` by default
  `-Werror` is good for CI testing but causes massive integrations issues, as compilers add warnings that developers cannot foresee. See also:
    - https://bugs.gentoo.org/644354
    - https://blog.flameeyes.eu/2009/02/future-proof-your-code-dont-use-werror/
2. Install documentation
3. Use CTest to control test suite via `BUILD_TESTING`. The `CTest` module, like `GNUInstallDirs`, provides an idiomatic way to enable and disable the testsuite.
4. Use `GNUInstallDirs` module
  `GNUInstallDirs` is a standardised way to change paths, which makes systems integration easier and allows for a more consistent user experience.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.0)
 
 project(CatchSelfTest)
 
+include(GNUInstallDirs)
+
 option(USE_VALGRIND "Perform SelfTests with Valgrind" OFF)
 option(BUILD_EXAMPLES "Build documentation examples" OFF)
 option(ENABLE_COVERAGE "Generate coverage for codecov.io" OFF)
@@ -366,14 +368,14 @@ if(BUILD_EXAMPLES)
     add_subdirectory(examples)
 endif()
 
-install(DIRECTORY "single_include/" DESTINATION "include/catch")
+install(DIRECTORY "single_include/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/catch")
 
 ## Provide some pkg-config integration
 # Don't bother on Windows
 if(NOT WIN32 OR NOT CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
 
     set(PKGCONFIG_INSTALL_DIR
-        "${CMAKE_INSTALL_PREFIX}/share/pkgconfig"
+        "${CMAKE_INSTALL_DATAROOTDIR}/pkgconfig"
         CACHE PATH "Path where catch.pc is installed"
     )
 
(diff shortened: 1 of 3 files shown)


horenmar:
1. How does GNUInstallDirs deal with Windows?

2. `-Werror` by default is here to stay. Everything outside of `single_include/` is for developer convenience. However, I am willing to add a toggle that prevents it from being used.

3. Removing an externally used option with no warning isn't going to happen.

SoapGentoo:
1. `GNUInstallDirs` just enables the GNU conventions, that is, being able to change installation directories etc. Given that by default, it currently doesn't change anything, users don't see visible changes. This might be important for us if we want to install different versions of Catch side-by-side etc. So however it worked before on Windows, it works the same way still
2. So shall I change the default `ENABLE_WERROR` then?
3. Which externally used option do you mean?

martinmoene:
For users of cmake-gui (and perhaps others) it may be convenient to group the Catch CMake options with a `CATCH_` or `CATCH_OPT_` prefix, like:

- CATCH_OPT_USE_VALGRIND 
- CATCH_OPT_BUILD_EXAMPLES
- CATCH_OPT_ENABLE_COVERAGE
- CATCH_OPT_DISABLE_WERROR

horenmar:
Some other projects (e.g. Microsoft/GSL) use the `NO_SELFTEST` option when obtaining Catch to build it, so removing it without notice is a bad idea\*.

Anyway I am not against changing the CMakelists to accommodate needs of maintainers, but quick testing shows some trouble with the PR as proposed:

* As mentioned before, GSL and other users of `NO_SELFTEST` would have to build all of Catch's tests on install
* Documentation is installed under `CatchSelfTest` folder





\* As much as I think that cloning ~9MB repo instead of downloading ~400KB released header per build is dumb, people do it anyway.

horenmar:
@martinmoene Possibly yeah. I always disliked the "simulate namespacing by prefixing names with strings" approach, but CMake doesn't really have better option as far as I know.

SoapGentoo:
@horenmar I can add some boilerplate code to create a backwards compatibility for `NO_SELFTEST`. Do you mind me changing the name of the project to "Catch2" (or whatever you like), such that the documentation is installed in a suitable directory.

SoapGentoo:
@horenmar all done, the old variables continue to work but display a deprecation notice

horenmar:
I took the liberty of changing the message to be deprecation warning instead of just status.

PR #1155, opened 2018-01-18
More CMake fixes

SoapGentoo:
1. Use `VERSION` specifier in `project()`. This is the modern idiomatic CMake way of specifying the version, for instance see https://rix0r.nl/blog/2015/08/13/cmake-guide/. All regexes for the build scripts have been updated.
2. Namespace all Catch options as suggested by @martinmoene

### DIFF
--- a/scripts/releaseCommon.py
+++ b/scripts/releaseCommon.py
@@ -128,13 +128,12 @@ def updateConanTestFile(version):
         f.write( line + "\n" )
 
 def updateCmakeFile(version):
-    cmakeParser = re.compile(r'set(CATCH_VERSION_NUMBER \d+\.\d+\.\d+)')
     with open(cmakePath, 'r') as file:
         lines = file.readlines()
     with open(cmakePath, 'w') as file:
         for line in lines:
-            if 'set(CATCH_VERSION_NUMBER ' in line:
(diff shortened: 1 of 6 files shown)


scripts/releaseCommon.py
```diff
@@ -128,13 +128,13 @@ def updateConanTestFile(version):
         f.write( line + "\n" )
 
 def updateCmakeFile(version):
-    cmakeParser = re.compile(r'set(CATCH_VERSION_NUMBER \d+\.\d+\.\d+)')
+    cmakeParser = re.compile(r'project(Catch2 LANGUAGES CXX VERSION \d+\.\d+\.\d+)')
```
horenmar: This line can be just deleted completely -- it stopped being used at some point.

SoapGentoo:
@horenmar line removed

horenmar:
Apart from that, `CMake/Findcodecov.cmake` probably shouldn't be changed to make future updates simpler. One possibility is to have `CATCH_ENABLE_COVERAGE` also define `ENABLE_COVERAGE`, the other is to just leave it unprefixed. Personally I'd go with the former.

SoapGentoo:
@horenmar done

Issue #1152, opened 2018-01-14
Don't use -Werror in production


ignatenkobrain:
Adding `-Werror` by default is very very bad idea, because it works fine only on one compiler, in future it might change. `-Werror` should be used **only in CI**.

horenmar:
What production? The released artefact for further consumption is `single_include/catch.hpp` and potentially the reporters in the same folder. Everything else is for development purposes.

ignatenkobrain:
There are people who package catch and they run cmake and that is failing on Fedora with

```cpp
[ 71%] Building CXX object CMakeFiles/SelfTest.dir/include/internal/catch_tag_alias_autoregistrar.cpp.o
/usr/bin/c++   -I/home/brain/Projects/fedora/rpms/catch/Catch2-2.1.0/include  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic   -Wall -Wextra -Wunreachable-code -Werror -std=c++11 -o CMakeFiles/SelfTest.dir/include/internal/catch_tag_alias_autoregistrar.cpp.o -c /home/brain/Projects/fedora/rpms/catch/Catch2-2.1.0/include/internal/catch_tag_alias_autoregistrar.cpp
make[2]: Leaving directory '/home/brain/Projects/fedora/rpms/catch/Catch2-2.1.0/build'
/home/brain/Projects/fedora/rpms/catch/Catch2-2.1.0/include/internal/catch_commandline.cpp: In member function 'Catch::clara::detail::ParserResult Catch::clara::detail::BoundLambda<L>::setValue(const string&) [with L = Catch::makeCommandLineParser(Catch::ConfigData&)::<lambda(bool)>]':
/home/brain/Projects/fedora/rpms/catch/Catch2-2.1.0/include/internal/catch_commandline.cpp:138:67: error: 'temp' may be used uninitialized in this function [-Werror=maybe-uninitialized]
             | Opt( [&]( bool flag ) { config.showDurations = flag ? ShowDurations::Always : ShowDurations::Never; }, "yes|no" )
[...]

PR #1157, opened 2018-01-19
Add value initializer to silence warning

SoapGentoo:
@horenmar Fixes the warning in #1152


philsquared:
Thanks for this. It's the correct fix, but that code is part of Clara, so I've applied it there and will bring a new build of Clara back into catch soon.

See: https://github.com/catchorg/Clara/commit/82ad930b816c7f8a731165c0e02300dc3d6e4091

SoapGentoo:
@philsquared great, pls before the 2.1.1 release!

PR #1159, opened 2018-01-23
Fixing clara Opt namespace

sakamoto-poteko:
Opt is under namespace Catch::clara

## Description

When following the guide of "Adding your own command line options", the compiler reports error of not declaring `Opt`. It is under namespace `Catch::clara`, however, this doc does not reflect that. To fix, either add a line of `using Catch::clara::Opt;` or add `Catch::clara::` directly before `Opt`. For the sake of simplicity, I added the namespace prefix inline before `Opt`.

## GitHub Issues

Not opened.

### DIFF
--- a/docs/own-main.md
+++ b/docs/own-main.md
@@ -83,9 +83,9 @@ int main( int argc, char* argv[] )
   // Build a new parser on top of Catch's
   auto cli 
     = session.cli() // Get Catch's composite command line parser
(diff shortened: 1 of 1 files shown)


philsquared:
Thanks for that! Looks like a copy & paste error.
The intention was to use a `using namespace` so I've done it that way rather than taking this PR as-is.

PR #1167, opened 2018-01-28
Fix pkg config file

### LINKED ISSUE #1164: Change directory layout to make Catch trivially bundleable
Hi @horenmar and @philsquared 
would it be possible to change the `single_include/` structure to make bundling and unbundling of others packages a lot easier for distributions? The reason is the following:

In general, given how your pkg-config file points to `/usr/include` as includedir, the expectation is that users use Catch by doing `#include <catch/catch.hpp>`. If people wish to bundle Catch, they will pull it in as a submodule, and will often do things like `#include "single_include/catch.hpp"`. This makes unbundling Catch from these libraries for distributions a massive pain and requires us to hack buildsystems constantly. Here is my suggestion:

1. rename `single_include/` -> `catch/`
2. `ln -s catch single_include`

With 1. you can just change include paths between a system pkg-config found Catch2 (because the catch prefix is the same now), and 2. keeps compatibility for users of current semantics.

Would you be willing to accept a PR for this?

SoapGentoo:
@horenmar This fixes #1164. After checking the documentation, I realised that `#include <catch.hpp>` was the preferred way to include Catch2, and not `#include <catch/catch.hpp>`. This made things significantly easier, and for that the pkg-config file needs fixing.

Furthermore, I have added a subproject guard to all `install()` directives, which prevents Catch2 from being installed when bundled with another project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.0)
 # detect if Catch is being bundled,
 # disable testsuite in that case
 if(NOT DEFINED PROJECT_NAME)
-  set(NOT_SUBPROJECT ON)
+  set(CATCH_NOT_SUBPROJECT ON)
 endif()
 
 project(Catch2 LANGUAGES CXX VERSION 2.1.1)
@@ -304,7 +304,7 @@ endif()
 
 include(CTest)
 
-if (BUILD_TESTING AND NOT_SUBPROJECT)
+if (BUILD_TESTING AND CATCH_NOT_SUBPROJECT)
     add_executable(SelfTest ${TEST_SOURCES} ${IMPL_SOURCES} ${REPORTER_SOURCES} ${SURROGATE_SOURCES} ${HEADERS})
     target_include_directories(SelfTest PRIVATE ${HEADER_DIR})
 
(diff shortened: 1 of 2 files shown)


lilianmoraru:
Could you please re-run the builds that hang at `CMake` download? So we can get those green ticks?

lilianmoraru:
I like the "subproject" fix but the includepath is debatable.
The tutorial [does mention](https://github.com/catchorg/Catch2/blob/master/docs/tutorial.md#what-did-we-do-here) using `#include "catch.hpp"` but the fully qualified path is more reliable in my own opinion.
What I am saying is that I personally think that it might be better to change the suggestion in the tutorial.

I would also like to note that this is a breaking change(not API-wise but it will break people's compilation).

The **counter-argument**: this would allow to mark this include path as a system include path, thus silencing any warnings for the user, that the headers might generate(ex: future toolchains, additional warnings that the user might use and doesn't want to see from code that they don't "own", etc).

I actually convinced myself. It would be nice to have this include path, even if it is breaking, if instead of `-I[include path]`, it uses `-isystem [include path]`.

SoapGentoo:
1. `-I[include path]` vs `-isystem [include path]` has mostly only to do with warnings and hiding those of dependencies beyond your control.
2. `-isystem` cannot be specified cleanly in build systems (except in CMake, but pkg-config in CMake is a minefield anyway)
3. Whether the API is to use `#include "catch.hpp"` or `#include <catch/catch.hpp>` is the choice of the library designers (although I personally prefer `#include <catch/catch.hpp>` for namepsacing reasons). I think the fact is that when pkg-config support was added, the `/catch` addition to `includedir` was just forgotten.

SoapGentoo:
@horenmar I still think this approach is the best one to avoid changing a ton of `#include` decls, and allwos compatibility with `-isystem`

horenmar:
Sorry for taking so long, I ended up being very busy and there was a discussion going on whether we should keep using `#include <catch.hpp>`, or should move to `#include <catch2/catch.hpp>`. We ended up deciding to go with the latter, so we will be moving documentation/cmake/pkgconfig towards that, starting after next release (where we want to warn people that the change is coming).

Related discussion: #1202

SoapGentoo:
Fixed with https://github.com/catchorg/Catch2/commit/319bddd5b89e0a89e1e6e4badf4930242d081664

PR #1170, opened 2018-01-29
Export catch as cmake package and as a 'linkable' target

djrollins:
## Description

Create a namespaced catch::catch target that is 'linkable' through
`target_link_libraries()` and export it so it is findable through
`find_package()`.

This makes catch a lot easier to use in CMake-based projects. Whether it
is found using `find_package` or included in the client project as a
subdirectory, the client can include the catch headers per-target with
`target_include_directories(target PRIVATE catch::catch).

Also exported is a catch-config-version.cmake file that handles
versions if specified through `find_package(catch 2.1.1 REQUIRED)`.
It will find versions with the exact same major number and with a minor
number >= specified minor number.

LIMITATIONS: As catch.hpp is stored inside the single_include directory,
but installed to include/catch, the `#include` directive changes based
on whether catch is a subdirectory or installed package.

    subdirectory: `#include <catch.hpp>` or `#include "catch.hpp"`
    packaged:     `#include <catch/catch.hpp>`

This could be resolved if catch.hpp was moved into single_include/catch,
but I'll leave that decision to someone else.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.1)
 
 # detect if Catch is being bundled,
 # disable testsuite in that case
@@ -403,3 +403,37 @@ if(NOT WIN32 OR NOT CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/catch.pc DESTINATION ${PKGCONFIG_INSTALL_DIR})
 
 endif()
+
+# add catch as a 'linkable' target
+add_library(Catch INTERFACE)
+
+# depend on some obvious c++11 features so the dependency is transitively added dependants
+target_compile_features(Catch INTERFACE cxx_auto_type cxx_constexpr cxx_noexcept)
+
+target_include_directories(Catch
+	INTERFACE
+		$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/single_include>
+		$<INSTALL_INTERFACE:include/catch>
+		$<INSTALL_INTERFACE:include>)
+
+# provide a namespaced alias for clients to 'link' against if catch is included as a sub-project
+add_library(Catch2::Catch ALIAS Catch)
+
+set(CATCH_CMAKE_CONFIG_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/Catch2")
+
+# create and install an export set for catch target as Catch2::Catch
+install(TARGETS Catch EXPORT Catch2Config DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
+install(EXPORT Catch2Config
+	NAMESPACE Catch2::
+	DESTINATION ${CATCH_CMAKE_CONFIG_DESTINATION})
+
+# install Catch2ConfigVersion.cmake file to handle versions in find_package
+include(CMakePackageConfigHelpers)
+
+write_basic_package_version_file(Catch2ConfigVersion.cmake
+	COMPATIBILITY SameMajorVersion)
+
+install(FILES
+	"${CMAKE_BINARY_DIR}/Catch2ConfigVersion.cmake"
+	DESTINATION ${CATCH_CMAKE_CONFIG_DESTINATION})


djrollins:
Example usage of the target here: https://github.com/djrollins/catch-cmake-usage

SoapGentoo:
1. CMake config files are an anti-pattern that don't scale to other buildsystems and are mostly just a waste of time, and the code this PR adds is 100x more complex than the pkg-config file.
2. pkg-config + `pkg_check_modules` can do all of that already, and is generalisable to other build systems
3. Why do you want to use `target_link_libraries()`? The whole point of Catch2 is to be header-only
4. The pkg-config file should be fixed to specify the `/catch` dir for the `includedir`

lilianmoraru:
I was actually planning to do this today after work :)

A few notes:
- I think `target_include_directories` should do `SYSTEM INTERFACE  include/catch`(not exact command) - `include/catch` because that's how the tutorial wants you to use it and `SYSTEM` to mark this as a system directory, so those that don't "own" the Catch2 source, won't get warnings generated by its source.
I prefer `#include <catch/catch.h>` but I personally think it is more important to mark it as a system directory(which you shouldn't if it is just `include`, which will mark that entire directory as a system directory) than to have the fully qualified path( can't have both :( unless you create a directory structure something like `include/catch/catch/catch.h`, and then mark `-isystem include/catch`, lol ).
- You can avoid the `cmake-config.cmake.in`(and the helper function stuff) if you do:
```cmake
install(
    TARGETS      catch
    EXPORT       FindCatch2
    DESTINATION  lib
)

install(
    EXPORT       FindCatch2
    NAMESPACE    catch::
    DESTINATION  lib/cmake/Catch2
)
```
- I was planning to add this under `NOT_SUBPROJECT`, thus this install step to run only when the project is not used as a subproject.

djrollins:
@SoapGentoo:

Some poeple don't like pkg-config and this just adds another option for people with CMake-based projects and want a familar system for managing dependencies.

In 'Modern CMake' `find_package()` and `target_link_libraries()` are often considered a way of simply stating your dependencies and having all of their dependencies transitively added your your targets through a single unified interface. Whether it is static, dynamic or header-only doesn't matter because CMake should just Do The Right Thing.

I don't suggest that this should replace pkg-config or that it is generalisable to other build systems. I only believe that this makes Catch so much easier to use in CMake-based projects and I like it when libraries I use support it.

@lilianmoraru:

Thanks for the feedback.

I opened this PR early to see if it was anything you'd consider because I love it when libraries support `find_package` and `target_link_libraries`. I am happy to make the changes you have suggested if there is interest this sort of support.

lilianmoraru:
@djrollins I think this PR is a must(considering that most in the C++ community use CMake).
`pkg-config` is just the last resort, this is way better - the product should tell you how you should use it(correct include paths, transitive dependencies, compiler flags, C++ standard dependencies, etc...) and that is what CMake targets are good at. I should just express the target I depend on, not to add include paths manually and then break when a refactoring happens...

Try to define something like the `target_compile_features` line you added, in `pkg-config`...

@djrollins Ignore trolling, this PR is necessary.

djrollins:
@lilianmoraru Thank you for the support.

I've pushed up changes that clean up the export set code and to set the installed include directory to 'include/catch' as per your feedback.

With regards to exporting them as system headers, I looked into it and found this would be redundant as there is already a `system_header` pragma in catch.hpp that suppresses the warnings.

I agree with putting the installation stuff behind an `if(NOT_SUBPROJECT)`, however this alread an open PR for this and I will make the change once that is accepted.

Thanks again for your feedback.

horenmar:
There have been 2 PRs in 2 days for this, so there is definite interest, but I am going to ask the three of you to hash it out and propose one PR.

Looking through them, some notes:

1) I think the name for the package should be something like `Catch2::Catch`, along the lines of `Boost::filesystem` and `Qt::Core`. Because the old name was fairly ungoogleable, we are trying to slowly migrate places to the new (marginally more searchable) name and there is no back-compat to break.
2) `#include <catch.hpp>` vs `#include <catch/catch.hpp>` is a split that kinda happened because it seemed like a good idea at the time. Our preferred version is the first one, but people wanted to have the Catch include "namespaced" by a folder. Because the intended usage was to just version the single header inside user's own repo and CMake integration was an afterthought at best, the change was implemented for CMake and... well here we are.
3) Huge ➕ for the example project using the changes.

horenmar:
@SoapGentoo There is more to this world than Gentoo and other Linux distros.

djrollins:
@horenmar Thank you for the great feedback. I am happy to see that there is interest. You can see all of the `CMAKE_CXX_KNOWN_FEATURES` per-version [here-(https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES). I personally would like to use `cxx_std_11` as Catch2 is advertised as requiring the whole of C++11 standard rather than just specifc parts of it, but that was only introduced CMake 3.8. A huge and unnecessary jump in versions.

@gabm and @lilianmoraru, with respect to horenmar's points, how do you feel about the following:

1) Settling on `Catch2::Catch` as the target name. I think most `find_package` scripts are designed to be searched using the namespace rather than a specific target, so we'd need to also have the export file be called `Catch2Config.cmake` (Moving to PascalCase to match the target name).

2) Exporting both `include` and `include/catch` so that clients have the choice whether they namespace the include directive or not. This keeps the old functionality stable for those that prefer it, but also allows people to use the same unnamespaced include directive whether they are using catch as an installed library or as a sub-project.

3) Do we want to include more features in the feature list? Whilst unnecessary because even just one would be enough to transitively add the required flags, it could be nice to document which features are used in Catch.

gabm:
I agree that requiring the whole C++11 standard is sufficient and reasonable.

1. Catch2::Catch sounds great to me, using namespaces is common practise and makes sense... 

2. sounds good..

3. I always prefer to create minimal solutions to the problems we have and extending more features when needed. Currently i think the availability of package configs is the most important

djrollins:
@gabm Re: point 3, I was talking about the list of features like `cxx_decltype` in the `target_compile_features` command rather than cramming more stuff in this PR. Sorry it wasn’t clear.

gabm:
woops, then: never mind.. sry for the noise.

djrollins:
No worries. Thanks for the feedback. I’m away from a computer until tomorrow afternoon. I’ll make the changes then.

djrollins:
@horenmar @gabm @lilianmoraru 

I have have made the changes outlined above. I have also tested and confirmed installation working on Arch Linux, OSX and Windows. If we're all in agreement then I'll can squash the commits for merging.

Cheers.

lilianmoraru:
@djrollins I don't work with Windows toolchains but from what I know, it doesn't have the concept of C++11, C++14, etc... It just supports features - my guess is that's why CMake added these checks.
`constexpr` could be missing with VC++[some version] Patch 1 but to be available with Patch 2.
Having per feature checks seems better to me, unless it is a big burden to take care of correctly setting and maintaining these.

The rest:
1. No opinion - I think the main maintainers should decide here(although, I do like a simple `find_package(Catch2)`).
2. I personally think that it should be how the tutorial mentions it - adding the "root" include path(to be able to do `catch/catch.hpp`) might introduce surprises(picking headers the user didn't want).
3. Yes?(Helps with Windows support?)
4. No opinion(although I imagine this will complicate the lives of those that do tooling stuff around Catch).

horenmar:
Squash this a bit and I'll merge it.

Figuring out what the do about the include paths will have to come later.

lilianmoraru:
It is squashed now.

horenmar:
Thanks. Would it have any practical effect if all of the C++11 features used in Catch were explicitly listed?

PR #1173, opened 2018-02-03
Remove unnecessary semicolons

mikezackles:
## Description

I have a [meson](http://mesonbuild.com/) build that will build and run the Catch tests when Catch is used as a subproject. Those tests fail to build with the `-Wpedantic` flag because of the extra semicolons removed by this PR. This is very minor, but I figured I'd go ahead and submit in case you want the changes.

(Also, I don't know if you want to support a meson build in addition to CMake, but if that sounds appealing let me know and I will submit a PR.)

Thanks!


mikezackles:
Reopening as the travis failure seems fishy.

horenmar:
Thanks. I'll add `-Wpedantic` to the CI builds.

PR #1176, opened 2018-02-06
Avoid out of line template definition

cor3ntin:
This change lets catch be build as as dynamic library rather than an header.
Out of line template definition prevented to do so as the templates would never be instantiated while building the library and then be undefined when using the library.


horenmar:
After some off-site discussion this PR is effectively abandoned -> I am going to close it.

Issue #1175, opened 2018-02-06
list tests shows hidden


dvirtz:
## Description

Currently --list-tests and --list-test-names-only options list hidden tests as well. I think that the default behavior should be to list the tests that will run when you give the same test spec. 
This is specifically problematic when using the catch_discover_tests CMake command as it uses --list-test-names-only to discover which tests to run and so runs hidden tests as well.

### Steps to reproduce

see above

### Extra information

* Catch version: **v2.0.1**
* Operating System: **Windows 10**
* Compiler+version: **VS 15.6.0**

horenmar:
This makes sense I guess

PR #1179, opened 2018-02-08
don't list hidden tests by default

dvirtz:
resolves https://github.com/catchorg/Catch2/issues/1175


dvirtz:
I unified all the places where the default test spec was defined to one place. 
Now testSpec() always has filters so I added hasTestFilters() method to Config.
I aligned the tests to check the new behavior.

horenmar:
Thanks, looks good, but please don't use 2 space indent next time.

dvirtz:
Thanks!

PR #1181, opened 2018-02-09
the file Catch2ConfigVersion.cmake is contained in ${CMAKE_CURRENT_BI…

BastianDoerig:
## Description

The file Catch2ConfigVersion.cmake is contained in ${CMAKE_CURRENT_BINARY_DIR} and not always in ${CMAKE_CURRENT_DIR}.
Changed the install location to ${CMAKE_CURRENT_BINARY_DIR}

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -434,9 +434,10 @@ install(EXPORT Catch2Config
 # install Catch2ConfigVersion.cmake file to handle versions in find_package
 include(CMakePackageConfigHelpers)
 
-write_basic_package_version_file(Catch2ConfigVersion.cmake
+write_basic_package_version_file(
+	"${CMAKE_CURRENT_BINARY_DIR}/Catch2ConfigVersion.cmake"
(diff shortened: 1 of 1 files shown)


horenmar:
Reading the documentation I think that the top level binary dir is what is actually desired here.

@djrollins @lilianmoraru thoughts?

djrollins:
Having run some tests, it does seem like the current version will break if Catch is installed whilst it is a subproject. The ConfigVersion.cmake file is written into the main projects build directory.

I think the fix I'd prefer is to explicity write the ConfigVersion.cmake file into the `${CMAKE_CURRENT_BINARY_DIR}` dir and leave the `install(FILES)` alone. I think being explict about directories is preferable over relying on CMake's (undocumented?) defaults.

```
write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/Catch2ConfigVersion.cmake
	COMPATIBILITY SameMajorVersion)
```

PR #1199, opened 2018-02-24
Translation to spanish of commercial-users.md

Osovich:
Let me know if there's anything wrong!


JoeyGrajciar:
I think that you should move the translation into folder for spanish e.g `docs/es`. so the original english version remains intact :)

PR #1203, opened 2018-02-28
Fix 'defined but not used' warning

zemasoft:
The warning occurred when !CATCH_CONFIG_WINDOWS_SEH && !CATCH_CONFIG_POSIX_SIGNALS which is the case in my setup.


horenmar:
Ok, thanks.

PR #1216, opened 2018-03-08
CLR Object Stringify

davidaue:
Changes to allow stringification of CLR objects. This enables REQUIRES(a == b) and such print proper strings in reports when used with CLR objects.

I have been using Catch2 in C++/CLI and this helps make it nicer.


horenmar:
Thanks for the PR, I cleaned it up a bit before merging.

PR #1223, opened 2018-03-12
Do not violate UTF-8 in XML output

### LINKED ISSUE #1207: Invalid xml generated by reporter
## Description

The xml reporter is not escaping certain special character in `<Expanded>` section

### Steps to reproduce
Run the following test, using `-r xml` option:
```
#define CATCH_CONFIG_MAIN
#include <catch2/catch.hpp>
#include <cstddef>

using size_type = uint8_t;
constexpr static size_type npos = 255;
size_type foo() { return npos - 1; }

TEST_CASE("Dummy") {
    CHECK(foo() == npos);
}
```

### Extra information

Depending on console/terminal options, the output on console may **not** show special characters. In this is the case, please redirect the output to a file or directly to `less`. For example on my console I can see:
```
$ tests/tests Dummy -r xml
<Catch name="tests">
  <Group name="tests">
    <TestCase name="Dummy" filename="/home/LAN/bronek/Sources/market/tests/main.cpp" line="9">
      <Expression success="false" type="CHECK" filename="/home/LAN/bronek/Sources/market/tests/main.cpp" line="10">
        <Original>
          foo() == npos
        </Original>
        <Expanded>
          '' == ''
        </Expanded>
      </Expression>
      <OverallResult success="false"/>
    </TestCase>
    <OverallResults successes="0" failures="1" expectedFailures="0"/>
  </Group>
  <OverallResults successes="0" failures="1" expectedFailures="0"/>
</Catch>
```
The xml **appears** to be well-formed above, but the reason for failing test is shrouded in mystery. The error becomes visible only if I do:

```
$ tests/tests Dummy -r xml | less
```

The error is also visible for succeeding test cases (if `-s` option is used, of course).

* Catch version: **v2.1.2**
* Operating System: **Linux**
* Compiler+version: **GCC 7.3** and **Clang 5.0.1**

Bronek:
According to UTF-8 encoding rules, there are no valid one byte long codepoints which start with a bigh bit set (i.e. 0x80 or higher value). Hence such XML output needs to be escaped, same as a control character.

## Description
Without this change, xml output may contain invalid UTF-8 codepoints which consist of a single byte and yet have the highest bit set. This is invalid, for example as explained here (https://en.wikipedia.org/wiki/UTF-8#Description). An initial byte of UTF-8 codepoint may have the highest bit set only if it starts a sequence of 2, 3 or 4 bytes. Additionally, under no circumstances can an UTF-8 sequence start with values 0xFC to 0xFD. 

## GitHub Issues
Closes #1207

### DIFF
--- a/include/internal/catch_xmlwriter.cpp
+++ b/include/internal/catch_xmlwriter.cpp
@@ -45,10 +45,14 @@ namespace Catch {
                 default:
                     // Escape control chars - based on contribution by @espenalb in PR #465 and
                     // by @mrpi PR #588
-                    if ( ( c >= 0 && c < '\x09' ) || ( c > '\x0D' && c < '\x20') || c=='\x7F' ) {
+                    // Also take care to avoid violation of UTF-8 encoding rules
+                    if ( ( c >= 0 && c < '\x09' )
+                         || ( c > '\x0D' && c < '\x20')
+                         || ( c & 0x80 )
+                         || c=='\x7F') {
                         // see http://stackoverflow.com/questions/404107/why-are-control-characters-illegal-in-xml-1-0
                         os << "\\x" << std::uppercase << std::hex << std::setfill('0') << std::setw(2)
-                           << static_cast<int>( c );
+                           << static_cast<int>( static_cast<unsigned char>(c) );
                     }
                     else
                         os << c;
--- a/projects/SelfTest/Baselines/compact.sw.approved.txt
+++ b/projects/SelfTest/Baselines/compact.sw.approved.txt
@@ -905,6 +905,8 @@ Xml.tests.cpp:<line number>: passed: encode( stringWithQuotes, Catch::XmlEncode:
 "don't &quot;quote&quot; me on that"
 Xml.tests.cpp:<line number>: passed: encode( "[/x01]" ) == "[//x01]" for: "[/x01]" == "[/x01]"
 Xml.tests.cpp:<line number>: passed: encode( "[/x7F]" ) == "[//x7F]" for: "[/x7F]" == "[/x7F]"
+Xml.tests.cpp:<line number>: passed: encode( "[/x80]" ) == "[//x80]" for: "[/x80]" == "[/x80]"
+Xml.tests.cpp:<line number>: passed: encode( "[/xFF]" ) == "[//xFF]" for: "[/xFF]" == "[/xFF]"
 ToStringVector.tests.cpp:<line number>: passed: Catch::Detail::stringify( empty ) == "{  }" for: "{  }" == "{  }"
 ToStringVector.tests.cpp:<line number>: passed: Catch::Detail::stringify( oneValue ) == "{ 42 }" for: "{ 42 }" == "{ 42 }"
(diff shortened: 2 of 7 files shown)


Bronek:
It appears that `./scripts/approvalTests.py` is reporting lots of bogus differences, or I misunderstood something in `contributing.md`. Anyway, I've updated the commit with hand-crafted changes in the Baselines directory.

horenmar:
Apparently the new VS 2017 update has a bug where `__FILE__` always returns lower cased string and AppVeyor decided to update. We will likely have to improve normalization in the checks.

I am not sure if you want to escape all non-ascii text, or only sequences that do not make a valid UTF-8 sequence, because your PR as it stands does the former -- the result of printing this `Here be 👾` to console using changed xml reporter is "Here be \xF0\x9F\x91\xBE".

Bronek:
You make a fair point, but we can do very little about it. C++ support for UTF-8 encoding is in `char` type, and there is no way to mark a `char` (or, more usefully a `std::string`) as containing a UTF-8 encoded text (rather than something else, which **might** appear to be UTF-8 - for example CESU-8). However, since XML output from Catch is encoded in UTF-8, you have to take care of encoding of whatever `char` or `std::string` the users chose to compare. So the choices are:

* output a number, like we do for other non-printable characters (as proposed here)
* perform a heuristic on a string and choose to either output a number or a UTF-8 string as-is. This would have to be able to recognize valid UTF-8, including recognizing the [difference between CESU-8 and UTF-8](http://unicode.org/faq/utf_bom.html#utf8-4) (and I do not see how it could be done in a robust and compact manner)
* extend the interface to allow users to output the strings as-is, including characters which do not fit in UTF-8 (this may lead to corrupt XML output which cannot be displayed if a test is failing - i.e. exactly when users are interested in output the most)
* perform heuristic **or** extend the interface to explicitly support UTF-8 in user supplied strings, but rather than output literally, use \u form instead (so no risk of corruption of XML output, but still plenty of work or change of interface)

While we discuss the above options, do you have other solution which could be applied if the user wanted to perform a check on an **integral** type such as e.g. `uint8_t` or `int8_t` , as reported originally? Note, overloading for those is not a great solution, because in some implementation either of these could typedef to char.

Bronek:
I do not know how to recognize a valid UTF-8 sequence, and looking at [Unicode FAQ](http://unicode.org/faq/utf_bom.html#utf8-4) I suspect this can be pretty difficult. Also, what do you want to do if user simply wanted to perform a check on an integral type such as e.g. `uint8_t` or `int8_t`, as reported originally? Note, overloading for those is not a great solution because in some implementation either of these could a `typedef char`. So, in the interest of simplicity, I think it is best to simply show a numeric sequence.

Bronek:
I found a rather compact UTF-8 validity test at http://bjoern.hoehrmann.de/utf-8/decoder/dfa/ which could be ported/added to Catch2, but I suppose this is outside of the scope of the issue being fixed here.

horenmar:
Breaking valid UTF-8 output is unacceptable.

9171e9b55c26515d012364ae61ae3a836ca597d3 modifies the XML encoder to deal with UTF-8 semi-properly instead and I am going to merge it to master once I add more tests.

Bronek:
Thank you, I reviewed your code and it looks good.

PR #1228, opened 2018-03-19
fix typo in WithinAbs::match() and add selftests that would have failed previously

Jofagi:
## Description
I encountered a condition in `WithinAbs()` that should not fail from my understanding of that feature.
Looking into the code, I believe I found the issue.

Added a few selftests that would have pointed out this problem before the fix.

Obviously `approvalTests.py` fails and reports theses new tests but guess I'm not supposed to approve these differences.


horenmar:
Nice find.

As to the approval tests, you actually are supposed to approve them and commit them alongside the changes. It will still fail on Windows anyway, because current VS 2017 has a bug where it reports wrong paths, so the approvals fail, but it should pass on Travis.

horenmar:
I squashed the commits and merged them, thanks.

Issue #1202, opened 2018-02-26
#include <catch2/catch.hpp> vs #include "catch.hpp"


MikeGitb:
Just wondering: It seems more ideomatic to me to include the library name as part of the include name instead of just using the header file name alone. On the other hand, for a single header library there is of course not a big need/difference for it unless the header name is likely to clash with a different file name.

So I was wondering whether your preference of 

    #include "catch.hpp"

over

    #include <catch2/catch.hpp>

was deliberate or more like "for historic reasons" / "didn't care".

Quincunx271:
For single-header libraries, it's common for it to just be

```
#include <libname.hpp>
```

Alternatively, sometimes when the name is multiple words:

```
#include <super/awesome/library.hpp>
```

The code doesn't read as well, and people will be annoyed when writing it, if you had to write:

```
#include <libname/libname.hpp>
#include <super/awesome/library/library.hpp>
```

MikeGitb:
Considering that a test file usually contains much more code than just the include I find it strange that people get annoyed if they have to type a few characters more once in a while, but I'm using an IDE with auto complete, so that might skew my view. 

Thanks for the explanation anyway. I'll consider it for my own library.

martinmoene:
Or : #include <_namespace_/_filename_>

See also [Directories = namespaces?](https://arne-mertz.de/2016/06/organizing-directories-namespaces/#Directories_namespaces) by Arne Mertz.

SoapGentoo:
Personally I don't mind either way from a distribution point of view - just be aware that if you're supposed to use `catch2/catch.hpp`, then trying to use it with CMake's `SYSTEM` directive will cause `-isystem /usr/include` to go up in a big explosion, due to the way GCC handles special system headers (`math.h` etc). Staying with the current approach would allow `-isystem`. Personally I don't consider `-isystem` compatibility that important, given how Catch tries to be warning-free by design.

MikeGitb:
@SoapGentoo: Maybe I missunderstood you, but we might think of different implementation strategies:

If a packet manager / distributor wants to switch from "catch.h" to "catch2/catch.h" the way that seems most logical to me would be to move the file from "$BaseDir/catch.h" to "$BaseDir/catch2/catch.h". Any include directives should stay the same as before, so this should not pose any new problems. This might also be beneficial if someone wants to install both catch and catch2 in parallel (or it might just be more confusing).

[...]

PR #1239, opened 2018-04-01
Small fix to generate pc with all include path

jeandet:
In CMake module both include and include/catch are added to includes
lookup path. Examples are built with #include "catch.hpp" not #include "catch/catch.hpp". 
This should be the same with pkg-config.
Fixing pc file would ease Catch2 use from build systems like [Meson](http://mesonbuild.com/index.html).

### DIFF
--- a/catch.pc.in
+++ b/catch.pc.in
@@ -3,4 +3,4 @@ includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
 Name: Catch
 Description: Testing library for C++
 Version: @Catch2_VERSION@
-Cflags: -I${includedir}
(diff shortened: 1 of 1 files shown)


horenmar:
I am not sure if I want to merge this. We have plans to move to `#include <catch2/catch.hpp>` everywhere soon(ish), see #1202.

jeandet:
@horenmar that's fine to me as long as pkg-config and CMake module do the same.

horenmar:
Upon further thought, I've decided to merge this. While we will remove it in the future, it is probably better to have the includes uniform in the meantime.

PR #1244, opened 2018-04-06
Patch to support python 3

tomhughes:
The new approval test is failing when run under python 3 with:

```
8: Running approvals against executable:
8:   /home/tom/src/fedora/catch/Catch2-2.2.2/build/SelfTest
8: 
8: console.std:
8:   Results matched
8: 
8: Traceback (most recent call last):
8:   File "/home/tom/src/fedora/catch/Catch2-2.2.2/scripts/approvalTests.py", line 177, in <module>
8:     approve("console.sw", ["~[!nonportable]~[!benchmark]~[approvals]", "-s", "-w", "NoAssertions", "--order", "lex"])
8:   File "/home/tom/src/fedora/catch/Catch2-2.2.2/scripts/approvalTests.py", line 144, in approve
8:     for line in rawFile:
8:   File "/usr/lib64/python3.6/codecs.py", line 321, in decode
8:     (result, consumed) = self._buffer_decode(data, self.errors, final)
8: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 1867: invalid continuation byte
8/8 Test #8: ApprovalTests ....................***Failed    0.28 sec
```

The problem is that there are (presumably deliberate) invalid UTF8 sequences in the output.

This patch (tested with python 2.7.14 and 3.6.4) uses the "surrogateescape" error handler to encode any invalid bytes as special surrogates when reading and writing the output files.


scripts/approvalTests.py
```diff
@@ -73,9 +73,9 @@
 overallResult = 0
 
 def diffFiles(fileA, fileB):
-    with open(fileA, 'r') as file:
+    with open(fileA, 'r', errors='surrogateescape') as file:
```
marcusnaslund: `surrogateescape` does not exist in python 2, though

tomhughes: I wasn't expecting it to be yet it appeared to work, but I think I must have done something wrong because I just tried again and it didn't.

I'll see what I can do...

horenmar: IIRC you have to use `io` module for python 2.7 (same as for working with encodings), so you have to make a small polyfill over `open`/`io` usage.

tomhughes: Well the python 2 case doesn't actually need to worry about the encoding but a polyfill is the answer yes and I've done that now.

PR #1245, opened 2018-04-10
Support for parenthesizing types with commas.

BRevzin:
Currently, you cannot use template specializations with commas in them as fixtures. This because you need to parenthesize them in order for the macro to treat it as a single argument, and that single argument is used as a base class type specifier... and you can't have parentheses there.

In other words, this doesn't work today:

    template <typename, typename>
    struct Fixture { };

    TEST_CASE_METHOD((Fixture<int, int>), "a", "[b]") {
        CHECK(1 == 1);
    }

With less than helpful compile errors like:

```
foo.cxx:7:18: error: expected class-name before ‘(’ token
 TEST_CASE_METHOD((Fixture<int, int>), "a", "[b]") {
                  ^
foo.cxx:7:18: error: expected ‘{’ before ‘(’ token
foo.cxx:7:36: error: invalid declarator before ‘)’ token
 TEST_CASE_METHOD((Fixture<int, int>), "a", "[b]") {
                                    ^
foo.cxx:7:1: error: incomplete type ‘{anonymous}::____C_A_T_C_H____T_E_S_T____0’ used in nested name specifier
 TEST_CASE_METHOD((Fixture<int, int>), "a", "[b]") {
 ^
foo.cxx:7:1: error: invalid use of incomplete type ‘struct {anonymous}::____C_A_T_C_H____T_E_S_T____0’
 TEST_CASE_METHOD((Fixture<int, int>), "a", "[b]") {
 ^
foo.cxx:7:1: note: forward declaration of ‘struct {anonymous}::____C_A_T_C_H____T_E_S_T____0’
 TEST_CASE_METHOD((Fixture<int, int>), "a", "[b]") {
 ^
```

This PR makes this work by introducing a new type trait which is a spot that _does_ allow for extra parenthesis. Both `void(T)` and `void((T))` are valid types, so we can take advantage of that to pull out the `T` properly. All previously functional code should remain functional.

### DIFF
--- a/docs/assertions.md
+++ b/docs/assertions.md
@@ -6,6 +6,7 @@
 [Exceptions](#exceptions)<br>
 [Matcher expressions](#matcher-expressions)<br>
 [Thread Safety](#thread-safety)<br>
+[Expressions with commas](#expressions-with-commas)<br>
 
 Most test frameworks have a large collection of assertion macros to capture all possible conditional forms (```_EQUALS```, ```_NOTEQUALS```, ```_GREATER_THAN``` etc).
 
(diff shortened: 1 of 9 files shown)


horenmar:
I wonder how many people actually ran into this problem, but it is a nice idea 👍 

I see there are no tests though.

horenmar:
As it turns out, this still won't allow the code to compile under MSVC

BRevzin:
I never use MSVC so I don't know a clever way of fixing it. This at least seems to fix this for gcc/clang, so still worthwhile?

horenmar:
I will have to think about it. Without the workaround working at least on the big three, I wonder if it wouldn't be better to point users to typedef-based workaround instead:

```cpp
template <typename, typename>
struct Fixture { };

using fixture_int_int = Fixture<int, int>;

TEST_CASE_METHOD(fixture_int_int, "a", "[b]") {
    SUCCEED();
}
```

BRevzin:
Here's a solution with just macros that seems to compile across all gcc, clang, and msvc (at least the versions I tried).

BRevzin:
I... don't understand this build output. Seems like the fact that we added one more test, that passed, broke things?

horenmar:
@BRevzin See our [contributing.md](https://github.com/catchorg/Catch2/blob/master/docs/contributing.md#testing-your-changes) for explanation.

But in reality that one is on me. Since the test didn't compile on my primary platform, I didn't bother updating approval tests from Linux.

As to the new version of the workaround, I will merge it in later today.

PR #1250, opened 2018-04-15
Added export(...) commands to add Catch to CMake package registry

samaursa:
## Description
Without `export(...)`, CMake's `find_package` fails to find an installed `Catch` library because it doesn't exist in the global registry (which, on Linux for e.g., exists at `~/.cmake/packages`) without providing `HINTS`.


horenmar:
No.

The problem with `EXPORT` commands (as opposed to `install(EXPORT)`, because it wouldn't be CMake if things were simple and obvious) is that
* It registers packages in user-specific repository
* It does not install utility scripts the way the proper install does
* It points to the _build_ tree

The last one is the largest problem, as it means that if you clone the repository and later delete, or move, it, `EXPORT`ed targets will no longer work. Instead, users should install the current package, either globally with admin rights, or locally by setting `CMAKE_INSTALL_PREFIX`.

PR #1254, opened 2018-04-18
Changed to c++ style includes

marcusnaslund:
## Description

Now includes the c++ headers `cassert`, `csignal` and `cstdio` instead of `assert.h`, `signal.h` and `stdio.h`.
I've had problems in another multi-library project when moving to newer versions of clang when _not_ using c++ headers, although these aren't the usual culprits.

A rationale and basis for discussion is here: https://stackoverflow.com/questions/43011737/is-it-better-to-include-cassert-or-assert-h?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

### DIFF
--- a/include/internal/catch_test_case_tracker.cpp
+++ b/include/internal/catch_test_case_tracker.cpp
@@ -10,7 +10,7 @@
 #include "catch_enforce.h"
 
 #include <algorithm>
-#include <assert.h>
+#include <cassert>
 #include <stdexcept>
 #include <memory>
 #include <sstream>
--- a/include/reporters/catch_reporter_bases.cpp
+++ b/include/reporters/catch_reporter_bases.cpp
@@ -13,7 +13,7 @@
 #include <cstring>
 #include <cfloat>
 #include <cstdio>
-#include <assert.h>
+#include <cassert>
 #include <memory>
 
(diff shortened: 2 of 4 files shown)


marcusnaslund:
Interesting, appveyor fails consistently because a path is different in the output of one mode.
Tests pass great on travis and locally. Could something be wrong in the test config?

horenmar:
VS 2017 update has a bug that breaks approval tests, so those builds are currently allowed to fail.

Anyway, changing `<assert.h>` to `<cassert>` is okay, but the other changes are not -- some platforms (the usual culprit is QNX) actually enforce namespace separation between `<stdio.h>` and `<cstdio>`, such that the former provides only non-namespaced functions (e.g. `printf`) and the latter provides only namespaced ones (e.g. `std::printf`). This means that without also properly namespacing all uses of C stdlib, the changes have potential to break compilation on some platforms.

marcusnaslund:
Oh, that's unfortunate. I see the comment about why you're using `sprintf` over `stringstream` as well.
Alright then, I'll change to just keep assert includes.

marcusnaslund:
There we go.

horenmar:
Ok, this version will work.

PR #1258, opened 2018-04-23
Add static library target for efficient recompilation

jpcima:
I add a static library target for catch, as an alternative to the single header solution.
I created this change to save build time and memory on computers and VMs which have limited CPU and RAM resource. This compiles the internal classes separately a single time, rather than with each test program.

I took one of the example programs and compared header-only catch vs static library.
As you can judge by results, a recompilation of the test took 7% of time and 24% of the memory.
The productivity gain is quite considerable.

```
#/usr/bin/time -f "real:%es user:%Us sys:%Ss max-rss:%MkB" make 010-TestCase
real:39.92s user:37.12s sys:1.24s max-rss:693944kB
#/usr/bin/time -f "real:%es user:%Us sys:%Ss max-rss:%MkB" make 010-TestCase-static
real:2.92s user:2.43s sys:0.37s max-rss:163224kB
```

This was the project used for a benchmark.

```
cmake_minimum_required(VERSION 3.1)

add_subdirectory("../Catch2" Catch2.dir)

add_executable(010-TestCase-static 010-TestCase.cpp)
target_link_libraries(010-TestCase-static PRIVATE Catch2::Catch-static)

add_executable(010-TestCase 010-TestCase.cpp)
target_link_libraries(010-TestCase PRIVATE Catch2::Catch)
```

Edit: I make this an object/interface library target. A static library is not going to work, because the linker will discard some required symbols which initialize the framework.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -424,7 +424,8 @@ endif()
 add_library(Catch INTERFACE)
 
 # depend on some obvious c++11 features so the dependency is transitively added dependants
-target_compile_features(Catch INTERFACE cxx_auto_type cxx_constexpr cxx_noexcept)
+set(COMPILE_FEATURES cxx_auto_type cxx_constexpr cxx_noexcept)
+target_compile_features(Catch INTERFACE ${COMPILE_FEATURES})
 
 target_include_directories(Catch
 	INTERFACE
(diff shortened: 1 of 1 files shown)


onqtam:
I don't see a reason for the linker to remove anything if you use Catch in its single header form when building the static library. Also you could compile the framework header with the test runner implementation in some source file into a static library in your own projects - should be pretty easy with any proper build system

martinmoene:
In [examples/CMakeLists.txt](https://github.com/catchorg/Catch2/blob/c651f239f01dde9e132b1a3d06fd9b1b6246b189/examples/CMakeLists.txt#L71) Catch is already compiled once for those examples that intentionally follow the *idiomatic way* to compile Catch's source once separately from the tests themselves. This *idiomatic way* is also intentionally not used for several simple examples.

jpcima:
Thank you for explaining this. While a separate compilation of `main` is certainly helpful, and nullifies some of the points which I have made, one remains:
A compilation of the catch library as a single unit is demanding in terms of memory. On a build where cpp files are built in separation, I measure a diminution of peak memory by roughly 52%.

>I don't see a reason for the linker to remove anything if you use Catch in its single header form when building the static library.

It's what I thought initially; but static initializers will be discarded unless the main program contains references to them. You would have this startup error: `No reporter registered with name: 'console'`
An object library is a portable solution requiring no linker flags.

>Also you could compile the framework header with the test runner implementation in some source file into a static library in your own projects

It's absolutely correct; I can have these rules in my own project. I suggest them as a convenience.

Issue #537, opened 2015-11-10
Init / cleanup in tests?


rcdailey:
Is there a mechanism to perform initialization before any test cases run, and cleanup after they are done?

How is this being accomplished today?

nabijaczleweli:
```c++
struct test_init {
  test_init() {
    // perform your init here
  }
} test_init_instance;
```

rcdailey:
That will happen at static initialization time. That might cause issues. It would be better if catch's test execution worked like so:

```
int main()
{
  InitTests();
  Run();
  CleanupTests();
}
```

*Yes,* I know I can implement this myself, but I'm just asking if there is a built-in alternative. I dug through the source code and I didn't see anything but just want to be sure.

jfemia:
Implementing your own main is the documented way. It is how I accomplish the global setup in my test runner.

Darelbi:
There's no need for setup teardown:

    TEST_CASTE(){
        // setup
        SECTION(){

        }

        // TEARDOWN
    }


the setup and teardown are implicit in how catch work (wich is what I like most). Of course the same apply to nested cases:

    TEST_CASTE(){
        // setup A
        SECTION(){  // A

            // setup B,C
            SECTION(){ //B

            }

            SECTION(){ // C

            }
            // TEARDOWN B,C
        }

        // TEARDOWN A
    }

philsquared:
Further to @jfemia: https://github.com/philsquared/Catch/blob/develop/docs/own-main.md

airelil:
A remark about @Darelbi  comments. The "teardown" part of a test doesn't always work. On my MS VS2010 machine (catch ver. 1.21.) the following code produces memory leak because "arr" isn't deleted. It leaves the "section A" on REQUIRE exception immediately and doesn't continue to the "delete" part. Oddly though, I see that it enters the "test A" right after that and this time performs new/delete cycle without entering into SECTION.
```C++
TEST_CASE("test A")
{
  char *arr = new char[8];
  SECTION("section A")
  {
    REQUIRE(1 == 0);
  }
  delete []arr;
}
```
So I had to revert to ```st::unique_ptr```
```C++
TEST_CASE("test B")
{
  std::unique_ptr<char[]> arr(new char[8]);
  SECTION("section B")
  {
    REQUIRE(1 == 0);
  }
}
```

nabijaczleweli:
How do you "revert" to `std::unique_ptr`?

philsquared:
[...]

PR #1261, opened 2018-04-24
New init / deinit listener methods

rcdailey:
Relates to #537

### DIFF
--- a/include/internal/catch_interfaces_reporter.h
+++ b/include/internal/catch_interfaces_reporter.h
@@ -203,12 +203,48 @@ namespace Catch {
 
         virtual void skipTest( TestCaseInfo const& testInfo ) = 0;
 
+        /// Invoked prior to "leaf" test logic being executed for global initialization. "leaf" in
+        /// this case refers to a lone TEST_CASE or nested SECTION, depending on the structure of
+        /// the test itself. A "step" in the test is 1 leaf. The purpose of this method is to
+        /// perform identical global initialization prior to every test step's execution. This is
+        /// designed to happen before any local objects in a test step are constructed.
+        virtual void testStepInit( TestCaseInfo const& ) {}
+
+        /// Invoked after a "leaf" test to perform global cleanup. This is executed after all local
+        /// objects within test step scope have been destroyed. This method is meant to allow
+        /// shared global cleanup of state between test step executions.
+        virtual void testStepDeinit( TestCaseInfo const& ) {}
+
         // Default empty implementation provided
         virtual void fatalErrorEncountered( StringRef name );
 
         virtual bool isMulti() const;
     };
+
     using IStreamingReporterPtr = std::unique_ptr<IStreamingReporter>;
+
+    /// Utilizes RAII to invoke IStreamingReporter::testStepInit() and
+    /// IStreamingReporter::testStepDeinit(). This is useful in cases where exceptions can occur and
+    /// you want to make sure that the deinit method is still invoked.
+    class ScopedReporterStep
+    {
+    public:
+        ScopedReporterStep(IStreamingReporter& reporter, TestCaseInfo const& info)
+            : m_reporter(reporter)
+            , m_info(info)
+        {
+            m_reporter.testStepInit(m_info);
+        }
+
+        ~ScopedReporterStep()
+        {
+            m_reporter.testStepDeinit(m_info);
+        }
+
+    private:
+        IStreamingReporter& m_reporter;
+        TestCaseInfo const& m_info;
+    };
 
     struct IReporterFactory {
         virtual ~IReporterFactory();
--- a/include/internal/catch_run_context.cpp
+++ b/include/internal/catch_run_context.cpp
@@ -299,6 +299,7 @@ namespace Catch {
 
     void RunContext::invokeActiveTestCase() {
         FatalConditionHandler fatalConditionHandler; // Handle signals
+        ScopedReporterStep step{*m_reporter, m_activeTestCase->getTestCaseInfo()};
         m_activeTestCase->invoke();
         fatalConditionHandler.reset();
     }
--- a/include/reporters/catch_reporter_listening.cpp
+++ b/include/reporters/catch_reporter_listening.cpp
@@ -91,6 +91,21 @@ namespace Catch {
         m_reporter->assertionStarting( assertionInfo );
     }
 
+    void ListeningReporter::testStepInit( TestCaseInfo const& testInfo ) {
+        for (auto const& listener : m_listeners) {
+            listener->testStepInit( testInfo );
+        }
+        m_reporter->testStepInit( testInfo );
+    }
+
+    // The return value indicates if the messages buffer should be cleared:
+    void ListeningReporter::testStepDeinit( TestCaseInfo const& testInfo ) {
+       for (auto const& listener : m_listeners) {
+          static_cast<void>(listener->testStepDeinit(testInfo));
+       }
+       return m_reporter->testStepDeinit(testInfo);
+    }
+
     // The return value indicates if the messages buffer should be cleared:
     bool ListeningReporter::assertionEnded( AssertionStats const& assertionStats ) {
         for( auto const& listener : m_listeners ) {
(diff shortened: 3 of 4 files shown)


rcdailey:
I decided to go ahead and create an init / deinit method pair, similar to the started/ended methods we currently have in reporters. This allows setup and teardown logic to occur per "leaf" test (could be a section or test case)

rcdailey:
I amended a change that enables the `testStepDeinit()` method to be invoked when a test case throws an exception (utilizing an RAII object).

@philsquared I opened this 16 days ago and have continued to keep it updated. When can I expect someone to have time to get me some feedback and possibly merge the change?

rcdailey:
@horenmar If maybe phil is not available, would you be able to contribute some feedback?

rcdailey:
Thanks Phil, and again sorry to rush, I just wanted to make sure any major suggestions are made as early as possible so I can make any changes you need. I've written unit tests on my side to depend on these changes, so in order to avoid our code depending on it too long, I wanted to get it reviewed & merged with the final interface (if changes are needed).

I can be patient. Hopefully someone will have time for this soon. Thanks for replying back!

onqtam:
Hi!

Isn't what you want already possible with ```testCaseStarting```/```sectionStarting``` and ```testCaseEnded```/```sectionEnded```?

Or if the test case related methods are called only once even for a test case with multiple sections - isn't that what the ```testGroupStarting```/```testGroupEnded``` are for?

Seems to me that the behavior you are looking for can be emulated in a custom reporter/listener with some custom logic and overriding these methods... Or perhaps I'm missing something.

rcdailey:
@onqtam The way these various functions work makes sense from a reporting perspective, but to do any kind of functional (program-affecting) logic, they're not sufficient. My new methods utilize RAII to make sure the init happens prior to entering a test case, and that deinit happens after the closing scope of the test case. They enable you to do setup/tear-down logic at a global level in the window of time where no objects should be alive in any active test cases or sections. They also allow tear-down to happen when exceptions are thrown.

There were 2 main problems I couldn't solve with the existing reporter methods:

* Singletons could not be released between each test case or section
* Working directory files could not be deleted

To execute any of the above actions requires that no active objects in a test case or section be alive. You can't tear-down in `sectionEnded()`, for example, because this is executed prior to the test case scope ending (objects still alive at the test-case level) and you can't use `testCaseEnded()` because this is not executed more than once with 2 or more sections inside it.

philsquared:
Sorry - been ultra busy for a while (again) - just finished with one (work) release, so have a brief moment of downtime - thanks for the ping offline...

There's a couple of ways you could achieve the same goal without changing anything.

1. You could keep a count of section nesting (++ in `sectionStarting`, -- in `sectionEnded`. Before and after, respectively, if you check for 0 you know you're dealing with the section call that corresponds to the test case.
2. You could use a(n old school) fixture class and do your init/ deinit in the ctor/ dtor of that.

If necessary I could maybe add a flag to `SectionInfo` that indicates if it's for a `TEST_CASE` or not, so you could check that.

Give that we already make a reporter/ listener call for this event (`sectionStarting`/`Ended`) I'd rather not add an extra one - especially given there are a couple of ways to do it that aren't too burdensome. I admit the first option is a little less obvious - but the second one, at least, seems to be the right fit?

rcdailey:
Do you have an example of the fixture class? I'm not understanding how that
fixes the issue.

rcdailey:
@philsquared Ping? Right now I'm doing your 1st suggestion:

```
void GlobalStateCleanup::sectionStarting(Catch::SectionInfo const& _sectionInfo)
{
   ++m_sectionDepth;
}

void GlobalStateCleanup::sectionEnded(Catch::SectionStats const&)
{
   if (m_sectionDepth > 0 && --m_sectionDepth == 0)
   {
      DeleteGlobalState();
   }
}
```

While this is simple enough, I'd still like to see a code example of the fixture class you were talking about. I've never done that in Catch and not familiar with it.

philsquared:
You are correct that it requires all test cases that you want to partake in the common setup/ teardown to use `TEST_CASE_METHOD`. I can see that retrofitting that might not be optimal - so I think the `sectionEnded()` technique with the  `m_sectionDepth` trick is probably the best workaround.

I don't think this is a common enough need to add an extra pair of handlers to every reporter, so I'm reluctant to do that, given that the workaround is ok. Are you satisfied with that approach?

rcdailey:
I'm starting to realize this doesn't solve my problem. Basically, my code change addresses cleanup at the leaf-section level, whereas this solution only does the cleanup after the root section (the test case). If I have a test case with 3 sections inside of it, I need all 3 sections inside to get cleaned up, but the outer test case (or even intermediate sections) do not need cleanup since they would have already been cleaned up by the nested section. Does this make sense?

philsquared:
I'm afraid I'm not quite following. Perhaps you could sketch a simple outline example we could talk about in more concrete terms?

philsquared:
(FTR, AFAICS, the `m_sectionDepth` trick should give you the same semantic as your proposed init/deinit methods)

rcdailey:
The more I'm thinking about this, I think the more I might be getting confused. Let me write an MCVE and get back with you

rcdailey:
Ok you're right, this is the same. Here's a live sample:
https://wandbox.org/permlink/b8XwAfXxopzhU6Fw

This code:

```cpp
class GlobalStateCleanup : public Catch::TestEventListenerBase
{
public:

    // Inherit constructor
    using TestEventListenerBase::TestEventListenerBase;

    void sectionStarting(Catch::SectionInfo const&) override
    {
        ++m_sectionDepth;
    }

    void sectionEnded(Catch::SectionStats const&) override
    {
        if (m_sectionDepth > 0 && --m_sectionDepth == 0)
        {
            DeleteGlobalState();
        }
    }

private:

    void DeleteGlobalState()
    {
        std::cout << "Global State Deleted\n";
    }

    unsigned m_sectionDepth{};
};

CATCH_REGISTER_LISTENER(GlobalStateCleanup)

TEST_CASE("Root Test", "[test]")
{
    std::cout << "Begin Test Case\n";
    
    SECTION("Section One")
    {
        std::cout << "Section One\n";
    }
    
    SECTION("Section Two")
    {
        std::cout << "Section Two\n";
    }
    
    std::cout << "End Test Case\n";
}
```
Yields this output:

```
Begin Test Case
Section One
End Test Case
Global State Deleted
Begin Test Case
Section Two
End Test Case
Global State Deleted
test cases: 1 | 1 passed
assertions: - none -
```

Would it be useful to create a generic Listener for this in contrib? Basically something like:

```cpp

class GlobalStateCleanup : public Catch::TestEventListenerBase
{
public:

    // Inherit constructor
    using TestEventListenerBase::TestEventListenerBase;
    
    virtual void initTest(Catch::SectionInfo const&) {}
    virtual void deinitTest(Catch::SectionStats const&) {}

private:

    void sectionStarting(Catch::SectionInfo const& info) final
    {
        if (m_sectionDepth++ == 0)
        {
            initTest(info);
        }
    }

    void sectionEnded(Catch::SectionStats const& stats) final
    {
        if (m_sectionDepth > 0 && --m_sectionDepth == 0)
        {
            deinitTest(stats);
        }
    }

    unsigned m_sectionDepth{};
};
```

That way anyone that wants to do this only has to implement `GlobalStateCleanup` and implement either of `initTest()` and `deinitTest()`? If you think it's a good idea, maybe help me figure out some better names and I'm happy to open a new PR. Thanks @philsquared !!

rcdailey:
Better example:

```cpp
class TestInitDeinitListener : public Catch::TestEventListenerBase
{
public:

    using TestEventListenerBase::TestEventListenerBase;
    
    virtual void initTest(Catch::SectionInfo const&) {}
    virtual void deinitTest(Catch::SectionStats const&) {}

private:

    void sectionStarting(Catch::SectionInfo const& info) final
    {
        if (m_sectionDepth++ == 0)
        {
            initTest(info);
        }
    }

    void sectionEnded(Catch::SectionStats const& stats) final
    {
        if (m_sectionDepth > 0 && --m_sectionDepth == 0)
        {
            deinitTest(stats);
        }
    }

    unsigned m_sectionDepth{};
};

class GlobalStateCleanup : public TestInitDeinitListener
{
public:

    using TestInitDeinitListener::TestInitDeinitListener;
    
    virtual void initTest(Catch::SectionInfo const&)
    {
        std::cout << "Init Test\n";
    }
    
    virtual void deinitTest(Catch::SectionStats const&)
    {
        std::cout << "Deinit Test\n";
    }
};

CATCH_REGISTER_LISTENER(GlobalStateCleanup)
```

PR #1266, opened 2018-04-29
Cast to unsigned char when using std::isalnum

ianhattendorf:
std::isalnum expects an int in the range of unsigned char or -1 (EOF),
otherwise it exhibits undefined behavior. Casting from char to unsigned
char avoids this.

MSVC warns about this when compiling with /analyze.

### DIFF
--- a/include/internal/catch_test_case_info.cpp
+++ b/include/internal/catch_test_case_info.cpp
@@ -37,7 +37,7 @@ namespace Catch {
             return TestCaseInfo::None;
     }
     bool isReservedTag( std::string const& tag ) {
-        return parseSpecialTag( tag ) == TestCaseInfo::None && tag.size() > 0 && !std::isalnum( tag[0] );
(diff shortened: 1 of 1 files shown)


ianhattendorf:
Would it be worthwhile to add an option when compiling with MSVC to also perform static analysis (`/analyze`)? This way it could be performed automatically by appveyor.

horenmar:
Strictly speaking, we do not support UTF-8 and thus it is the error of the user if he runs into trouble with this, but there is no reason not to guard against it. Thanks,

horenmar:
@ianhattendorf I forgot to address that before, but as far as using `/analyze` during CI is concerned,I would rather avoid it, at least for each commit, as the time spent on Windows builds is already enormous. I think it would be possible to create a new branch that has an `/analyze` build enabled and periodically merge the master into it.

If you open up a PR, I would merge it.

PR #1268, opened 2018-04-30
Suggestion for adding libcluon that is also using Catch2 for testing

chrberger:
## Description
Suggestion for adding libcluon as we are also using Catch2 for testing.


horenmar:
Thanks.

PR #1272, opened 2018-05-02
Enable console colour in the approval tests on windows.

rupertsteel:
This method to enable colours only works on windows 10.

### DIFF
--- a/scripts/approvalTests.py
+++ b/scripts/approvalTests.py
@@ -12,6 +12,10 @@ import difflib
 import scriptCommon
 from scriptCommon import catchPath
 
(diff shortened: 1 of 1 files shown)


horenmar:
> This method to enable colours only works on windows 10.

And what happens on other versions of Windows?

rupertsteel:
I can't test as I don't have a non windows 10 machine, but I expect the behaviour to be as before which was outputting the ANSI escape code unmodified to the terminal.

horenmar:
I got someone with an old VM to test it, and the behaviour is the same as before. This means that it is ok to merge.

PR #1276, opened 2018-05-05
Fixed Catch2 for GCC 8

cjdb:
## Description

While trying to install Catch2 on a Docker system with GCC 8, I encountered this warning:

```
[32/100] Building CXX object CMakeFiles/SelfTest.dir/projects/SelfTest/UsageTests/Misc.tests.cpp.o
FAILED: CMakeFiles/SelfTest.dir/projects/SelfTest/UsageTests/Misc.tests.cpp.o
/usr/local/bin/c++   -I/usr/src/catch/Catch2/include -O3 -DNDEBUG   -Wall -Wextra -Wunreachable-code -Wpedantic -Werror -std=c++11 -MD -MT CMakeFiles/SelfTest.dir/projects/SelfTest/UsageTests/Misc.tests.cpp.o -MF CMakeFiles/SelfTest.dir/projects/SelfTest/UsageTests/Misc.tests.cpp.o.d -o CMakeFiles/SelfTest.dir/projects/SelfTest/UsageTests/Misc.tests.cpp.o -c /usr/src/catch/Catch2/projects/SelfTest/UsageTests/Misc.tests.cpp
/usr/src/catch/Catch2/projects/SelfTest/UsageTests/Misc.tests.cpp: In function 'void {anonymous}::MiscTests::____C_A_T_C_H____T_E_S_T____77()':
/usr/src/catch/Catch2/projects/SelfTest/UsageTests/Misc.tests.cpp:307:65: error: type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]
         auto const s = const_cast<wchar_t* const>( L"wide load" );
                                                                 ^
```

Essentially, GCC 8 is warning that the `const` in the `const_cast` is ignored. This appears to be the only break for GCC 8, but it wasn't testable outside of my machine, so I added GCC 8 to Travis too.

## GitHub Issues

No GitHub issue right now, but I can add one.

## Other

I am running a clang-6.0 branch through Travis, and will make a PR for that too, if it passes.

PR #1277, opened 2018-05-05
Updated Travis for LLVM 6.0

cjdb:
## Description

While adding updating Catch2 for GCC 8 (see #1276), I noticed that Clang 6 is missing.
This patch will add Clang 6 to Travis.

It is separate to #1276, since #1276 also targets a warning surfaced in GCC 8.


horenmar:
Thanks.

PR #1283, opened 2018-05-08
Add gdbinit and lldbinit files to skip stepping into Catch code during debugging

### LINKED ISSUE #904: lldb: script to allow "step-in" to skip the Catch frames
Has anyone developed any lldb (or gdb) scripts that set it up to allow lldb to skip stepping into Catch code?

For example, if I have this:
   CHECK(foo());
and I am debugging, I want to step-into the foo function, but right now, it steps into all the CHECK infrastructure and it's annoying to try to get into foo directly. 

Debuggers have ways to inform them that certain frames are not interesting and should be skipped when doing debugging. 

Has anyone developed such scripts for Catch? I can maybe take a stab, but I figure someone may have already done it, and I think it'd be a great addition to the Catch repo.

mrpilot2:
## Description

The commands provided have to be executed in the current gdb/lldb session or copied
into the users ~/.gdbinit ~/.lldbinit files to permanently skip debugging Catch code.

## GitHub Issues
Fixes #904

### DIFF
--- a/contrib/gdbinit
+++ b/contrib/gdbinit
@@ -1,0 +1,16 @@
+#
+# This file provides a way to skip stepping into Catch code when debugging with gdb.
+#
+# With the gdb "skip" command you can tell gdb to skip files or functions during debugging.
+# see https://xaizek.github.io/2016-05-26/skipping-standard-library-in-gdb/ for an example
+#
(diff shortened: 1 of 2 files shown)


horenmar:
Nice, thanks.

Issue #443, opened 2015-06-18
Support C++11 pseudo-random generators


prapin:
CATCH has a command line option `--rng-seed` to initialize pseudo-random generator before running unit tests, either with a fixed value or based on time.

It currently only calls `srand` with the specified value, to prepare next calls to `rand`.

The problem is that `srand` and `rand` are _legacy_ C functions inherited by C++98.
In C++11, a whole new and quite complete API has been developed for pseudo-random generators (http://en.cppreference.com/w/cpp/numeric/random)

My proposal is to add in CATCH a global variable `Catch::random`, of type `Catch::random_t` that can be defined as for example as `std::default_random_engine`. That variable is initialized before each test (see issue #442) with the same seed as `srand`. Then, unit tests can use that variable to generate random numbers with `std::uniform_int_distribution` or any similar function.

Of course, to continue supporting old C++98 compilers, that feature has to be in a `#if` compilation switch block.

f-koehler:
I support this idea, would be great!

philsquared:
I've taken a first step in this direction.
First, `srand()` is now called (with the same seed) before every test case is invoked.
Secondly, the value of the seed is now globally available (from test case) by calling `Catch::rngSeed()` - so you can use it to see your own generator deterministically and repeatably.
I do plan to provide an alternate to the current use of `srand()`, `rand()` and `std::random_shuffle()` for C++14+ compilers - but I believe what I've done already should meet your immediate needs?
It's available in the `v1..2.1-develop.2` build.

Morwenn:
@philsquared I like that `Catch::rngSeed()` exists, but is there a way to know whether `--rng-seed` has been passed to the command-line, like `Catch::rngWasSeeded()`? When running my test suite in CI I want to seed it with `--rng-seed`, but otherwise I don't care that much and might automatize the generation from within the test suite itself with a poor man's seed.

philsquared:
@Morwenn - a bit late, I know - but `Catch::rngSeed()` returns 0 if it has not been set.

Morwenn:
Oh, fair enough, thanks :)

PR #1289, opened 2018-05-12
Now using <random> features instead of rand/srand

naslundx:
## Description
A modern C++11 framework shouldn't use rand/srand but rather the utilities in the `<random>` library.

Turns the RandomNumberGenerator into a wrapper over `default_random_generator`. The seed defaults to what's specified in the configuration, or can be manually overridden.

I do not quite follow why `seedRng` was called in so many places. Is Catch also meant to be setting a random seed for `rand()` calls inside test cases? If there is a good explanation I can reinstate this, but Catch itself does not use `std::srand` anymore.

## GitHub Issues
Should take care of #443

### DIFF
--- a/docs/command-line.md
+++ b/docs/command-line.md
@@ -244,7 +244,7 @@ Randomly sorted. Test names are sorted using ```std::random_shuffle()```. By def
 ## Specify a seed for the Random Number Generator
 <pre>--rng-seed &lt;'time'|number&gt;</pre>
 
-Sets a seed for the random number generator using ```std::srand()```. 
+Sets a seed for the random number generator. 
 If a number is provided this is used directly as the seed so the random pattern is repeatable.
 Alternatively if the keyword ```time``` is provided then the result of calling ```std::time(0)``` is used and so the pattern becomes unpredictable.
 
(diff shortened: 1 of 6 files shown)


philsquared:
Sorry it's taken a while to get to this. I'm looking into modernising the RNG right now (tbh I thought we'd done this a while back, so was surprised we still had the `std::rand` calls in).

I've had a look at your PR and might use it as reference, but one immediate problem is the removal of `seedRng()`!

Users have the ability to seed the RNG from the command line, and that should make the sequence repeatable. Imagine you're run tests on a CI server using `time` seeded (pseudo)randomness - and you get a failure only on a particular run. When using a `time` seed Catch2 prints the seed it uses, so you can then explicitly seed a local run with that number to reproduce the same conditions.

So that call is essential.

philsquared:
In fact I've gone a bit of a different path - no need for the `RandomNumberGenerator` class at all, now - just needed a global, seeded, instance of an mt19937.

See: commit 35a57b070fd4db56759dc3ecb8156f40841111a3

naslundx:
So close this?

JoeyGrajciar:
Based on previous comment closing this

PR #1293, opened 2018-05-20
Add PARSE_CATCH_TEST_PROXY variable.

conradjones:
This causes catch tests to be executed through a proxy.

An example of this would be used like this to execute tests on a developers PC and CI device consistently :

SET(SERVER_ARGS "/usr/bin/xvfb-run" "--server-args=-screen 0 800x600x16" "-e" "/dev/stdout" )
set(PARSE_CATCH_TESTS_PROXY ${SERVER_ARGS})


conradjones:
not convinced by those stats tbh bot. I think the travis ci is broken, looks like it's having difficulty installing some ubuntu packages,

PR #1294, opened 2018-05-21
Add float/int literal for Approx

henryiii:
## Description

This adds access to Approx using C++11's custom literals. *(WHAT)* Since approximate values can be often written directly in the source code, this is a handy shortcut that reduces paratheses when checking lots of values. *(WHY)*

Example:

```cpp
// Before
CHECK(myval == Approx(1.5));
CHECK(otherval == Approx(1.5).epsilon(.1));

// After
using namespace Catch::literals;
CHECK(myval == 1.5_a);
CHECK(otherval == 1.5_a .epsilon(.1));
```

This also required a unary `-` operator for Approx due to the way literals are constructed.

## GitHub Issues

None, small new feature.

Notes:
* Is `_a` best? `_unc` is a bit long.
* Due to C++11 [syntax](http://en.cppreference.com/w/cpp/language/user_literal), a space is required between a number literal and a following dot.
* Probably should have ~~a test and~~ a note in the documentation. (Test added)

### DIFF
--- a/include/internal/catch_approx.cpp
+++ b/include/internal/catch_approx.cpp
@@ -35,6 +35,13 @@ namespace Detail {
         return Approx( 0 );
     }
 
+    Approx Approx::operator-() const {
+        auto temp(*this);
+        temp.m_value = -temp.m_value;
+        return temp;
+    }
+
+
     std::string Approx::toString() const {
         ReusableStringStream rss;
         rss << "Approx( " << ::Catch::Detail::stringify( m_value ) << " )";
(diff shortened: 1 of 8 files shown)


henryiii:
(Testing locally, will repush)

horenmar:
I am not 100% on the `_a` suffix, but I am going to merge it in, at least temporarily.

horenmar:
I ended up having to cherry-pick the changes in 283e2e6d41b181fd1c35cecd94c7b2f8e853d888, so I am gonna close this manually.

PR #1306, opened 2018-06-05
Clang templated stream op

### LINKED ISSUE #1285: Clang compilation fails when using REQUIRE with custom string classes that overload the << operator
## Description

We have our own custom string class that overloads the << operator, and it's causing issues when using the REQUIRE macro and compiling under Clang in Linux.

The REQUIRE line looks like this:

REQUIRE(str1 == "Item1");

The error we get is the following:

19>project_path\catch.hpp(810,21): error : use of overloaded operator '<<' is ambiguous (with operand types 'Catch::ReusableStringStream' and 'const MyCustomStringClass<char>')
19>                rss << value;
19>                ~~~ ^  ~~~~~
19>project_path\catch.hpp(829,114): note: in instantiation of function template specialization 'Catch::StringMaker<type, void>::convert<MyCustomStringClass<char> >' requested here
19>            return ::Catch::StringMaker<typename std::remove_cv<typename std::remove_reference<T>::type>::type>::convert(e);
19>                                                                                                                 ^
19>project_path\catch.hpp(1325,42): note: in instantiation of function template specialization 'Catch::Detail::stringify<MyCustomStringClass<char> >' requested here
19>                    ( os, Catch::Detail::stringify( m_lhs ), m_op, Catch::Detail::stringify( m_rhs ) );
19>                                         ^
19>project_path\MyCppFile.cpp(243,75): note: in instantiation of member function 'Catch::BinaryExpr<const MyCustomStringClass<char> &, char const &[6]>::streamReconstructedExpression' requested here
19>                    REQUIRE(str1 == "Item1");

### Extra information
  Disabling stringification with CATCH_CONFIG_DISABLE_STRINGIFICATION does not help solve this issue.

   Commenting out the following line in catch.hpp fixes the compilation error:
   This code starts in line 801:

    template <typename T, typename = void>
    struct StringMaker {
        template <typename Fake = T>
        static
        typename std::enable_if<::Catch::Detail::IsStreamInsertable<Fake>::value, std::string>::type
            convert(const Fake& value) {
                ReusableStringStream rss;

                // this line causes the problem
                // rss << value;

                return rss.str();
        }

* Catch version: **v2.2.2**
* Operating System: **Linux**
* Compiler+version: **Clang**

JulienNitard:
## Description
What: use function syntax to call operator<< in StringMaker...
Why: ...so that it is not ambiguous with user-defined templated operator<< under clang.
Currently, the code below is ambiguous between the user defined operator<< and operator<<(StringMaker&, T) because both are templates and need one substitution.

```
struct myType {/* ... */};

template <class stream>
stream& operator<<(stream& out, const myType& myT) {
  out << myT.string_value();
  return out;
}
```

## GitHub Issues
Closes  #1285

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -105,7 +105,9 @@ namespace Catch {
         typename std::enable_if<::Catch::Detail::IsStreamInsertable<Fake>::value, std::string>::type
             convert(const Fake& value) {
                 ReusableStringStream rss;
-                rss << value;
+                // NB: call using the function-like syntax to avoid ambiguity with
+                // user-defined templated operator<< under clang.
+                rss.operator<<(value);
                 return rss.str();
         }
 
--- a/projects/SelfTest/Baselines/compact.sw.approved.txt
+++ b/projects/SelfTest/Baselines/compact.sw.approved.txt
@@ -1065,6 +1065,9 @@ ToStringWhich.tests.cpp:<line number>: passed: ::Catch::Detail::stringify(item) 
 ToStringWhich.tests.cpp:<line number>: passed: ::Catch::Detail::stringify( item ) == "operator<<( has_operator )" for: "operator<<( has_operator )"
 ==
 "operator<<( has_operator )"
+ToStringWhich.tests.cpp:<line number>: passed: ::Catch::Detail::stringify( item ) == "operator<<( has_template_operator )" for: "operator<<( has_template_operator )"
+==
+"operator<<( has_template_operator )"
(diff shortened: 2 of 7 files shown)


horenmar:
Is the explicit specialization actually needed? By my reckoning, once the lhs of `operator<<` is concretized to the `ReusableStringStream` by using it as a member function, the call will become unambigous. This means that `rss.operator<<(value)` is enough.

JulienNitard:
You are correct. It is not needed. Removing.

horenmar:
Thanks for the fix, I took the liberty of squashing the changes into one atomic commit.

JulienNitard:
Thanks, perfect.

Issue #1267, opened 2018-04-29
JUnit reporter does not write all test results


greenrobot:
## Description
When we switched from Catch V2.0.1 to V2.2.2, the test count on our CI dropped to a quarter. 

### Steps to reproduce
Run `test -r junit -o test.xml` compiled with 1) V2.0.1 and 2) V2.2.2. 

* Catch version: **v2.2.2 or earlier, V2.0.1 still worked**
* Operating System: **Ubuntu 16.04**
* Compiler+version: **clang5**

ltjax:
I believe I have the same problem. Specificially, my output with -r junit in Version 2.2.2 is:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="tests" errors="0" failures="0" tests="2" hostname="tbd" time="0.000155" timestamp="2018-05-08T13:49:10Z">
    <system-out/>
    <system-err/>
  </testsuite>
</testsuites>
```
While with version 2.0.1, it is

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="tests" errors="0" failures="0" tests="2" hostname="tbd" time="0.000101" timestamp="2018-05-08T13:50:12Z">
    <testcase classname="tests.global" name="Parses empty string as empty row" time="0.000022"/>
    <testcase classname="tests.global" name="Parses a single number correctly" time="0.00003"/>
    <system-out/>
    <system-err/>
  </testsuite>
</testsuites>
```
I did not change anything but the catch.hpp and recompiled. This is using g++ 4.9 on debian 8.

Longhanks:
I, too, experience this problem. The XML, console and compact reporters are not affected.

horenmar:
Try using `-s` flag for now. The internals were changed so that successful assertions/tests/etc are not passed all the way down to reporters unless `-s` (and thus output for success) is enabled, but we didn't take into account that some reporters always need all results.

Longhanks:
Thanks, appended `-s` and the report is complete again.

merisanualex:
I faced a similar issue, successful tests didn't appear in the junit output except the ones that contained a REQUIRE_NOTHROW or REQUIRE_THROWS. So as a workaround, adding either of those macros worked.

horenmar:
Can you guys check if the current master addresses this issue? The JUnit and XML reporters should once again report all of the assertions even without `-s`.

Longhanks:
[...]

Issue #1264, opened 2018-04-25
assertionEnded is not called on reporters


berkus:
## Description

I included TAPReporter, enabled it via cmdline but report always comes out as `1..0` with nothing else. After debugging for a while I found that `assertionEnded` method of the reporter is never called, and this is crucial place for actually counting the results.

Any ideas why that might be?

### Steps to reproduce

1. #include "catch_reporters_tap.hpp"
2. `test -r tap`

### Extra information

* Catch version: **v2.2.2**
* Operating System: **macOS**
* Compiler+version: **Apple LLVM version 9.1.0 (clang-902.0.39.1)
Target: x86_64-apple-darwin17.5.0**

berkus:
I see that `void RunContext::assertionPassed() {` does not call assertionEnded() with success.

berkus:
And the fix was
```
        ConfigData data;
        data.showSuccessfulTests = true; // important for TAPReporter
```

I guess TAPReporter could set this internally in constructor, but i didn't test that yet.

PR #1310, opened 2018-06-15
Always report the topLevel testcase in JUnit reporting backend

achamayou:
## Description

This pull request changes the JUnit report to always report topLevel testcase entries in a testsuite (as opposed to the nested ones that are created to represent sections etc), regardless of wether they are empty/passing or not.

There are two reasons for this change:

1. With the latest Jenkins release, and with the Tutorial example, the JUnit plugin actually does not work and complains that the results of the tests are empty, since it is looking for at least one testcase element. With this change, it correctly parses the xml report, and displays the factorial test.
2. In general, there are two benefits to reporting passing testcases. One is that Jenkins will display their elapsed time, which is useful when trying to find out where a (passing) suite spends most of its time. The other is it's showing diffs of added and removed tests since last build. And then it's just nice to be able to navigate the table of all tests if you're new to a project.

## GitHub Issues

N/A, but happy to open one if it helps.


achamayou:
I am not sure why a few of the targets fails, and running them on the same commit in my fork, I haven't been able to reproduce the failures: https://github.com/achamayou/Catch2/pull/2

I am guessing they are transient, and will go away if the build is restarted, which seems to be something that only the repo owners can do?

achamayou:
Reconciled against latest master changes. Feedback would be lovely.

horenmar:
First, in regards to the build failure, Travis has started having transient failures lately, where the build image errors out while getting the required packages.

Anyway, if I understand the problem correctly, I think this is the wrong way to go about fixing it. Instead, the JUnit reporter should be given all of the finished assertions the way it used to before changes that caused #1264 and #1267.

I will be committing a change to address this soon, can you instead see if that works for you?

achamayou:
@horenmar thank you for taking a look. I thought the state of things was deliberate, and meant to minimise report size. I wasn't aware that it was an unintended side effect of a recent change, which is why I'd made this change to only report top level testcases on success.

If we can have full details on success instead, I'm certainly very happy with that. I will test your change and confirm that it addresses the issue (but it sounds like it will).

achamayou:
@horenmar I have tested the latest revision of master, including your change, and it does report passing tests in junit format in such a way that they are reported accurately in Jenkins. Thank you!

If it was possible to have a release including this change not too far from now, it would be very convenient. If I can help with that in any way, please let me know how.

horenmar:
v2.3.0 is out now.

achamayou:
Thank you!

PR #1320, opened 2018-06-29
fix link to single header version catch.hpp

svenevs:
Existing link goes to oblivion.  Thanks for Catch xD


horenmar:
Yeah, we knew about the problem but didn't have time to fix it until now. Thanks

PR #1333, opened 2018-07-10
Properly unset tags variable.

gegles:
## Description

Fix a bug where tags from one test would spill into the next test if this next text did not have any tags associated.

This simply unsets the "Tags" variable when the test does not have tags.


horenmar:
Thanks.

PR #1357, opened 2018-08-09
catch_matchers: Fix compilation error with -Wnon-virtual-dtor

iam:
Comping with clang and ```-Wnon-virtual-dtor``` fails as follows:

```
include/internal/catch_matchers.h:38:16: error:
  'Catch::Matchers::Impl::MatcherMethod<std::__1::basic_string<char> >'
  has virtual functions but non-virtual destructor [-Werror,-Wnon-virtual-dtor
```

This CL wraps the offending code in a pragma ignore.

### DIFF
--- a/include/internal/catch_matchers.h
+++ b/include/internal/catch_matchers.h
@@ -34,6 +34,11 @@ namespace Matchers {
             mutable std::string m_cachedToString;
         };
 
+#ifdef __clang__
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks

horenmar:
I modified your PR slightly to fit in with the rest of the code better.

PR #1358, opened 2018-08-09
add no-rtti and no-exception support to catch2

iam:
Hi, I'm trying to port catch2 so we can use it in aosp (https://source.android.com/) and also because rxcpp (https://github.com/ReactiveX/RxCpp) is using catch2 for its test suite (which I'm also in the process of downstreaming to aosp). 

All of our code is compiled with clang -fno-exceptions -fno-rtti, so do you have any interest in taking patches to add support for this?

This is my initial attempt. If there's interest I can make any changes as you see fit, I would like to avoid having an aosp-only fork.

-----------------------

(This is a work-in-progress)

try/catch/throw/typeid become illegal keywords when their compiler
support is disabled.

By quering the compiler __has__feature we can determine if those
keywords would be legal to use or not.

For RTTI, disable the offending tests and replace dynamic_cast with
virtual calls.

For exceptions, replace try/catch with macros and throw with an
std::terminate.

Limitations: Currently when a REQUIRE() call fails it immediately
calls std::terminate, causing the rest of the test run to abort.

Correctness: Running the SelfTest main by itself passes 100%.

However running approvalTests script (without rtti and exceptions) fails,
largely because the .txt files that are diffed against have test output
that is now disabled without exceptions.

### DIFF
--- a/include/external/clara.hpp
+++ b/include/external/clara.hpp
@@ -41,6 +41,12 @@
 #define CATCH_CLARA_TEXTFLOW_CONFIG_CONSOLE_WIDTH 80
 #endif
 
+#if CATCH_CONFIG_USE_EXCEPTIONS == 0
+// std::cerr + std::terminate instead of 'throw'
+#include <iostream>
+#include <exception>
+#endif
+
 
 namespace Catch { namespace clara { namespace TextFlow {
 
@@ -580,12 +586,11 @@ namespace detail {
 
     protected:
         void enforceOk() const override {
-            // !TBD: If no exceptions, std::terminate here or something
             switch( m_type ) {
                 case ResultBase::LogicError:
-                    throw std::logic_error( m_errorMessage );
+                    Exception::doThrow( std::logic_error( m_errorMessage ) );
                 case ResultBase::RuntimeError:
-                    throw std::runtime_error( m_errorMessage );
+                    Exception::doThrow( std::runtime_error( m_errorMessage ) );
                 case ResultBase::Ok:
                     break;
             }
@@ -667,15 +672,20 @@ namespace detail {
         NonCopyable &operator=( NonCopyable && ) = delete;
     };
 
+    struct BoundValueRefBase;
+
     struct BoundRef : NonCopyable {
         virtual ~BoundRef() = default;
         virtual auto isContainer() const -> bool { return false; }
+        virtual auto isValueRefBase() const -> bool = 0;  // Support for compiling without rtti.
     };
     struct BoundValueRefBase : BoundRef {
         virtual auto setValue( std::string const &arg ) -> ParserResult = 0;
+        virtual auto isValueRefBase() const -> bool { return true; }
     };
     struct BoundFlagRefBase : BoundRef {
         virtual auto setFlag( bool flag ) -> ParserResult = 0;
+        virtual auto isValueRefBase() const -> bool { return false; }
     };
 
     template<typename T>
@@ -907,7 +917,7 @@ namespace detail {
             if( token.type != TokenType::Argument )
                 return InternalParseResult::ok( ParseState( ParseResultType::NoMatch, remainingTokens ) );
 
-            assert( dynamic_cast<detail::BoundValueRefBase*>( m_ref.get() ) );
+            assert( m_ref.get()->isValueRefBase() );
             auto valueRef = static_cast<detail::BoundValueRefBase*>( m_ref.get() );
 
             auto result = valueRef->setValue( remainingTokens->token );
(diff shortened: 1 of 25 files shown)


horenmar:
Ok, I've thought about this and we will support `-fno-exceptions`.

I am going to look through the changes more closely later, but here are some quick notes:

* Changes to Clara need to go to the Clara repo. Also, Clara should already work with `-fno-rtti`, is that not the case?
* All of the supported compilers need to work with this patch. In practice this means that either the detection needs to work for all of them automatically, or there needs to be user-accessible compile-time toggle.
* `CHECK_CONFIG_USE_EXCEPTIONS` define comes out of nowhere.

iam:
Hi Martin,

Thanks for replying, looking forward for more feedback.

> Changes to Clara need to go to the Clara repo. 

Got it, I'll upstream the changes for Clara there. Do I leave the copy-of-clara in catch2 untouched?

> Also, Clara should already work with -fno-rtti, is that not the case?

`dynamic_cast` seems to be used by clara,  this keyword causes a compiler error with `-fno-rtti`.

> All of the supported compilers need to work with this patch. In practice this means that either the detection needs to work for all of them automatically, or there needs to be user-accessible compile-time toggle.

I see on the CI it's using clang, g++, VS2015 and VS2017. Are those the only supported compilers? I don't currently have VS2015/17 myself but is it enough to rely on presubmit to validate the vc support?

By the "user-accessible compile-time toggle" do you mean a CMake option? When I read `docs/configuration.md` it seems most configs are cpp `#define` macros, so my attempt was to be as close to that as possible, is there better way?

> `CHECK_CONFIG_USE_EXCEPTIONS` define comes out of nowhere.

Not sure I understood, can you elaborate?

horenmar:
> Do I leave the copy-of-clara in catch2 untouched?

Yes. Also don't modify anything under `single_include`, those files are updated on releases only.

> `dynamic_cast` seems to be used by clara, this keyword causes a compiler error with `-fno-rtti`.

You might want to rebase your branch on current master. At some point Clara didn't support `-fno-rtti`, but that is considered a bug for both Catch and Clara.

> I see on the CI it's using clang, g++, VS2015 and VS2017. Are those the only supported compilers? I don't currently have VS2015/17 myself but is it enough to rely on presubmit to validate the vc support?

We try to support more compilers and platforms, but only those on Travis/AppVeyor are considered as having first-class support. The reason for that is simple, without CI it is hard to ensure that e.g. QNX keeps working, so it is just fixed if someone complains we broke it.

Also, you will need to check VS 2017 build log manually, CMake doesn't let us set the projects up properly until [this](https://gitlab.kitware.com/cmake/cmake/issues/18261#note_445928) issue is fixed.

---------

As far as the configuration defines are concerned, currently\* Catch does not use valued defines for configuration (e.g. it checks if `CATCH_CONFIG_FOO` is defined or not, not whether it is `1` or `0`). Checking if `CATCH_CONFIG_USE_EXCEPTIONS` is true or false means that the user would have to define it as `-DCATCH_CONFIG_USE_EXCEPTIONS=1`, while all the other configuration options would still use the `-DCATCH_CONFIG_FOOBAR` variant.

\* I am not convinced that it is the better option, but it is not painful enough to warrant changing.


And the problem with `CHECK_CONFIG_USE_EXCEPTIONS` is that it is only used in couple of test files, and the other test files use `CATCH_CONFIG_USE_EXCEPTIONS` which suggests that it is a typo. Also it isn't set anywhere, further suggesting that it is a typo.

onqtam:
regarding the 2017 tests on appveyor: I've added an option to doctest itself to remove paths from filenames (called ```--no-path-filenames```) exactly for the purpose of testing the framework itself with reference output. Also there is the option to have some ```.vcxproj.user``` file copied by cmake next to the generated build files by cmake that has the appropriate flag set.

horenmar:
> What about the approvalTests script, how do you want to reconcile the differences when running it against a Catch2 built without rtti/without exceptions?

I gave this some thoughts and I think the answer is to not run approval tests against `-fno-exceptions` Catch. Instead, the tests for `-fno-rtti` and `-fno-exceptions` should be similar to the examples we already have -> compile a separate binary and do some basic sanity tests against that.

I already planned to add something like that, so I should be able to add the scaffolding to master tomorrow.

horenmar:
@onqtam Great idea, it worked.

@iam The ExtraTests infrastructure is now in place on master. The idea is that you compile a single binary with the no-rtti, no-exceptions configuration and then register 1-or-more CTest tests that check for the correct behaviour. In your case that should be something like "is passing tests properly reporter", "is failing tests properly reported/aborted" and maybe some others.

PR #1362, opened 2018-08-14
Add STATIC_REQUIRE for compile time assertions

### LINKED ISSUE #1356: Feature Request: STATIC_REQUIRE macro
I've just starting using Catch2, and find it very expressive.

There are some tests that can be done at compile time; for example I am writing an adapter library and need to test if two types are implicitly convertible.

I could use `REQUIRE(std::is_convertible_v<A, B>);` which fits in nicely with this test framework.
However the argument is known at compile time. I've found myself doing something like:

    static_assert(std::is_convertible_v<A, B>, "A should be convertible to B");
    SUCCEED("A and B are convertible");

Such that the static test is logged and documented, but done at compile time.

Could there be a macro be included to wrap around that?

    #define STATIC_REQUIRE(expression, message) {  \
        static_assert(expression, message);        \
        SUCCEED(message);                          \
    }

That way it's automatically statically tested, and logged. It would be nice for the expression to be in the output itself; which I guess can be emulated with:

    #define STATIC_REQUIRE(expression, message) {  \
        static_assert(expression, message);        \
        REQUIRE(expression);                       \
    }

which would actually test twice, once at compile time and once at run time; though it should be inexpensive as its likely to have been calculated at compile-time.

parnmatt:
## Description

A constant expression can be checked at compile time; if the user wishes
to check something they known can  be checked at compile time, then they
can now STATIC_REQUIRE.

By  default  this  will  use  REQUIRE  and  be  at  runtime,  such  that
other  assertions   can  be   run.  It  can   be  enabled   by  defining
CATCH_USE_STATIC_REQUIRE

This uses static_assert; as some compilers cannot handle the modern:
    static_assert(expr)
and require the older form with a message:
    static_assert(expr, desc)
the expression has been included as  the message, rather than leaving it
as an empty string, which can be a warning on some linters.

## GitHub Issues
Closes #1356

### DIFF
--- a/include/catch.hpp
+++ b/include/catch.hpp
@@ -144,6 +144,14 @@
 
 #define CATCH_ANON_TEST_CASE() INTERNAL_CATCH_TESTCASE()
 
+#ifdef CATCH_USE_STATIC_REQUIRE
+#define CATCH_STATIC_REQUIRE( expr )       static_assert( expr, #expr ); CATCH_SUCCEED( #expr )
+#define CATCH_STATIC_REQUIRE_FALSE( expr ) static_assert( !(expr), "!(" #expr ")" ); CATCH_SUCCEED( !(#expr) )
+#else // CATCH_USE_STATIC_REQUIRE
+#define CATCH_STATIC_REQUIRE( expr )       CATCH_REQUIRE( #expr )
+#define CATCH_STATIC_REQUIRE_FALSE( expr ) CATCH_REQUIRE_FALSE( #expr )
+#endif // CATCH_USE_STATIC_REQUIRE
+
 // "BDD-style" convenience wrappers
 #define CATCH_SCENARIO( ... ) CATCH_TEST_CASE( "Scenario: " __VA_ARGS__ )
 #define CATCH_SCENARIO_METHOD( className, ... ) INTERNAL_CATCH_TEST_CASE_METHOD( className, "Scenario: " __VA_ARGS__ )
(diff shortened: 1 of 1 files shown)


include/catch.hpp
```diff
@@ -144,6 +144,14 @@
 
 #define CATCH_ANON_TEST_CASE() INTERNAL_CATCH_TESTCASE()
 
+#ifdef CATCH_USE_STATIC_REQUIRE
+#define CATCH_STATIC_REQUIRE( expr )       static_assert( expr, #expr ); CATCH_SUCCEED( #expr )
```
horenmar: These macros need to be variadic to work properly when `expr` contains commas, e.g. `std::is_same<int, int>::value`, [see this](https://godbolt.org/z/V38BwI), specifically the fact that the last macro use causes compilation error, as it is parsed as a macro with two arguments, not just one.

horenmar:
There also should be some tests present, e.g. via ExtraTests in `projects` folder. You can take a look at the one test already there, or the conversation in #1358

include/catch.hpp
parnmatt: As noted in [the docs](https://github.com/catchorg/Catch2/blob/master/docs/assertions.md#expressions-with-commas), using an alias/typedef, or surrounded the `expr` with parentheses is what is required.

Now I would have put the extra parens in myself, but then the behaviour would be inconsistent with `REQUIRE` as it currently written.

You example would work with:

```cpp
STATIC_REQUIRE((std::is_same_v<int, int>));
```

which is how I am currently using it in my own work.

horenmar: The trick is that it doesn't have to be that way for single-argument macros:

```cpp
#include <type_traits>
#define MY_ASSERT2(...) static_assert(__VA_ARGS__, #__VA_ARGS__)
MY_ASSERT2(std::is_same<int, int>::value);
```
[works](https://godbolt.org/z/uf7u6p) and uses the same trick `REQUIRE` does.

PR #1363, opened 2018-08-14
Add BDD AND_GIVEN based macros

### LINKED ISSUE #1360: Feature Request: AND_GIVEN
To be consistent with the `AND_WHEN` and `AND_THEN`; 

With BDD:

there are times when you require to expand the context, if you feel a need to type *'and'* in the context of a `GIVEN`, `WHEN`, and/or `THEN`, you should use `AND`.

we don't have an `AND`, because that just doesn't work in this abstraction; thus we have `AND_WHEN` and `AND_THEN`; thus we should also have `AND_GIVEN`.

parnmatt:
## Description

It is possible to have multple  given contexts in a single BDD scenario;
if you have to type 'and' in the GIVEN description; it's very likely you
need an AND.

A generic AND  is not possible, thus a AND_GIVEN  is added to complement
the AND_WHEN and AND_THEN.

Can be used without needing to increase indent:

```cpp
SCENARIO("...") {
    GIVEN("...")
    AND_GIVEN("...") {
        WHEN("...") {
            THEN("...") {
                // ...
            }
        }
    }
}
```

would correctly output, when requested/needed:

```
    Given: ...
And given: ...
     When: ...
     Then: ...
```

The padding had to be increased by a character in the output message, to
continue to be uniform.

## GitHub Issues
Closes #1360


parnmatt:
The differences in expected output are due to the extra space for padding (to keep consistency)

horenmar:
I like this, just gonna fix up the commit first.

PR #1365, opened 2018-08-18
Allow overriding of Python interpreter

SoapGentoo:
* Calling `python` does not allow overriding
  downstream when running tests.


SoapGentoo:
@horenmar the CI seems to always time out on the same clang config

horenmar:
Yeah, Travis images are kinda flaky.

Anyway, thanks.

PR #1371, opened 2018-08-28
Tests: TABs to Spaces

ax3l:
Fix TABs and none-PEP8 spaces in approval test.
Does not yet fix overlong lines for full `flake8` compliance.


horenmar:
Thanks, but to be honest the Python scripts are very much about supporting further development. In other words, fixing things like 898d111f72f6d31c79ec5de9e40333e29cabdb06 is much more important than changing the code to conform to a style guide (even if it is one very closely tied with the language like PEP8).

Issue #1375, opened 2018-09-04
std::string_view is stringified without quotes


melak47:
## Description
`std::string_view` probably uses some fallback stringification, and is expanded without quotes (and also bypasses pretty printing of invisibles).

### Steps to reproduce
```cpp
#define CATCH_CONFIG_MAIN
#include <catch2/catch.hpp>

#include <string>
#include <string_view>

TEST_CASE("string_view == string") {
    std::string s = "foo";
    std::string_view v = s;

    CHECK(s == v);
}
```
run with `-s`:
```
PASSED:
  CHECK( s == v )
with expansion:
  "foo" == foo
```

### Extra information
Catch version: **v2.4.0**

PR #1376, opened 2018-09-04
Add StringMaker for std::string_view

melak47:
## Description
Adds StringMaker specializations for std::string_view and std::wstring_view,
so that string_views gets stringified correctly.

## GitHub Issues
#1375

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -16,6 +16,10 @@
 #include "catch_compiler_capabilities.h"
 #include "catch_stream.h"
 
+#ifdef CATCH_CONFIG_CPP17_STRING_VIEW
+#include <string_view>
+#endif
+
 #ifdef __OBJC__
 #include "catch_objc_arc.hpp"
 #endif
@@ -152,10 +156,11 @@ namespace Catch {
     struct StringMaker<std::string> {
         static std::string convert(const std::string& str);
     };
-#ifdef CATCH_CONFIG_WCHAR
+
+#ifdef CATCH_CONFIG_CPP17_STRING_VIEW
     template<>
-    struct StringMaker<std::wstring> {
-        static std::string convert(const std::wstring& wstr);
+    struct StringMaker<std::string_view> {
+        static std::string convert(std::string_view str);
     };
 #endif
 
@@ -169,6 +174,18 @@ namespace Catch {
     };
 
 #ifdef CATCH_CONFIG_WCHAR
+    template<>
+    struct StringMaker<std::wstring> {
+        static std::string convert(const std::wstring& wstr);
+    };
+
+# ifdef CATCH_CONFIG_CPP17_STRING_VIEW
+    template<>
+    struct StringMaker<std::wstring_view> {
+        static std::string convert(std::wstring_view str);
+    };
+# endif
+
     template<>
     struct StringMaker<wchar_t const *> {
         static std::string convert(wchar_t const * str);
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -127,6 +127,7 @@ Catch's selection, by defining either `CATCH_CONFIG_CPP11_TO_STRING` or
 ## C++17 toggles
 
     CATCH_CONFIG_CPP17_UNCAUGHT_EXCEPTIONS  // Use std::uncaught_exceptions instead of std::uncaught_exception
+    CATCH_CONFIG_CPP17_STRING_VIEW          // Provide StringMaker specialization for std::string_view
 
 Catch contains basic compiler/standard detection and attempts to use
 some C++17 features whenever appropriate. This automatic detection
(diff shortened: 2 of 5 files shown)


include/internal/catch_tostring.h
```diff
@@ -16,6 +16,18 @@
 #include "catch_compiler_capabilities.h"
 #include "catch_stream.h"
 
+#ifndef CATCH_CONFIG_STRING_VIEW
```
horenmar: This belongs to `catch_compiler_capabilities.h` and should work a bit differently (there are other similar macros there already, you can use them as a rough template).

include/internal/catch_tostring.h
```diff
@@ -152,11 +164,26 @@ namespace Catch {
     struct StringMaker<std::string> {
         static std::string convert(const std::string& str);
     };
+
+#ifdef CATCH_CONFIG_STRING_VIEW
```
horenmar: I would move one of the two specializations so that there is less conditionally compiled blocks of code and things are easier to read, but that is a minor nitpick.

(This also applies to the implementation in the .cpp file)

horenmar:
Thanks for the PR, I added some comments on changes that need to be done before it can be merged.

include/internal/catch_tostring.h
melak47: Work differently how?
Using `_MSCVC_LANG` is necessary to detect whether the user is compiling with e.g. /std:c++17 on msvc, and is going to be necessary at least until the next major release .
Or do you mean putting this logic into some `CATCH_INTERNAL_MSVC_CPP17_OR_GREATER` macros?

include/internal/catch_tostring.h
melak47: The `std::wstring` specialization should only be present when `CATCH_CONFIG_WCHAR` is defined,
`std::string_view` only when string_view support is on, `std::wstring_view` only when both are on - I could move blocks around so the ifdefs don't nest, but wouldn't it still take three ifdef'ed blocks?

melak47:
I'm happy to add tests, but there seem to be no tests that specifically check whether `std::string` and `const char*` are stringified the same, so I'm not entirely sure what you'd want the tests for `std::string_view` to look like.
There are some tests for stringifying containers of strings, where the quoting behavior is indirectly tested for `std::string`.

Do you think something simple like 

```cpp
TEST_CASE("String views are stringified like other strings", "[toString][approvals]") {
    std::string_view view{"abc"};
    CHECK(Catch::Detail::stringify(view) == R"("abc")");

    std::string_view arr[] { view };
    CHECK(Catch::Detail::stringify(arr) == R"({ "abc" })");
}
```

is enough? If not, could you elaborate on what you're looking for?

-----
As for the inefficient implementation, the `char*`/`const char*`/ `wchar_t*`/ `const wchar_t*` specializations are implemented the same way.

include/internal/catch_tostring.h
horenmar: Yeah, this is much better -- in our experience, the autodetection will always break down on some obscure platform, so feature toggles need to let the user override them as they see fit, which is why there is the `INTERNAL_*` and `NO_*` distinction in defines.

2 more things:
1) `CATCH_CONFIG_STRING_VIEW` should be `CATCH_CONFIG_CPP17_STRING_VIEW` (respectively `CATCH_INTERNAL_CONFIG_CPP17_STRING_VIEW`)
2) Is there an actual need for MSVC-specific C++ standard defines? (e.g. cannot it just toggle `CATCH_CPP14_OR_GREATER` and `CATCH_CPP17_OR_GREATER`?)

horenmar:
Yep, those tests are enough.

I remembered that the PR will need one more thing though, documentation. The define needs to be added to `docs/configuration.md`, specifically the "C++17 toggles part". Just add something like

```
CATCH_CONFIG_CPP17_STRING_VIEW // Provide StringMaker specialization for std::string_view
```


------------------------

> As for the inefficient implementation, the char*/const char*/ wchar_t*/ const wchar_t* specializations are implemented the same way.

Yeah, that's why it is a minor nitpick in a comment, rather than part of the review 😃  At some point I will need to rewrite the stringification implementations -- it is causing performance problems when there is a lot of output to be written.

include/internal/catch_tostring.h
melak47: `__cplusplus` is still `199711L` on MSVC, and `_MSVC_LANG` matches what standard you selected with compiler flags.
With recent versions, you can also use `/Zc:__cplusplus` to make `__cplusplus` match `_MSVC_LANG`, but this will stay disabled by default until at least the next *major* release of MSVC. See https://blogs.msdn.microsoft.com/vcblog/2018/04/09/msvc-now-correctly-reports-__cplusplus/.

horenmar:
I fixed the issue on older compilers and squashed down the PR before merge.

PR #1380, opened 2018-09-10
Add StringMaker for std::variant, std::monostate

melak47:
## Description
Adds StringMaker specializations for `std::variant` and `std::monostate`, which can be enabled with `CATCH_CONFIG_ENABLE_VARIANT_STRINGMAKER` or `CATCH_CONFIG_ENABLE_ALL_STRINGMAKERS`.

Also includes configuration macros for detecting C++17 and <variant> header presence, so it doesn't break the build for anyone on C++ < 17 who's using `CATCH_CONFIG_ENABLE_ALL_STRINGMAKERS`.

Tests are modelled after the tuple tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -260,7 +260,7 @@ matrix:
       addons:
           apt:
               sources: *all_sources
-              packages: ['clang-6.0', 'libstdc++-7-dev']
+              packages: ['clang-6.0', 'libstdc++-8-dev']
       env: COMPILER='clang++-6.0' CPP17=1
 
     - os: linux
@@ -268,7 +268,7 @@ matrix:
       addons:
           apt:
               sources: *all_sources
-              packages: ['clang-6.0', 'libstdc++-7-dev']
+              packages: ['clang-6.0', 'libstdc++-8-dev']
       env: COMPILER='clang++-6.0' CPP17=1 EXAMPLES=1 COVERAGE=1 EXTRAS=1
 
 install:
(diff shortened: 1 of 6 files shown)


melak47:
Hm, seems the tests fail to compile on clang 6 + libstdc++ because of a bug: https://bugs.llvm.org/show_bug.cgi?id=31852
Newer clang should have a fix, newer libstdc++ should have a workaround, but I'm not sure what to do now :/

melak47:
I tried changing it to libstdc++-8-dev, but that doesn't seem to make a difference.
Also, I just noticed that the output says `clang 5.0.0` despite the travis configuration asking for 6.0.

horenmar:
The output is the image's clang, later on `CXX` is set to `${COMPILER}` which is `clang++-6.0` for that image.

--------

Anyway, that sounds like it is time for more compile time configuration, specifically `_GLIBCXX_RELEASE`(https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_macros.html).

The basic scheme is to have a check like this
```c++
#include <ciso646> // Get libstdc++ and libc++ version macros
#if defined(__CLANG__) && defined(_GLIBCXX_RELEASE) && (CLANG_MAJOR < 7) && (_GLIBCXX_RELEASE < 8)
// This combination of Clang and libstdc++ can't compile std::variant
#define CATCH_INTERNAL_CONFIG_NO_VARIANT
#endif
```

I don't actually remember Clang's macros, but you get the idea 😃
The "fun" part is that including `<ciso646>` has observable side effect under MSVC, so it has to be guarded against that as well.

melak47:
Apparently `__GLIBCXX__` is useless to detect the actual libstdc++ version, since it can be the datestamp of whenever the snapshot was taken: <https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html#abi.versioning.__GLIBCXX__>

So that leaves disabling the feature for all versions of libstdc++ <= 8, and letting the user override with `CATCH_CONFIG_CPP17_VARIANT` :/

horenmar:
Yeah, `__GLIBCXX__` does not work.

I am going to do a quick fixup and then squash + merge the PR into master.

Issue #1391, opened 2018-09-22
Multiple assertions in single test result in incorrect abort test


michaelcowan:
## Description
The _abort after_ value is tested for equality against _total failed assertions_ [here](https://github.com/catchorg/Catch2/blob/master/include/internal/catch_run_context.cpp#L315) :
```c++
bool RunContext::aborting() const {
    return m_totals.assertions.failed == static_cast<std::size_t>(m_config->abortAfter());
}
```

However it is possible, under the incorrect use of `REQUIRE`, for this comparison to fail by nesting calls to `REQUIRE`.

### Steps to reproduce
I couldn't see a nice way to write a test to demonstrate this, but here is a contrived example

```c++
namespace MyTests {
    bool doSomeRequirement() {
        REQUIRE(false);        // Clearly we should not nest REQUIRE but this is still valid
        return false;
    }

    TEST_CASE("Session should abort after this test") {
        REQUIRE(doSomeRequirement());
    }

    TEST_CASE("This test should not be executed") {
        REQUIRE(false);        // This is executed even though the first test fails
    }
}

void main() {
    Catch::ConfigData config;
    config.abortAfter = 1;     // I want to abort after a single assertion fails

    Catch::Session session;
    session.useConfigData(config);
    session.run();
}
```

### Extra information
* Catch version: v2.4.0
* Operating System: Windows 10
* Compiler+version: 19.14.26433

horenmar:
Well, that is kinda embarrassing.

Anyway, I agree that nested `REQUIRE`s are perfectly reasonable, even though they can cause some bugs (there is at least one more open, but that one is mostly cosmetic).

PR #1392, opened 2018-09-22
Abort when assertions failed is greater than or equal to config value

michaelcowan:
## Description

Changes the `aborting()` comparison to also handle when `m_totals.assertions.failed` is greater than `m_config->abortAfter()`.

It is possible for this to occur under incorrect (but valid) use of assertions such as `REQUIRED`.

## GitHub Issues

Fixes issue https://github.com/catchorg/Catch2/issues/1391


horenmar:
Thanks

PR #1395, opened 2018-09-25
add enable_testing() in cmake examples

Seppone:
### What
I added `enable_testing()` to usage examples.
### Why
By that you're able to run tests with `ctest`

## GitHub Issues
-


horenmar:
[That is what `include(CTest)` does.](https://cmake.org/cmake/help/latest/module/CTest.html)

PR #1397, opened 2018-09-26
Fix Cygwin compiler error

### LINKED ISSUE #1396: std::to_string is not a member of 'std'
## Description 

It's my understanding that this is due to [cygwin issues](https://stackoverflow.com/questions/33500074/cygwin-g-5-2-to-string-is-not-a-member-of-std)
Mailing list: [here](https://cygwin.com/ml/cygwin/2015-10/msg00421.html)

I cannot modify my gcc:cygwin. 

### Steps to reproduce

```powershell
PS C:\application\common\tests> g++ -std=c++11 -Wfatal-errors -Wall -I. -c 000-CatchMain.cpp
In file included from 000-CatchMain.cpp:9:0:
catch.hpp: In function 'std::string Catch::to_string(const T&)':
catch.hpp:8418:16: error: 'to_string' is not a member of 'std'
         return std::to_string(t);
                ^
compilation terminated due to -Wfatal-errors.
```

### Extra information

```powershell
PS C:\application\common\tests> head -n 3 .\catch.hpp
/*
 *  Catch v2.4.0
 *  Generated: 2018-09-04 11:55:01.682061
```

```cmd
C:\Users\me>ver

Microsoft Windows [Version 10.0.16299.611]
```

```powershell
PS C:\application\common\tests> g++ --version
g++ (GCC) 4.9.3
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

wimo7083:
## Description

What: Made a change to restrict to_string for compilers on specific cygwin systems.
Why: std::to_string doesn't exist on some cygwin compiler systems. 

## GitHub Issues

Fixes #1396 

Used the same preprocessor test as found in the libstd for basic_string:

https://gcc.gnu.org/onlinedocs/gcc-4.8.2/libstdc++/api/a01053_source.html

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -109,7 +109,14 @@
 // Required for some versions of Cygwin to declare gettimeofday
 // see: http://stackoverflow.com/questions/36901803/gettimeofday-not-declared-in-this-scope-cygwin
 #   define _BSD_SOURCE
+// some versions of cygwin (most) do not support std::to_string. Use the libstd check. 
+// https://gcc.gnu.org/onlinedocs/gcc-4.8.2/libstdc++/api/a01053_source.html line 2812-2813
+# if !((__cplusplus >= 201103L) && defined(_GLIBCXX_USE_C99) \
+	       && !defined(_GLIBCXX_HAVE_BROKEN_VSWPRINTF))
 
+#	define CATCH_INTERNAL_CONFIG_NO_CPP11_TO_STRING
+
+# endif
 #endif // __CYGWIN__
 
 ////////////////////////////////////////////////////////////////////////////////


horenmar:
Thanks for the PR.

At a quick look, the macros used for the check seem to be internal and I am a bit wary of basing the configuration on internal macros. Are these documented somewhere to be stable?

wimo7083:
Looking at the source for both [4.](https://gcc.gnu.org/onlinedocs/gcc-4.8.2/libstdc++/api/a01053_source.html)8 and [8.2](https://gcc.gnu.org/onlinedocs/gcc-8.2.0/libstdc++/api/a00290_source.html)

The macros persist, I didn't check in between and the usage of them varies. What would you recommend if this is not suitable?

horenmar:
Eh, I'll fix it ~in post~ when somebody complains.

PR #1399, opened 2018-10-03
Fix convert from char on ARM build

mgaio:
Some platforms set the signedness of char to unsigned (eg. ARM).
Convert from char should not assume the signedness of char.

Fix build issue with -Werror,-Wtautological-unsigned-zero-compare flags.

Signed-off-by: Miguel Gaio <mgaio35@gmail.com>

## Description
This fix build issue on ARM target with -Werror,-Wtautological-unsigned-zero-compare flags.
Following compiler logs is observed

catch2/include/internal/catch_tostring.cpp:217:21: error: comparison of 0 <= unsigned expression is always true [-Werror,-Wtautological-unsigned-zero-compare]
    } else if ('\0' <= value && value < ' ') {

## GitHub Issues
N/A

### DIFF
--- a/include/internal/catch_tostring.cpp
+++ b/include/internal/catch_tostring.cpp
@@ -205,7 +205,7 @@ std::string StringMaker<bool>::convert(bool b) {
     return b ? "true" : "false";
 }
 
-std::string StringMaker<char>::convert(char value) {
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks.

Issue #1402, opened 2018-10-07
feature request: '--reporter xml' should contain 'Randomness seeded to: ...'


matepek:
There is no information about the rng-seed in case of xml reporter.

horenmar:
Added in merging #1404.

PR #1404, opened 2018-10-09
XmlReporter: add information about rng-seed

JoeyGrajciar:
Xml result of reporter will now contain value of rng-seed in case it
is not zero.
The value will be stored in element Randomness and it's attribute seed.

Relates to #1402

### DIFF
--- a/include/reporters/catch_reporter_xml.cpp
+++ b/include/reporters/catch_reporter_xml.cpp
@@ -55,6 +55,9 @@ namespace Catch {
         m_xml.startElement( "Catch" );
         if( !m_config->name().empty() )
             m_xml.writeAttribute( "name", m_config->name() );
+        if( m_config->rngSeed() != 0 )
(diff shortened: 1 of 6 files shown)


horenmar:
Thanks, I added tests and will merge it after CI passes.

matepek:
Hello,
It doesn't work with single_include.
Bests,
Mate

horenmar:
@matepek Have you regenerated the single include before trying?

matepek:
No. I didn't know that it is a generated file. (Is it?)
Why is it versioned(added to source control) then? (a bit misleading)

horenmar:
The single include version is indeed generated from the files in `include/`, where the actual development is done. The version committed to the repository is always the state as of the last release, to let a user just clone the repository and use a stable version of Catch2. This does, however, mean, that if you want to test a version of Catch that has not been released, you need to regenerate the single-include yourself.

Issue #1403, opened 2018-10-09
Fails to compile with operator<< available for equality test types in Clang, but not GCC


thecppzoo:
Given a valid overload of `operator<<` as in this code:

```c++
#include <array>
#include <iostream>

template<typename T, std::size_t L>
std::ostream &operator<<(std::ostream &out, const std::array<T, L> &a) {
    return out << "[array]";
}

#include <catch.hpp>

TEST_CASE("fails to compile", "[bummer]") {
    std::array<int, 1> a1{1}, a2{2};
    if(a1 == a2) {
        std::cout << a1; // proves the operator<< is fine for a1
    }
    REQUIRE(a1 == a2);
}
```

Fails to compile with recent versions of Clang:
```
bug.cpp:16:16: note: in instantiation of member function
      'Catch::BinaryExpr<const std::__1::array<int, 1> &, const
      std::__1::array<int, 1> &>::streamReconstructedExpression' requested here
    REQUIRE(a1 == a2);
               ^
bug.cpp:5:15: note: 'operator<<' should be declared prior to the call site
std::ostream &operator<<(std::ostream &out, const std::array<T, L> &a) {
```

It can be reproduced in the compiler explorer:
[https://gcc.godbolt.org/z/Wkwab4](https://gcc.godbolt.org/z/Wkwab4)

Originally I was using Clang 7.0.0 on Mac Os X

thecppzoo:
I suspect the bug has to do with Clang being more strict about violations of using an overload in a template before the overload has been declared; that is, inside Catch2 there is a template that uses a not-yet-declared Catch2 function overload dependent on user types; the declaration happens afterwards, GCC incorrectly accepts it, but Catch2 correctly does not.

thecppzoo:
Guys, I think you may have a very serious error, here is why:
If you do this:
```c++
#include <internal/catch_stream.h>
#include <catch.hpp>
```
The problem goes away.  That means:
1. The semantics of your code is dependent on inclusion order!
2. In particular `operator<<` for `ReusableStringStream` is affected.

I think your `using ::operator<<` in catch.hpp is a smell.  Why do you need it? it seems you are putting global `<<` into namespace `Catch`, this alters ADL (Argument Dependent Lookup of overloads, aka Köenig Lookup).  The bug I reported is related to that, since if you don't do the `using` at least in Clang everything compiles.

[...]

PR #1405, opened 2018-10-10
Fixes disagreement between `operator<<` available for SFINAE and actually usable

thecppzoo:
## Description

Originally I reported a bug, [1403](https://github.com/catchorg/Catch2/issues/1403) about how user code provides an insertion overload (an overload for `operator<<`) for an `std` type in the global namespace, however, the code fails to compile in Clang at the point Catch tries to use it, at `ReusableStringStream::operator<<`.  This happens because the SFINAE tests for insertion availability happen after `using ::operator<<`, thus the global overload is available, which activates the `Reusable`..., but when actually attempted to use them at `ReusableStringStream::operator<<` they are **not** available because:
1. there hasn't been the `using ::operator<<` at that point
2. at the namespace `Catch` there has been one overload for `operator<<(std::ostream &, SourceLineInfo const &)` that makes the global overloads not available.

This is not a strict bug because the user is providing an overload for types in `std` in the global namespace, which is not "the same namespace" as indicated in the Catch documentation.

However, adding to the `std` namespace overloads of `operator<<` are **undefined behavior**, and thus the only other choice for a user who wants to allow catch to stream-insert `std` types would be to put the overloads for `std` types into `Catch`, which then are inconvenient to use from the global namespace or the user namespaces.  It seems the code tries to use `operator<<` global overloads, which works in GCC and perhaps other compilers that have broken two-phase lookup, but does not work in Clang which does it correctly in this case.

With the fix presented, users will now be able to provide inserters in either the same namespace as their types (this does not change) or the global namespace, and Catch2 would be able to *consistently* test for insertion and use the given insertion operations.

The fix is simple, to hoist the `using ::operator<<` directive to right after introducing an overload into `Catch` namespace, which is the point at which the global overloads become non-available.

I am thankful to Github user @mpark for his help elucidating ADL.

## GitHub Issues
[1403](https://github.com/catchorg/Catch2/issues/1403)
#Closes 1403

### DIFF
--- a/include/internal/catch_common.h
+++ b/include/internal/catch_common.h
@@ -22,6 +22,10 @@
 #include <string>
 #include <cstdint>
 
+// We need a dummy global operator<< so we can bring it into Catch namespace later
+struct Catch_global_namespace_dummy {};
+std::ostream& operator<<(std::ostream&, Catch_global_namespace_dummy);
+
 namespace Catch {
 
     struct CaseSensitive { enum Choice {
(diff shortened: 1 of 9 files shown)


horenmar:
Nice work on debugging the issue. 👍 

I am going to add a test before merging it in though.

PR #1407, opened 2018-10-11
Fix build in GCC 7

brgl:
The following error is emitted by gcc 7.3:

include/internal/catch_tostring.cpp:217:21: error: comparison is always true due to limited range of data type [-Werror=type-limits]
|      } else if ('\0' <= value && value < ' ') {
|                 ~~~~~^~~~~~~~
| cc1plus: all warnings being treated as errors

We can drop the first part of the if since '\0' == 0 and it will always
be <= than value of char (which is unsigned by default).

Signed-off-by: Bartosz Golaszewski <bgolaszewski@baylibre.com>

### DIFF
--- a/include/internal/catch_tostring.cpp
+++ b/include/internal/catch_tostring.cpp
@@ -214,7 +214,7 @@ std::string StringMaker<char>::convert(char value) {
         return "'\\n'";
     } else if (value == '\t') {
         return "'\\t'";
-    } else if ('\0' <= value && value < ' ') {
+    } else if (value < ' ') {
         return ::Catch::Detail::stringify(static_cast<unsigned int>(value));
     } else {
         char chstr[] = "' '";


mgaio:
The PR #1399 fix is related to this PR.
But issue observed is not related to compiler but to ARM target.

include/internal/catch_tostring.cpp
```diff
@@ -214,7 +214,7 @@ std::string StringMaker<char>::convert(char value) {
         return "'\\n'";
     } else if (value == '\t') {
         return "'\\t'";
-    } else if ('\0' <= value && value < ' ') {
+    } else if (value < ' ') {
```
thecppzoo: This changes the semantics, whether `char` is signed is implementation defined. In some platforms their ABI may indicate `char` is unsigned, some others signed: In Visual C++ char is signed by default, in GCC unsigned...
I think the fix is to use this instead:
`if(unsigned(value) < ' ')`

horenmar:
Thanks, but I merged #1399 instead.

PR #1411, opened 2018-10-17
Catch::LeakDetector: added cleanup call to destructor

JoeyGrajciar:
simple code with provided main function which just returns 0
leaks memory due to fact that singletons are not cleaned up

running valgrind on such simple application reports that 752 bytes
are still available in 11 blocks

this commit adds destructor to Catch::LeakDetector which calls
Catch::cleanUp()

## Description

There is already existing https://github.com/catchorg/Catch2/pull/1063 which added the destructor but seems to be abandoned.

### DIFF
--- a/include/internal/catch_leak_detector.cpp
+++ b/include/internal/catch_leak_detector.cpp
@@ -6,6 +6,7 @@
  */
 
  #include "catch_leak_detector.h"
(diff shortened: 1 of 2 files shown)


horenmar:
This looks fine enough.

PR #1412, opened 2018-10-17
Add the optional variable OptionalCatchTestLauncher

delpinux:
This is related to the `contrib/ParseAndAddCatchTests.cmake` script.

This is a very small but (I think) useful situational change.

This variable is set to allow the use of the nice `ParseAndAddCatchTests` script
in the case where a launcher is needed to execute  the script.

This is introduced to allow to launch unit tests using mpi. In this case one can write for instance `set(OptionalCatchTestLauncher ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${NUMPROC})` before calling the `ParseAndAddCatchTests` function.

Note that the `OptionalCatchTestLauncher` is not a global CMake variable/option but is meant to be set to different values while registrating tests. In my case for instance, I register sequential unit tests as well a MPI unit tests.

Best regards,
Stéphane.

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -39,6 +39,11 @@
 #    PARSE_CATCH_TESTS_ADD_TO_CONFIGURE_DEPENDS (Default OFF)                                      #
 #    -- causes CMake to rerun when file with tests changes so that new tests will be discovered    #
 #                                                                                                  #
+# One can also set (locally) the optional variable OptionalCatchTestLauncher to precise the way    #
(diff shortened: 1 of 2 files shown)


horenmar:
I am not against the change, but you should also update the related documentation in `docs/cmake-integration.md`.

delpinux:
Hello @horenmar.
Thank you for pointing out that I forgot the documentation.
I made the changes. Hopefully, it is alright, but check twice my changes since english is not my mother language.

Regards.

horenmar:
The documentation is good enough 👍

PR #1413, opened 2018-10-18
Add stdio/err capturing in test header as well. 

tortoisedoc:
Potentially fixes issue 1243.
the stdout / err was not captured for the test


horenmar:
Thanks for the PR, but you will have to explain why you think this change is relevant to that issue.

tortoisedoc:
Any idea why this is failing on the checks?

Issue #1200, opened 2018-02-24
ApprovalTests fail when using out-of-source build


Quincunx271:
## Description
I tried running the tests for Catch2 after building it with CMake + Ninja. The ApprovalTests failed. The only difference I could see is that all the `filename`s were `../project/...` rather than `project/...`.

### Steps to reproduce

```
$ git clone https://github.com/catchorg/Catch2.git
$ cd Catch2 && mkdir build && cd build
$ cmake -G Ninja ..
$ ninja
$ ctest
```

This did not show up with the default `make` generator.

### Extra information

* Catch version: [7b6e49d](https://github.com/catchorg/Catch2/commit/7b6e49d7957359e5b8108b6008c733af0e60d403)
* Operating System: Ubuntu 17.10 64 bit, on VirtualBox
* Compiler+version: GNU 7.2.0, Clang 6.0.0
* Python version: 2.7.14 (through pyenv)

[Link to gist of the output](https://gist.github.com/Quincunx271/e0cc53f96eb6bacdb264ca9e2685ac7f)

PR #1414, opened 2018-10-18
Path normalization for approvalTests.py script improved | Related to …

MaciejPatro:
…#1379

## Description

This fixes 2 out of 3 problems that I've encountered with CMake + Ninja:

1. Problem described in #1379 issue regarding 
```
-    <TestCase name="# A test name that starts with a #" filename="projects/<exe-name>/UsageTests/Misc.tests.cpp" >
+    <TestCase name="# A test name that starts with a #" filename="../projects/<exe-name>/UsageTests/Misc.tests.cpp" >
```
2. For Windows environment when there was relative path available replacement of  "\\" to "/" wasn't triggered.

-------------------
On my linux environment I've noticed also a problem with escape sequence for compact.sw
```
- ToStringGeneral.tests.cpp:<line number>: passed: tab == '\t' for: '\t' == '\t'
+ ToStringGeneral.tests.cpp:<line number>: passed: tab == '/t' for: '/t' == '/t'
```

## GitHub Issues
#1200
#1379

### DIFF
--- a/projects/SelfTest/Baselines/compact.sw.approved.txt
+++ b/projects/SelfTest/Baselines/compact.sw.approved.txt
@@ -224,16 +224,16 @@ Tricky.tests.cpp:<line number>: passed: y.v == 0 for: 0 == 0
 Tricky.tests.cpp:<line number>: passed: 0 == y.v for: 0 == 0
 ToStringGeneral.tests.cpp:<line number>: passed: true with 1 message: 'i := 2'
 ToStringGeneral.tests.cpp:<line number>: passed: true with 1 message: '3'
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks for the PR.

The escaping problem is caused by over-eager path standardization, when you replace all `\` with `/`, even outside of paths. This happens to the compact reporter because it has both the path and the `\t` in single line.

MaciejPatro:
Thank you - I've missed that!

This update should fix the issue in that case - replacing takes place only for path itself.

horenmar:
Yep, looks good now. 👍

PR #1423, opened 2018-10-25
Add the static library Catch2::Main for non-standalone builds

denizevrenci:
While the current method of adding the compile definition to a single translation unit that builds up a program is viable for multiple translation units, single executable scenarios, it is also possible that each translation unit becomes its own program. In such cases, the user needs to make a dummy source such as the one in this PR and link all objects against the dummy object corresponding to it.

This PR adds a new static library that lets users to opt-in to linking in CMake to streamline using the library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,14 @@ endif()
 # provide a namespaced alias for clients to 'link' against if catch is included as a sub-project
 add_library(Catch2::Catch2 ALIAS Catch2)
 
+if (NOT NOT_SUBPROJECT)
+    add_library(Catch2Main src/catch_with_main.cpp)
+
+    target_link_libraries(Catch2Main Catch2)
+
+    add_library(Catch2::Main ALIAS Catch2Main)
+endif ()
+
 # Only perform the installation steps when Catch is not being used as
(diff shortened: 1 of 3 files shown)


JoeyGrajciar:
@denizevrenci can you please also add a documentation (hints) about the usage?

denizevrenci:
I modified the idiomatic examples to link with the new static library. And added simple documentation about how to use it.

denizevrenci:
Thanks for the review and pointing out with Conan. After a few trials and errors through CI, I am now stuck with undefined references in the exported binary. Pretty weird, considering the same binary has no issues in linkage with the unit tests. I will try more extensively in the coming week.

denizevrenci:
"Solved" the issue with Conan by removing the library `Catch2::Main` from standalone builds.

While it is possible to ship a compiled binary in a Conan package, it still requires ABI compatibility between the library and the code that uses it. I am inclined to defer this issue to a separate PR at this point.

blackliner:
If possible, please also provide a bazel target 👍

blackliner:
Can be closed --> https://github.com/catchorg/Catch2/pull/2029

Issue #850, opened 2017-03-10
Implement Property Based Testing/ Data Generators


philsquared:
This has been a long-standing promise/ point of discussion and a number of issues have been raised that amount to this, so I thought I'd write up the definite feature request to direct them all to.

There are two strands to this:

1. Data generators, or parametrised tests: i.e. you want to re-use the same core set of assertions with a range (possible a wide range) or different inputs - including the cross product of multiple possible inputs. In some cases you want to specifically control the inputs, in others generate it as an increasing range, or randomly. There's a variation where different *types* are used to parameterise. While strictly speaking that's a different feature I suspect it's all going to get tied up in them same solution, so I'll include that here too. We can always unbundle it later, if necessary.

2. Building on (1) is the idea of [Property-Based Testing.](http://hypothesis.works/articles/what-is-property-based-testing/) This is a more formally structured approach to working with generated ranges of inputs and also includes features such as *shrinking* (where failing inputs are automatically reduced to the simplest/ smallest possible failure). The tests themselves are for *properties* which are usually invariants that should always hold - although sometimes an alternate (e.g. simple or reference) implementation is compared against.

Support for generators was experimentally added in the very early days of Catch, but were never progressed. In fact quite a lot of work was done in support of them, and a few alternate implementations started. The main obstacle was interaction with the `SECTION` tracking infrastructure. That code was reworked with generators in mind a couple of years ago, and a new proof-of-concept generators implementation was successfully written against it. However by that time the goal was full property-based testing and the path there seemed tortuous in a C++03 constrained context, so the effort was deferred to such a time as we could at least rebase on C++11 (and ideally would leverage C++14 and beyond to allow for things like Ranges v3 to be supported in the generators part). C++11 rebasing was decided for Catch2 - with generators/ PBT being one of the main motivators. At time of this writing a proof-of-concept implementation of Catch2 exists which consists of a mostly rewritten core. That work has paused while the "Catch Classic" backlog is tamed but will resume again soon - with generators and PBT being one of the first big features to be worked on next.

I'll keep this issue open until the feature is ready so others can be closed in favour of it.

capsocrates:
You asked me this in #558 

> I believe I have answered this query now (even if not entirely satisfactorily) - and objection to me closing it? (@capsocrates, if you're still watching?)

If you meant that Catch2 is an answer to the query, I suppose that's satisfactory enough. :)

Quincunx271:
I found myself needing this, so I made a workaround:

[...]

PR #1425, opened 2018-10-28
Type parametrised test cases

JoeyGrajciar:
## Description

This PR adds the support, docs and some unit tests for type parametrised test cases. Currently there is support for up to 8 types for these test cases.

Introduces `TEMPLATE_TEST_CASE` and `TEMPLATE_TEST_CASE_METHOD` macros. Usage of these macros is very similar to `TEST_CASE` and `TEST_CASE_METHOD`. Difference is that name and description (even empty string) is required as types for test are specified at the end. Example:
```c++
TEMPLATE_TEST_CASE("Templated test","",int,float, std::string)
{
    std::vector<TestType> v( 5 );
    REQUIRE( v.size() == 5 );
    REQUIRE( v.capacity() >= 5 );
}

template<typename T>
struct TemplateFixture
{
    TemplateFixture() : m_a( 1 ) {}

    int m_a;
}

TEMPLATE_TEST_CASE_METHOD(TemplateFixture,
    "A TEMPLATE_TEST_CASE_METHOD based test run that fails", 
    "[.][class][template][failing]", int, float, double)
{
        REQUIRE( TemplateFixture<TestType>::m_a == 2 );
}
```

## GitHub Issues

#46 #357 #850 ... 
Implementation is based on #382


include/catch.hpp
```diff
@@ -292,6 +308,14 @@ using Catch::Detail::Approx;
 
 #define CATCH_ANON_TEST_CASE() INTERNAL_CATCH_TESTCASE_NO_REGISTRATION(INTERNAL_CATCH_UNIQUE_NAME( ____C_A_T_C_H____T_E_S_T____ ))
 
+#ifdef CATCH_PLATFORM_WINDOWS
```
horenmar: I assume that this is done to work around the non-conforming MSVC preprocessor. Because we support non-MSVC compilers on Windows, these changes should be guarded differently.

You might also want to incorporate a check for the new and fixed MSVC preprocessor:
https://blogs.msdn.microsoft.com/vcblog/2018/07/06/msvc-preprocessor-progress-towards-conformance/

JoeyGrajciar: Yes it was done exactly for that reason :) Okay will try to limit this to something like
```c++
#if defined(_MSC_VER) && (!defined(_MSVC_TRADITIONAL) || (defined(_MSVC_TRADITIONAL) && _MSVC_TRADITIONAL)
//do the hacks
#else
//do the proper way
#endif
```

myrgy:
Hi all,
is is MR is going to be merged?
because for my daily job it will be extremely useful!

horenmar:
I am going to close this in favour of #1437

PR #1433, opened 2018-11-04
ConsoleReporter: minor formatting fix

### LINKED ISSUE #360: Reporting message formatting
There is a minor inconsistency with FAILED and PASSED message formatting.

1)

```
test.cpp:19: FAILED
    CHECK(add(2,2)==5)
with expansion
    4 == 5
```

2)

```
test.cpp:19: 
PASSED:
     CHECK(add(2,2)==4)
with expansion
     4 == 4
```

It would be good if PASSED appears on the same line as filename and line number, just like the case with FAILED message formatting.

Fix suggested by commenting in this snippet:

```
 8198  if  (stats.totals.assertions.total() > 0) {
 8199         //if ( results.isOk() )      //commented by snawaz
 8200         //    stream << "\n";        //commented by snawaz
 8201         printResultType();
 8202         printOriginalExpression();
 8203         printReconstructedExpression();
 8204  }
```

JoeyGrajciar:
PASSED will now appear on the same line as filename and line number,
just like the case with FAILED message formatting

## Description

Minor formatting change to console reporter and unit tests output.

## GitHub Issues

Closes #360

### DIFF
--- a/include/reporters/catch_reporter_console.cpp
+++ b/include/reporters/catch_reporter_console.cpp
@@ -111,8 +111,6 @@ public:
     void print() const {
         printSourceInfo();
         if (stats.totals.assertions.total() > 0) {
-            if (result.isOk())
-                stream << '\n';
             printResultType();
             printOriginalExpression();
             printReconstructedExpression();
--- a/projects/SelfTest/Baselines/console.std.approved.txt
+++ b/projects/SelfTest/Baselines/console.std.approved.txt
@@ -392,8 +392,7 @@ Message.tests.cpp:<line number>: FAILED:
 explicitly with message:
   This is a failure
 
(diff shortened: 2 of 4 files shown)


horenmar:
ISTR that there was a reason for the formatting differences (even though at the time I thought it is a really bad reason).

@philsquared Thoughts?

horenmar:
Okay, I am going to merge it and wait until someone complains -- console reporter output is not intended to be machine parseable, so it isn't a real breaking change.

philsquared:
IIRC we originally had both PASSED and FAILED on the _next_line, which was my preference. But it was XCode, I believe, that would recognised FAILED after a file/ line and put a red dot next to the source line, so seemed like a good integration.

PR #1434, opened 2018-11-04
StreamingReporterBase: add Verbosity::Quiet and Verbosity::High

### LINKED ISSUE #1426: EventListener does not support verbosity other than "normal"
## Description
`TextEventListenerBase` only supports verbosity "normal" and can not be used with a Report Processor that supports other verbosities and using one that is not `-v normal`.

### Steps to reproduce
Define a report processor that accepts verbosities other than "normal" - i.e. this way:
```cpp
 static std::set<Catch::Verbosity> getSupportedVerbosities() {
    return { Catch::Verbosity::Quiet, Catch::Verbosity::Normal, Catch::Verbosity::High};
}
```
This report processor will then accept runs with `-v quiet` and `-v high` (as well as `-v normal`).

When also adding an Event Listener. It no longer works to run with `-v high` or `-v quiet`. The result
is that the message `Verbosity level not supported by this reporter` is printed.

It is not possible to define the `getSupportedVerbosities` in the event listener since the base class
that is required does not support parametrization. 

I tried this:
```cpp
struct GlobalStateGuard : Catch::TestEventListenerBase {

private:
    char* tz_env_before   = NULL;

public:
    using TestEventListenerBase::TestEventListenerBase; // inherit constructor

    // This is called before any tests are executed. Use this method to record
    // any global state that needs to be reset after each test.
    //
    // This implementation manages timezone and TZ environment variable
    //
    void testRunStarting(Catch::TestRunInfo const& testRunInfo) override {
        // remember the TZ in effect
        tz_env_before = getenv("TZ");
        // Cause side effects of setting TZ (timezone and other external variables are set).
        tzset();
    }

    void testCaseEnded( Catch::TestCaseStats const& testCaseStats ) override {
        // Reset timezone and cause side effects (setting external timezone etc) by
        // calling tzset. It is ok if tz_env_before is NULL - that is taken as UTC.
        //
        setenv("TZ", tz_env_before, 1);
        tzset();
    }

    /*  Must state it accepts all verbosities or it will fail with verbosity
     * is something other than "normal"
     */
    static std::set<Catch::Verbosity> getSupportedVerbosities() {
        return { Catch::Verbosity::Quiet, Catch::Verbosity::Normal, Catch::Verbosity::High};
    }

};
CATCH_REGISTER_LISTENER( GlobalStateGuard )
```
And the `getSupportedVerbosities` has no effect since the framework will call the default
version of that method in `StreamingReporterBase<TestEventListenerBase>`

If I patch `StreamingReporterBase<TestEventListenerBase>` and include
the static `getSupportedVerbosities` there - it works using `-v high` and `-v quiet`

### Suggestion
Either make event listeners accept all verbosities (is perhaps the best, since why would it matter?), or make the base class for event listeners templated so it is possible to specify
that it supports different verbosities.

### Extra information
* Catch Version 2.4.1

JoeyGrajciar:
adding also other verbosities to static method getSupportedVerbosities

It will allow custom reporters and listeners to be used with these
verbosities from commandline.

Closes #1426

### DIFF
--- a/include/reporters/catch_reporter_bases.hpp
+++ b/include/reporters/catch_reporter_bases.hpp
@@ -42,7 +42,7 @@ namespace Catch {
         }
 
         static std::set<Verbosity> getSupportedVerbosities() {
-            return { Verbosity::Normal };
+            return { Verbosity::Quiet, Verbosity::Normal, Verbosity::High };
         }
 
         ~StreamingReporterBase() override = default;


horenmar:
This is not the right fix for #1426.

Issue #461, opened 2015-07-08
Catch crash. Invalid read of size 8


iwasz:
Hi

   When I have two test_cases with same name and description, a crash appears. Test cases:

```
TEST_CASE ("A", "[obd parsing]")
{
}

TEST_CASE ("A", "[obd parsing]")
{
}
```

valgrind output:

```
==14080== Invalid read of size 8
==14080==    at 0x4ECB3A9: std::ostream::sentry::sentry(std::ostream&) (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19)
==14080==    by 0x4ECBA68: std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19)
==14080==    by 0x4ECBE57: std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char) (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19)
==14080==    by 0x46D725: Catch::(anonymous namespace)::PosixColourImpl::setColour(char const*) (catch.hpp:6375)
==14080==    by 0x46D51C: Catch::(anonymous namespace)::PosixColourImpl::use(Catch::Colour::Code) (catch.hpp:6353)
==14080==    by 0x46D8B8: Catch::Colour::use(Catch::Colour::Code) (catch.hpp:6409)
==14080==    by 0x46D7F9: Catch::Colour::Colour(Catch::Colour::Code) (catch.hpp:6401)
==14080==    by 0x485C61: Catch::TestRegistry::registerTest(Catch::TestCase const&) (catch.hpp:5755)
==14080==    by 0x46CE1E: Catch::(anonymous namespace)::RegistryHub::registerTest(Catch::TestCase const&) (catch.hpp:5994)
==14080==    by 0x46CB8D: Catch::AutoReg::registerTestCase(Catch::ITestCase*, char const*, Catch::NameAndDesc const&, Catch::SourceLineInfo const&) (catch.hpp:5860)
==14080==    by 0x46CA44: Catch::AutoReg::AutoReg(void (*)(), Catch::SourceLineInfo const&, Catch::NameAndDesc const&) (catch.hpp:5845)
[...]

PR #1442, opened 2018-11-16
Remove std::ios_base::Init

karyon:
One test executable in our project was consistently crashing on shutdown due to a nullptr access when destructing that `std::ios_base::Init` instance. I haven't figured out why that happens, but removing this member fixed the crash. 

This was introduced in https://github.com/catchorg/Catch2/issues/461. It should not be necessary since C++11.

I was using visual studio 15.7.6.


horenmar:
Reading through the linked issue, the problem was with g++ 4.8.4 not initializing iostreams at the right time. 

However, we still support that version, so unless it only happens in C++03(98) mode, I don't think we can delete it just yet.

PR #1443, opened 2018-11-16
Conan recipe Update + Conan package upload

### LINKED ISSUE #926: Conan recipe for Catch project
## Description
I have written Conan recipes for Catch, since version 1.4.1, you can see [here](https://github.com/uilianries/conan-catch). However, would be nice if Catch provides its own recipe in the project.

I can help with this, starting to write the recipe.

uilianries:
## Description
1 year after #943 :birthday: I updated that PR creating a new one.

This PR includes:

- Updated Conan recipe:
  - Removed version from recipe. It will be retrieved from tag name (branch)
  - Add topics to be used on search engine
  - Updated license name to SPDX format
- Moved test package dir to .conan
  - As test_package can be customized I moved to keep root path as clean as possible.
  - Using factorial example to check Conan package
  - Validate license copy
- Build Conan package on Travis CI
  - Add job number 8 using linux + python 3.7
  - Build Conan package for every push
  - Upload Conan package ONLY when matching tag name format

**What should be done to upload Conan packages after this PR?**

The python package Conan Package Tools (CPT) is a helper to build and upload Conan package when some conditions are valid, this includes:

- [X] create a Bintray account. Already done by Phil: https://bintray.com/philsquared/Catch2
- [x] set `CONAN_PASSWORD` on Travis CI settings (do not forget to set as secret). The password is the Bintray API Key, it can be collect on https://bintray.com/profile/edit, clicking on "API Key".
- [X] set `CONAN_USERNAME` to be used on Conan package. By default is configured to **catchorg**, same organization name. The result will be *Catch2/\<version\>@catchorg/stable*
- [X] set `CONAN_LOGIN_USERNAME` to be used on Bintray login. By default is  **philsquared**.
- [X] set `CONAN_UPLOAD` to configure remote address where the package might be uploaded. By default is *https://bintray.com/philsquared/Catch2*

Thus the only requirement is `CONAN_PASSWORD` in Travis settings. All other variables are configured as needed.

## GitHub Issues
closes #926 
closes #943 

Adding some Conan experts to review:
/cc @memsharded @danimtb @Croydon @SSE4

Regards!


.conan/build.py
```diff
@@ -43,7 +43,7 @@ def upload_only_when_stable(self):
     def stable_branch_pattern(self):
         """ Only upload the package the branch name is like a tag
         """
-        return os.getenv("CONAN_STABLE_BRANCH_PATTERN", r"v?\d+\.\d+\.\d+-.*")
+        return os.getenv("CONAN_STABLE_BRANCH_PATTERN", r"v?\d+\.\d+\.\d+.*")
```
horenmar: If I am reading this properly, it would still pick up e.g. `v2.5.5-RC1`, which is not really a stable branch.

Also, you can rely on the "v" prefix always being there (generally I think the regex should be as strict as possible, to avoid publishing wrong things).

uilianries: sure! updating it now

uilianries: done!

horenmar:
So, if I understand it correctly, every push to master will build and upload a package, but it will be in "Testing" channel?

.conan/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+project(test_package CXX)
+cmake_minimum_required(VERSION 3.2.0)
```
SSE4: cmake_minimum_require must go fist

conanfile.py
```diff
@@ -3,19 +3,15 @@
...
-    name = "Catch"
-    version = "2.4.2"
+    name = "Catch2"
     description = "A modern, C++-native, header-only, framework for unit-tests, TDD and BDD"
-    author = "philsquared"
-    generators = "cmake"
-    # Only needed until conan 1.5 is released
-    settings = "compiler", "arch"
-    exports_sources = "single_include/*", "CMakeLists.txt", "CMake/catch2.pc.in", "LICENSE.txt"
+    author = "Phil Nash <github@philnash.me>"
```
SSE4: homepage?

.conan/build.py
```diff
@@ -0,0 +1,96 @@
...
+    builder = ConanMultiPackager(
+        reference=settings.reference,
+        channel=settings.channel,
+        upload=settings.upload,
+        upload_only_when_stable=settings.upload_only_when_stable,
+        stable_branch_pattern=settings.stable_branch_pattern,
+        login_username=settings.login_username,
+        username=settings.username,
+        test_folder=os.path.join(".conan", "test_package"))
+    builder.add(settings={"compiler":"gcc", "compiler.version": "8",
```
SSE4: do we need to specify compiler version here, if we already have `CONAN_GCC_VERSIONS` in `.travis.tml`?

uilianries: it's no longer required I think.

.conan/test_package/CMakeLists.txt
uilianries: done. Also, I will update our template.

conanfile.py
uilianries: include with same address from url

.conan/build.py
uilianries: I agree. removed

horenmar:
I am going to squash this and merge it into master, unless someone objects.

SSE4:
@horenmar I think now it's safe to merge, as conan developer has reviewed and approved it (@danimtb)

uilianries:
I agree!

horenmar:
Ok, thanks everyone.

uilianries:
Let's wait for the next release to have Catch2 on Bintray!

Please, don't forget to configure your Conan password (Bintray API Key) on Travis settings.

Regards!

PR #1445, opened 2018-11-17
operator<< works from the global namespace

thecppzoo:
Since https://github.com/catchorg/Catch2/pull/1405 was merged and propagated to the single-include declaring a user operator<< in the global namespace makes it available to Catch2 string converters.

## Description
Many times it is not practical or legal to add overloads in the namespace of a type, for example, the namespace `std`.  However, the user may want Catch2 to be able to use their `operator<<` overloads in the global namespace.  Since the fix mentioned in pull request 1405 these overloads are available and the documentation should reflect that.


horenmar:
Thanks.

PR #1452, opened 2018-11-21
ParseAndAddCatchTests: Add PARSE_CATCH_TESTS_HIDDEN_TESTS_DISABLED option

drdanz:
This PR adds the `PARSE_CATCH_TESTS_HIDDEN_TESTS_DISABLED` option to `ParseAndAddCatchTests`.

If this option is enabled and `PARSE_CATCH_TESTS_NO_HIDDEN_TESTS` option is disabled, the test is be added, but the [`DISABLED`](https://cmake.org/cmake/help/latest/prop_test/DISABLED.html) property is set, therefore CTest shows it as `Not Run (Disabled)` instead of `Passed`.

The tests that are not executed do not count in the number of executed tests.

I think this would probably be a good option to have enabled by default, but it is set it to `OFF` instead, since this would change current behaviour, and since it the option does not exist before CMake 3.9

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -32,8 +32,6 @@
 #    -- enables debug messages                                                                     #
 #    PARSE_CATCH_TESTS_NO_HIDDEN_TESTS (Default OFF)                                               #
 #    -- excludes tests marked with [!hide], [.] or [.foo] tags                                     #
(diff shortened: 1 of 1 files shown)


horenmar:
Sorry for not getting to this sooner, and thanks for the PR.

I think that instead of having a separate option for this, the script should check whether the CMake version is modern enough (3.9+) and if it is, set the disabled property by default.

horenmar:
I merged the new code with the logic for hiding tests.

PR #1458, opened 2018-11-26
Add cppcodec in opensource users

ndusart:
Add an entry in the list of opensource user for [cppcodec](https://github.com/tplgy/cppcodec)


horenmar:
Thanks

PR #1459, opened 2018-11-26
Deprecated 'sudo: false' removed from travis config

offa:
Deprecated `sudo: false` removed from travis config.

## Description

Removes the now deprecated option `sudo: false` from the travis config file. This is related to the current Travis Linux infrastructure migration. 

Please see the [recent Travis announcement](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration) for more information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: cpp
-sudo: false
 
 branches:
   except:


horenmar:
Yeah, Travis builds sometimes just die, and need to be restarted. Does this change mean a return to the slower startup time of VM builds?

horenmar:
Thanks.

offa:
> Yeah, Travis builds sometimes just die, and need to be restarted. Does this change mean a return to the slower startup time of VM builds?
> 
> [Seems that way.](https://travis-ci.community/t/combining-the-linux-infrastructures/310/5) Oh well.

Yes, but hopefully with better boot time. According to the linked thread they are working on improving performance. Keep my fingers crossed ...

PR #1463, opened 2018-11-29
Update tutorial.md

MappingSuite:
Updating header file URL


MappingSuite:
The Url points to a 404. Updating to the URL used at some other places.

horenmar:
Having it 404 is probably the better thing that could happen, that link points to master, so until recently it actually pointed to the latest version and not to the C++98 compatible version.

If you are looking through the documentation for the old version, could you also check that other links don't point to master? Thanks.

horenmar:
I checked and there should be no more broken links, so I am going to merge it.

Issue #1455, opened 2018-11-23
INFO() does not print message when starting with newline


an-ky:
## Description
When using INFO() to record messages, strings starting with newline are not printed on console.
I didn't see anything about this mentioned in the docs, is this a bug or by design?

### Steps to reproduce
```
#include <catch.hpp>

TEST_CASE("INFO surprise")
{
  SECTION("without newline")
  {
    INFO("Message without newline");
    REQUIRE(false);
  }

  SECTION("with newline")
  {
    INFO("\nMessage with newline");
    REQUIRE(false);
  }
}
```
**Output**
```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
base_test.exe is a Catch v2.4.2 host application.
Run with -? for options

-------------------------------------------------------------------------------
INFO surprise
  without newline
-------------------------------------------------------------------------------
D:\Projekte\monalysis\modules\base\tests\scope_exit_test.cpp(15)
...............................................................................

D:\Projekte\monalysis\modules\base\tests\scope_exit_test.cpp(18): FAILED:
  REQUIRE( false )
with message:
  Message without newline

-------------------------------------------------------------------------------
INFO surprise
  with newline
-------------------------------------------------------------------------------
D:\Projekte\monalysis\modules\base\tests\scope_exit_test.cpp(21)
...............................................................................

D:\Projekte\monalysis\modules\base\tests\scope_exit_test.cpp(24): FAILED:
  REQUIRE( false )
with message:


===============================================================================
test cases: 1 | 1 failed
assertions: 2 | 2 failed
```

I'd expect that the message `Message with newline` also be printed.

### Extra information

* Catch version: **v2.4.2**
* Operating System: **Windows10 x64**
* Compiler+version: **Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27023.1 for x64**

horenmar:
That is very much a (weird) bug.

horenmar:
After further investigations, this happens only for the Console (default) reporter.

PR #1470, opened 2018-12-08
Fix the first linebreak of messages that aborts printing #1455

Danielhu229:
## Description

To handle something like this

```cpp
TEST_CASE( "INFO and WARN start with a linebreak", "[messages][.]" ) {
    INFO( "\nthis is a " << "message start with a linebreak" );
    WARN( "\nthis is a " << "warning start with a linebreak" );
}
```

## GitHub Issues

#1455

### DIFF
--- a/include/external/clara.hpp
+++ b/include/external/clara.hpp
@@ -111,7 +111,10 @@ public:
 			m_suffix = false;
 			auto width = m_column.m_width - indent();
 			m_end = m_pos;
+            if(!line().empty() && line()[m_pos] == '\n')
+                ++m_end;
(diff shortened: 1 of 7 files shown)


horenmar:
Thanks for the PR, but the fix should go through the [TextFlow](https://github.com/philsquared/textflowcpp) repository (as the error is in a code that is 3rd party to Catch).
@philsquared

Danielhu229:
All right I see that, but that's quite confusing, I didn't even notice there are 3rd party codes in catch2 until you remind me.

PR #1474, opened 2018-12-18
Remove redundant move

kraj:
fixes errors like below

/mnt/a/yoe/build/tmp/work/aarch64-yoe-linux-musl/catch2/2.5.0-r0/git/include/internal/catch_session.cpp:52:29:
error: redundant move in return statement [-Werror=redundant-move]
   52 |             return std::move(multi);
      |                    ~~~~~~~~~^~~~~~~
/mnt/a/yoe/build/tmp/work/aarch64-yoe-linux-musl/catch2/2.5.0-r0/git/include/internal/catch_session.cpp:52:29:
note: remove 'std::move' call

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/include/internal/catch_session.cpp
+++ b/include/internal/catch_session.cpp
@@ -42,14 +42,19 @@ namespace Catch {
                 return createReporter(config->getReporterName(), config);
             }
 
-            auto multi = std::unique_ptr<ListeningReporter>(new ListeningReporter);
-
+            // On older platforms, returning std::unique_ptr<ListeningReporter>
+            // when the return type is std::unique_ptr<IStreamingReporter>
+            // doesn't compile without a std::move call. However, this causes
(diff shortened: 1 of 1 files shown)


horenmar:
As you can see on Travis, it is not quite redundant enough if you are using older gcc+libstdc++ combination.

kraj:
@horenmar Yes, I see, may be we can add conditionals for checking Clang >= 4 and gcc >= 5 to enable it. Otherwise keep the std::move

horenmar:
I took a slightly different approach, can you check if it eliminates the warning on your machine?

kraj:
> I took a slightly different approach, can you check if it eliminates the warning on your machine?

The alternate approach works for me. 

Tested- by: Khem Raj <raj.khem@gmail.com>

horenmar:
Ok, great.

Issue #1471, opened 2018-12-13
Boolean "or" not possible in REQUIRE


Wittmaxi:
## Description
I can not use a boolean or (without an extra parenthesis) in REQUIRE.

### Steps to 
in any `TEST_CASE`, put a 
```c++
 REQUIRE ((true) || (false));
```
### Extra information
Solution:
put an extra parenthesis around the operation (which should not be nessecary, IMHO):
```c++
 REQUIRE (((true) || (false)));
```
Compiler message:
```console
[maximilian@localhost ZENeural]$ make test
make -C ./library/tests  #make the unit tests
make[1]: Entering directory '/home/maximilian/Dokumente/CODING/NN/ZENeural/library/tests'
g++ -lpthread -O0 --std=c++17 test_main.cpp -g -o test -pedantic -Wall -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy -Wdisabled-optimization -Wformat=2 -Winit-self -Wmissing-include-dirs -Wold-style-cast -Woverloaded-virtual -Wredundant-decls -Wshadow -Wsign-conversion -Wsign-promo -Wstrict-overflow=5 -Wswitch-default -Wundef -Werror -Wno-unused
In file included from test_main.cpp:18:0:
test_GatedNeuron.hpp: In function ‘void ____C_A_T_C_H____T_E_S_T____18()’:
test_GatedNeuron.hpp:21:29: error: no match for ‘operator||’ (operand types are ‘Catch::ExprLhs<bool>’ and ‘bool’)
             REQUIRE ((true) || (false));
                             ^
test_GatedNeuron.hpp:21:29: note: candidate: operator||(bool, bool) <built-in>
test_GatedNeuron.hpp:21:29: note:   no known conversion for argument 1 from ‘Catch::ExprLhs<bool>’ to ‘bool’
make[1]: *** [makefile:10: all] Error 1
make[1]: Leaving directory '/home/maximilian/Dokumente/CODING/NN/ZENeural/library/tests'
make: *** [makefile:32: test] Error 2
[maximilian@localhost ZENeural]$
```
* Catch version: **Catch v2.3.0  Generated: 2018-07-23 10:09:14.936841**
* Operating System: **Linux**
* Compiler+version: **Gcc 7**

philsquared:
This is by design. All supported operators add extra complexity to the (heavily templated) mechanism for decomposing expressions, so only the comparison and relational operators are supported.

[...]

PR #1479, opened 2018-12-30
enfore the suggestion that expressions should not be overly complex b…

hbina:
…y deleting operator&& and operator|| from ExprLhs

## Description

**WHAT**
Delete the `operator&&` and `operator||` from `ExprLhs`

**WHY**
According to the documentation, specifically [here](https://github.com/catchorg/Catch2/blob/master/docs/assertions.md#natural-expressions), expressions cannot include || or && to prevent them from being overly complex.

## GitHub Issues

I think this issue is quite common. This is just the latest one I found: https://github.com/catchorg/Catch2/issues/1471 .

### DIFF
--- a/include/internal/catch_decomposer.h
+++ b/include/internal/catch_decomposer.h
@@ -142,7 +142,17 @@ namespace Catch {
         auto operator <= ( RhsT const& rhs ) -> BinaryExpr<LhsT, RhsT const&> const {
             return { static_cast<bool>(m_lhs <= rhs), m_lhs, "<=", rhs };
         }
+        
+        template<typename RhsT>
+        void operator&&( RhsT const&) = delete;
 
+        void operator&&(bool) = delete;
+        
+        template<typename RhsT>
+        void operator||( RhsT const&) = delete;
+
+        void operator||(bool) = delete;
+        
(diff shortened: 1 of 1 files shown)


include/internal/catch_decomposer.h
```diff
@@ -142,7 +142,17 @@ namespace Catch {
         auto operator <= ( RhsT const& rhs ) -> BinaryExpr<LhsT, RhsT const&> const {
             return { static_cast<bool>(m_lhs <= rhs), m_lhs, "<=", rhs };
         }
+        
+        template<typename RhsT>
+        void operator&&( RhsT const&) = delete;
+
+        void operator&&(bool) = delete;
```
horenmar: Why use an explicit specialization for `bool`?

hbina: I am not entirely sure about this but if we have something like
`REQUIRE(Factorial(1) == 1  && Factorial(2)==2);`
then the second operand after `&&` would be resolved as a `bool`?
And I think this why the original code specialized it for `bool` too.
`auto operator!=(bool rhs) -> BinaryExpr<LhsT, bool> const {
            return {m_lhs != rhs, m_lhs, "!=", rhs};
        }`
Just a way to tell that chaining with `&&` is not acceptable or something...

PR #1480, opened 2018-12-30
Add nice error messages for unsupported && and ||

### LINKED ISSUE #1273: Catch2 does not provide a helpful message when logical operators are used in an assertion
## Description

Catch Classic used to provide a fairly helpful error message when `&&` or `||` was used in an assertion, as the compilation failed inside a type whose name explained the error. For Catch2, the error message looks like this 

```
test1.cpp(6): error C2676: binary '&&': 'Catch::ExprLhs<int *const &>' does not define this operator or a conversion to a type acceptable to the predefined operator
```

which is not exactly helpful.

### Extra information
This was changed during the rewrite of capture layer, likely as an oversight.

* Catch version: **v2.x.x**

BiCapitalization:
## Description

As explained in issue #1273, `operator&&` and `operator||` should give
a proper compile time error on use instead of the compiler complaining
about them not being defined. This PR adds an `always_false` type in
`catch_meta.hpp` used for implementing a nice `static_assert` for both
of the abovementioned operators.

## GitHub Issues
Closes #1273

### DIFF
--- a/include/internal/catch_decomposer.h
+++ b/include/internal/catch_decomposer.h
@@ -10,6 +10,7 @@
 
 #include "catch_tostring.h"
 #include "catch_stringref.h"
+#include "catch_meta.hpp"
 
 #include <iosfwd>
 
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks

PR #1481, opened 2018-12-31
Additional static asserts for cascading operators

hbina:
## Description

**What**
Added a static_assert for when the user is trying to perform a cascading operations like `REQUIRE(1==1==1)`

**Why**
Since Catch2's philosophy is about simple and deterministic unit testing. These complex expressions should be decomposed to smaller ones.
Currently, doing so will cause a compile error but the error message does not really describe that this is in some sense intended.

## GitHub Issues

While I cannot find an issue that have this problem, doing cascading operators will fail and the error message is not nice.

### DIFF
--- a/include/internal/catch_decomposer.h
+++ b/include/internal/catch_decomposer.h
@@ -63,6 +63,62 @@ namespace Catch {
             m_op( op ),
             m_rhs( rhs )
         {}
+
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks, I cleaned out the history a did some minor changes.

PR #1482, opened 2019-01-01
Examples: Add semicolon after INFO

JoeyGrajciar:
Add semicolon after INFO in examples


horenmar:
The change in `catch_capture.hpp` needs to wait for next major.

JoeyGrajciar:
@horenmar without those change in catch_capture.hpp warning will not be suppressed in case of `INFO`. unless user explicitly uses the new suppression macro around such usages.

horenmar:
Right, I only read the first two commits. I meant removing the semicolon from `INTERNAL_CATCH_INFO`.

JoeyGrajciar:
I also meant that :) that will correctly fix the warning

PR #1484, opened 2019-01-04
contributing.md: Add build instructions

Lotterleben:
## Description
When looking into the `contributing.md`, I would’ve found it helpful to see at a glance how I can build and run the SelfTest binary correctly. This PR adds instructions for this.
Note that the instructions in this PR are what I came up with after some trial & error; if there’s a more elegant way to build the SelfTest binary I’d be more than happy to change the text accordingly :)

## GitHub Issues
--

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,8 +1,8 @@
 <a id="top"></a>
 # Contributing to Catch
 
-So you want to contribute something to Catch? That's great! Whether it's a bug fix, a new feature, support for 
-additional compilers - or just a fix to the documentation - all contributions are very welcome and very much appreciated. 
+So you want to contribute something to Catch? That's great! Whether it's a bug fix, a new feature, support for
+additional compilers - or just a fix to the documentation - all contributions are very welcome and very much appreciated.
 Of course so are bug reports and other comments and questions.
 
 If you are contributing to the code base there are a few simple guidelines to keep in mind. This also includes notes to
@@ -16,7 +16,7 @@ Ongoing development is currently on _master_. At some point an integration branc
 
 ## Directory structure
 
-_Users_ of Catch primarily use the single header version. _Maintainers_ should work with the full source (which is still, 
+_Users_ of Catch primarily use the single header version. _Maintainers_ should work with the full source (which is still,
 primarily, in headers). This can be found in the `include` folder. There are a set of test files, currently under
 `projects/SelfTest`. The test app can be built via CMake from the `CMakeLists.txt` file in the root, or you can generate
 project files for Visual Studio, XCode, and others (instructions in the `projects` folder). If you have access to CLion,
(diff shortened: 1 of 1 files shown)


Lotterleben:
Whoops, I just noticed that my editor auto-removed existing trailing whitespace as well. I can undo this or move it into a new PR if you like.

docs/contributing.md
```diff
@@ -43,14 +43,31 @@ you should still test your changes.
...
 once you open a PR.
-* Running the SelfTest binary. There should be no unexpected failures on simple run.
-* Running Catch's approval tests. Approval tests compare current output of the SelfTest binary in various configurations against
-known good output. Catch's repository provides utility scripts `approvalTests.py` to help you with this. It needs to be passed
-the SelfTest binary compiled with your changes, like so: `$ ./scripts/approvalTests.py clang-build/SelfTest`. The output should
-be fairly self-explanatory.
+
+* Running the SelfTest binary:
+```
+$ ./cmake-build-debug/projects/SelfTest
```
JoeyGrajciar: just a note that on Windows the binary is one more folder deeper usually Debug

docs/contributing.md
```diff
@@ -43,14 +43,31 @@ you should still test your changes.
...
+```
+There should be no unexpected failures on simple run.
+
+* Running Catch's approval tests:
+```
+$ ./scripts/approvalTests.py
+```
+Approval tests compare current output of the SelfTest binary in various configurations against
+known good output. If you've configured the SelfTest binary with your changes to be located in any other directory than
```
JoeyGrajciar: for newer documentation use shorter lines with manual breaks

docs/contributing.md
Lotterleben: Oh, good to know! I added a note for that, I hope I got the path right.

horenmar:
Thanks and don't worry about the whitespace changes.

docs/contributing.md
```diff
@@ -37,20 +37,41 @@ as these are managed by the scripts!__
...
-* Running Catch's approval tests. Approval tests compare current output of the SelfTest binary in various configurations against
-known good output. Catch's repository provides utility scripts `approvalTests.py` to help you with this. It needs to be passed
-the SelfTest binary compiled with your changes, like so: `$ ./scripts/approvalTests.py clang-build/SelfTest`. The output should
-be fairly self-explanatory.
+* Compiling Catch's SelfTest project:
+```
+$ cd Catch2
+$ cmake -Bcmake-build-debug -H. -DCATCH_BUILD_TESTING=ON
+$ cd cmake-build-debug/
+$ make
```
horenmar: This will only work for Unixish systems and not on Windows. To support Windows, the last two steps can be changed to `cmake --build cmake-build-debug` instead.

Also we support approvals through CTest now, so the building and testing can be done like so
```
cmake -Bcmake-build-debug -H. -DCMAKE_BUILD_TYPE=Debug
cmake --build cmake-build-debug
cd cmake-build-debug
ctest -C Debug --output-on-failure
```

The advantage of this approach is that it can be used without change on both Unixish systems and Windows.

The disadvantage is that to get better control, you need to familiarize yourself with another tool (CTest), so it is probably worth it to keep a separate guide for running the approval tests manually.

docs/contributing.md
```diff
@@ -37,20 +37,41 @@ as these are managed by the scripts!__
...
-once you open a PR.
-* Running the SelfTest binary. There should be no unexpected failures on simple run.
-* Running Catch's approval tests. Approval tests compare current output of the SelfTest binary in various configurations against
-known good output. Catch's repository provides utility scripts `approvalTests.py` to help you with this. It needs to be passed
-the SelfTest binary compiled with your changes, like so: `$ ./scripts/approvalTests.py clang-build/SelfTest`. The output should
-be fairly self-explanatory.
+* Compiling Catch's SelfTest project:
+```
+$ cd Catch2
+$ cmake -Bcmake-build-debug -H. -DCATCH_BUILD_TESTING=ON
```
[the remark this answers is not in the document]
horenmar: Also, you don't need `-DCATCH_BUILD_TESTING=ON`, as it is the default and will remain that way.

PR #1489, opened 2019-01-07
hexEscapeChar: restore stream

ax3l:
## Description

The ostream passed as reference to `hexEscapeChar` is manipulated and its original state not restored. This is an unexpected side effect and now fixed.

Seen via [coverity](https://scan.coverity.com/dashboard) in a [downstream project](https://github.com/openPMD/openPMD-api).


horenmar:
Thanks

PR #1490, opened 2019-01-07
ClockEstimate: Correct Average on Abort

ax3l:
## Description
The clock estimator has a potential division by zero and the average is wrong on abort. Using `iteration + 1` would be correct for an average and is consistent with the non-aborted result which is also `sum / (max(i)+1)`.

Seen via [coverity](https://scan.coverity.com/dashboard) in a [downstream project](https://github.com/openPMD/openPMD-api).

### DIFF
--- a/include/internal/catch_timer.cpp
+++ b/include/internal/catch_timer.cpp
@@ -40,7 +40,7 @@ namespace Catch {
                 // is terrible and we should move on.
                 // TBD: How to signal that the measured resolution is probably wrong?
                 if (ticks > startTime + 3 * nanosecondsInSecond) {
-                    return sum / i;
+                    return sum / ( i + 1u );
                 }
             }
 


ax3l:
No, the sum is already calculated at that point, should definitely be `i+1` in all cases

JoeyGrajciar:
so you want to divide sum by 2 in case of i=1?

ax3l:
the sum starts with zero and adds for `i=0` the first `delta`. Divided by one this is still `delta`.

The last iteration is `i=iterations-1`. At the same point where you divide `sum/iterations` you could divide `sum/(i+1)`.

JoeyGrajciar:
okay in that case  i acknowledge my mistake :)

hbina:
Is it possible to rename this PR? Division by Zero is a bit misleading because imo this will never happen. The actual problem (as correctly pointed out by the author) is that the mean/average is actually inaccurate.

ax3l:
Hi @hbina, this is zero-indexed for loop, with range for i from `0` to `iterations-1`.
```C++
# main.cpp
#include <cstddef>
#include <iostream>


int main()
{
    std::size_t iterations = 5u;

    for( std::size_t i = 0; i < iterations; ++i ) {
        std::cout << i << std::endl;
    }
    return 0;
}
```
```bash
$ g++ main.cpp && ./a.out                                                                                                                                         
0
1
2
3
4
$
```

[Here is how for-loops work in C++](https://en.cppreference.com/w/cpp/language/for):
> iteration_expression: any expression, which is executed *after every iteration of the loop* and before re-evaluating condition. Typically, this is the expression that increments the loop counter

ax3l:
@horenmar just in case: this is a simple fix for a corner-case, if you might want to take a look :)

horenmar:
Yeah, the estimate is wrong in case of early-abort, and veeeeeeery potentially can cause a division by zero (if your clock resolution is worse than 3 seconds, 🤷‍♂️ ).

There is no reason not to fix it though.

PR #1491, opened 2019-01-07
RunContext: Uninit Member

ax3l:
## Description

Fix an uninitialized member in `RunContext`.

Seen via [coverity](https://scan.coverity.com/dashboard) in a [downstream project](https://github.com/openPMD/openPMD-api).PR itself, so that maintainers can decide.
-->

### DIFF
--- a/include/internal/catch_run_context.h
+++ b/include/internal/catch_run_context.h
@@ -128,7 +128,7 @@ namespace Catch {
         TestRunInfo m_runInfo;
         IMutableContext& m_context;
         TestCase const* m_activeTestCase = nullptr;
-        ITracker* m_testCaseTracker;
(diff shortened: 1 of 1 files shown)


ax3l:
@horenmar just in case: this is a little housekeeping update, quick to review

horenmar:
I am not convinced that default initializing the pointer to an invalid value actually improves the code, but I am willing to merge the change to silence the static analysis warning.

ax3l:
Thanks, yes I think the same since having an undefined UB issue if it would be used before init in the future or using a nullptr will also be UB in most cases.

Nevertheless, silencing the static analysis and for consistency with the second pointer member that was already `nullptr` initialized was ok for me as well.

PR #1492, opened 2019-01-08
fix #1394: avoid superfluous re-run when specific section is selected

### LINKED ISSUE #1394: If a specific section is specified, actually two test runs are carried out
## Description
If I like a test in a specific section to be run, I can do this with this command:
`./MyExe Test -c sa`

Surprisingly, this causes two test runs:
1. Test "Test" with section "sa"
2. Test "Test" without any sections

Is there a reason for the second empty test run? In my application where the initialisation for the test case is lengthy this behaviour is inconvenient.

### Steps to reproduce
Put some log output in the test case "Test" outside any sections.

Lotterleben:
## Description
This PR fixes #1394.

Unfortunately, I couldn’t figure out how to properly create, filter and run a test case from within a test case, so instead of adding tests for the situation described in #1394 to `SelfTest`, I created the following:

1. A test file which contains some sections that should and shouldn’t be executed, as well as a mechanism to detect whether the “outer instructions” have been called twice. You can find it at `examples/300-FilteredSection.cpp`.
2. A shell script that calls the test file in such a way that it triggers the issue. You can find it at `examples/runFilteredSection.sh`.

I hope that helps too, but I’d be super curious to learn how to do it the right way.

## GitHub Issues
#1394


Lotterleben:
(pushed a small fix to hopefully make AppVeyor happy, will be squashed on request)

horenmar:
Thanks for the fix.

As to the tests, they should be invoked via CTest (see e.g. `project/CMakeLists.txt` for registering tests within CTest). The better question is whether they need their separate binary or not -- as they are right now, they definitely don't belong into examples, but we have `projects/ExtraTests` for tests that need to be compiled separately (and thus are too expensive to perform on every CI image).


However, I don't think they actually need separate compilation, but I'll have to think about it a bit first.

PR #1494, opened 2019-01-09
Missing <type_traits> include

amallia:
## Description
Missing `<type_traits>` include in `include/internal/catch_type_traits.hpp` is causing a compile error on Clang 6 on Ubuntu.

## GitHub Issues
Issue not created


horenmar:
Thanks

Issue #1493, opened 2019-01-08
ParseAndAddCatchTests misinterprets space before testname


Mike-Devel:
## Description
My test case headers are formatted like this:

    TEST_CASE( "name", "[tag]" )

note the space between `TEST_CASE(` and `"name"`.
That leads to an output in ctest like this:

    Start       1:  :name
    1/1 Test   #1:  :name .........................................   Passed    0.01 sec

when I would expect it to output

    Start       1: name
    1/1 Test   #1: name .........................................   Passed    0.01 sec

When executing ctest from the command line, this is merely looking strange, but when using Visual studio with the open folder functionality, the test explorer doesn't even show the tests.

I think this is related to some advanced functionality, that I'm not using, so I'm not sure how exactly to fix that, but a workaround for me is the following patch:

    @@ -109,7 +109,7 @@ function(ParseFile SourceFile TestTarget)
             # Get test type and fixture if applicable
             string(REGEX MATCH "(CATCH_)?(TEST_CASE_METHOD|SCENARIO|TEST_CASE)[ \t]*\\([^,^\"]*" TestTypeAndFixture "${TestName}")
             string(REGEX MATCH "(CATCH_)?(TEST_CASE_METHOD|SCENARIO|TEST_CASE)" TestType "${TestTypeAndFixture}")
    -        string(REPLACE "${TestType}(" "" TestFixture "${TestTypeAndFixture}")
    +        string(REGEX REPLACE "${TestType}\\([ \t]*" "" TestFixture "${TestTypeAndFixture}")

             # Get string parts of test definition
             string(REGEX MATCHALL "\"+([^\\^\"]|\\\\\")+\"+" TestStrings "${TestName}")

### Steps to reproduce
main.cpp:

    #define CATCH_CONFIG_MAIN
    #include "catch.hpp"

    TEST_CASE( "Foo","Bar" )
    {
        CHECK(true);
    }

CMakeLists.txt:

    cmake_minimum_required(VERSION 3.5)
    project( Foo )

    enable_testing()

    add_executable( Foo main.cpp)

    find_package(Catch2  REQUIRED)
    target_link_libraries( Foo PRIVATE Catch2::Catch2 )
    include(ParseAndAddCatchTests)

    set(PARSE_CATCH_TESTS_ADD_TARGET_IN_TEST_NAME Off)

    ParseAndAddCatchTests( Foo )

on console: 

    cmake .
    cmake --build .
    ctest .

### Extra information

[...]

PR #1495, opened 2019-01-10
fix ctest output with space before name

NeroBurner:
## Description
When PARSE_CATCH_TESTS_ADD_TARGET_IN_TEST_NAME is enabled the cmake helper script fails to extract the testcase name if a whitespace is before the name string. Use regex to consider and remove this whitespace.

## GitHub Issues
https://github.com/catchorg/Catch2/issues/1493


horenmar:
Thanks

Issue #760, opened 2016-12-16
Compare floating point containers


miqelm:
Will there be nice way to use Approx method for comparing floating point values stored in containers, e.g. std::vector, as it is for integral types?

For integral types I can write:
```
REQUIRE(computed == expected);
```
But for double value container I have to write loop for use Approx:
```
REQUIRE(computed.size() == expected.size());
for (size_t i = 0; i < computed.size(); ++i)
    REQUIRE(computed[i] == Approx(expected[i]));
```

or make ugly looking comparison with std::equal:
```
REQUIRE(computed.size() == expected.size());
REQUIRE(std::equal(computed.begin(),computed.end(),expected.‌​begin(), [](double x, double y)->bool { return x==Approx(y); }));
```

The best it would be to do:
```
REQUIRE(computed == Approx(expected));
```

bprichar:
I would like to second this request.

Any thoughts on how to implement it?

lightmare:
I once used something like this; not with Approx, but known expected values of any custom class work the same:
```c++
auto expected = { Approx(1.), 2., 3., 4., ... }; // initializer_list<Approx>
CHECK(computed == std::vector<Approx>(expected));
```

I also wanted to avoid the construction of the temporary vector, so I wrote a function template comparing two sequences:
```c++
template <typename C1, typename C2>
size_t count_differences(C1& c1, C2& c2);

CHECK(0 == count_differences(computed, expected));
```

bprichar:
For the time being I've resorted to a kind of workaround. I created the following two functions to be used wherever you want to check two vectors. (i.e. `REQUIRE(compareVectors(vec1, vec2));` )

```c++
bool compareVectors(std::vector<double> a, std::vector<double> b) {
    if (a.size() != b.size()) return false;
    for (size_t i = 0; i < a.size(); i++) {
        if (a[i] != Approx(b[i])) {
            std::cout << a[i] << " Should == " << b[i] << std::endl;
            return false;
        }
    }
    return true;
}

bool compare2dVectors(std::vector<std::vector<double>> a,
                      std::vector<std::vector<double>> b) {
    if (a.size() != b.size()) return false;
    for (size_t i = 0; i < a.size(); i++) {
        if (! compareVectors(a[i], b[i])) return false;
    }
    return true;
}
```
That way you'll at least be able to see the name of the vector that failed, and the first values that differed.

[...]

PR #1499, opened 2019-01-14
Vector Approx matcher

williamjameshandley:
## Description

In addition to Equals and UnorderedEquals, this PR implements an Approx matcher for std::vector<T>.

It is implemented along the lines of Equals and UnorderedEquals, with inspiration from previous attempts in issue #760, and [this StackOverflow discussion](https://stackoverflow.com/questions/41160846/test-floating-point-stdvector-with-c-catch)

## GitHub Issues

Addresses #760

### DIFF
--- a/include/internal/catch_matchers_vector.h
+++ b/include/internal/catch_matchers_vector.h
@@ -9,6 +9,7 @@
 #define TWOBLUECUBES_CATCH_MATCHERS_VECTOR_H_INCLUDED
 
 #include "catch_matchers.h"
+#include "catch_approx.h"
 
 #include <algorithm>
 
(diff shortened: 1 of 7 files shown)


horenmar:
Thanks, this looks good.

I have one note and one nitpick though.

Note: You should also add a test against strong typedef.
Nitpick: I think " is approx" would work much better as the stringification string inside the matcher.

horenmar:
If you go to `UsageTests/Approx.tests.cpp`, on line 192 there is a test that tests Approx against a strong-typedef, that is a class that can be explicitly converted to `double`, but is not a `double` itself. It should work because of the underlying `Approx` usage, but it would be nice to have it documented with a test.

------
Thinking about this further, I think this actually opens up a philosophical question about the constructor of the Matcher. Specifically, whether the constructor shouldn't be templated and allow for vector filled with types that are convertible to double, but are not `T`s themselves -- this would allow for e.g.

```cpp
std::vector<Temperature> temps;
// fill temps somehow
REQUIRE_THAT(temps, Catch::Approx({17.2, 17.3, 18.2}));
```

Which is nicer that having to convert the doubles to temperatures by yourself.


------


As to your other question, I am afraid that is unfixable without a different name, because of the `using Catch::Detail::Approx` in `catch.hpp` that cannot be removed without breaking backwards compatibility. 
You can also stop it from being ambigous by bringing in the function explicitly:
```cpp
using Catch::Matchers::Approx;
CHECK_THAT(v3, Approx(v3));
CHECK_THAT(v4, Approx(v4));
```

works.

williamjameshandley:
> You can also stop it from being ambigous by bringing in the function explicitly:

I've opted for this solution.

Regarding the strong typedefs, I had a go at this, but the main issue I seem to be getting if I try to implement a new `std::initialiser_list` call like so:

```c++
    template<typename T>                                                                                                                                                                          
    Vector::ApproxMatcher<T> Approx( std::initializer_list<T> const& comparator ) {                                                                                                               
        return Vector::ApproxMatcher<T>( comparator );                                                                                                                                            
    } 
```
is that this tends to cause segfaults due to the constant reference implicit in the `std::vector<T> const& m_comparator;` member function.

This also seems like a more wholesale change, given that it doesn't work for the `Equals`, `Contains` etc methods. Perhaps I could submit an issue to upgrade the whole `include/internal/catch_matchers_vector.h` header?

horenmar:
I modified the tests a bit and merged it. Thanks

Issue #674, opened 2016-06-13
warning generated using Catch in g++-4.7: suggest parentheses around comparison in operand of '=='


jcrada:
Hi,

The following warning is generated in my project using Catch in GNU g++-4.7:

```
/build/fuzzylite/test/BenchmarkTest.cpp: In function 'void fl::____C_A_T_C_H____T_E_S_T____32()':
/build/fuzzylite/test/BenchmarkTest.cpp:85:13: error: suggest parentheses around comparison in operand of '==' [-Werror=parentheses]
```

It would be great to be able to compile the tests without generating such a warning.

Project: https://github.com/fuzzylite/fuzzylite/tree/master
Commit: https://github.com/fuzzylite/fuzzylite/tree/d095efce076549b435ea5f9e21addf8a7346965c
Travis failing job: https://travis-ci.org/fuzzylite/fuzzylite/jobs/137058960

Thanks for such a great testing framework!

czipperz:
https://github.com/fuzzylite/fuzzylite/blob/d095efce076549b435ea5f9e21addf8a7346965c/fuzzylite/test/BenchmarkTest.cpp#L85

That `== true` should be deleted, it is invalid if I give you any truthy value instead of `true`.

jcrada:
Thank you very much for your feedback. Unfortunately, there are other cases where I am being required to enclose in parentheses different kinds of statements. For example,

https://github.com/fuzzylite/fuzzylite/blob/d095efce076549b435ea5f9e21addf8a7346965c/fuzzylite/test/activation/ThresholdTest.cpp#L31 

https://github.com/fuzzylite/fuzzylite/blob/d095efce076549b435ea5f9e21addf8a7346965c/fuzzylite/test/activation/ThresholdTest.cpp#L36

and so on, thereby requiring checks in the form `CHECK(( a == b ))` to use double parentheses in order to avoid triggering a warning. It seems a bit silly to have to use double parenthesis in such cases to avoid a warning. 

Would it not be possible to do such an enclosing in Catch instead? Hence, addressing this issue in g++-4.7 and possibly earlier versions of the compiler?

Thanks again!

czipperz:
I'll try rolling back g++ on my arch box and looking at that.  I'm in no means a Catch developer, I'm just trying to help if I can.

jcrada:
Thanks for your help!

[...]

PR #1508, opened 2019-01-24
Update catch_suppress_warnings.h

YarikTH:
fixed "pragma GCC diagnostic push" after "pragma GCC diagnostic ignored"

## Description
I found unusual gcc warnings pattern in catch_suppress_warnings.h that obviosly a bug added by someone who doesn't familar with diagnostic push-ignored-pop pattern.
```
#    pragma GCC diagnostic ignored "-Wparentheses"
#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wunused-variable"
#    pragma GCC diagnostic ignored "-Wpadded"
```
So I just moved "push" before first "ignored"

### DIFF
--- a/include/internal/catch_suppress_warnings.h
+++ b/include/internal/catch_suppress_warnings.h
@@ -19,8 +19,8 @@
      // GCC likes to warn on REQUIREs, and we cannot suppress them
      // locally because g++'s support for _Pragma is lacking in older,
      // still supported, versions
-#    pragma GCC diagnostic ignored "-Wparentheses"
 #    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wparentheses"
 #    pragma GCC diagnostic ignored "-Wunused-variable"
 #    pragma GCC diagnostic ignored "-Wpadded"
 #endif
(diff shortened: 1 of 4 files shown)


YarikTH:
Locally suppress hidden warnings on Class and Compile tests that appeared on gcc-4.8 gcc-4.9  and gcc-5.0

YarikTH:
I found #674 and understand comment above suppression
```
// GCC likes to warn on REQUIREs, and we cannot suppress them
// locally because g++'s support for _Pragma is lacking in older,
// still supported, versions
#    pragma GCC diagnostic ignored "-Wparentheses"
#    pragma GCC diagnostic push
```
So I close request. But it would be better to mark this more obviously. Because now it's unclear from comment before. e.g.:
```
// GCC likes to warn on REQUIREs, and we cannot suppress them
// locally because g++'s support for _Pragma is lacking in older,
// still supported, versions
#    pragma GCC diagnostic ignored "-Wparentheses" //< Don't move under push/pop. See #674
#    pragma GCC diagnostic push
```
or
```
// GCC likes to warn on REQUIREs, and we cannot suppress them
// locally because g++'s support for _Pragma is lacking in older,
// still supported, versions
// [
#    pragma GCC diagnostic ignored "-Wparentheses"
// ]
#    pragma GCC diagnostic push
```

PR #1509, opened 2019-01-24
Std deviation of clock resolution

hbina:
## Description

**What**
This commit attempts to find the standard deviation in getting the clock resolution of catch timer.

**Why**
To find the resolution of the clock, catch_timer will try to average a nanosecond in a system by looping 1000000 times. It is possible that during this period, the machine is disrupted by something else that will pollute the dataset. Performing a standard deviation calculation will allow us to see how much deviation have occurred in the dataset. From there we can choose to try again or accept the deviation.

## GitHub Issues

This commit does not reference to any issues afaik, because Benchmark is a WIP.


hbina:
There are still lots of problem with this PR. The problem comes from MSVC for some reason...will try to resolve this first before going ahead.

hbina:
Simple conversion fixed it.

include/internal/catch_timer.cpp
```diff
@@ -7,67 +7,94 @@
...
-                sum += delta;
+                    // We're just taking the mean, here. To do better we could take the std. dev and exclude outliers
+                    // - and potentially do more iterations if there's a high variance.
+                    result = sum / iterations;
+                    uint64_t squared_difference_sum = 0;
+                    for (std::size_t iterator = 0; iterator < iterations; ++iterator) {
+                        squared_difference_sum += (deltas[iterator] - result) * (deltas[iterator] - result);
+                    }
+                    long double std_deviation = static_cast<long double>(squared_difference_sum) / iterations;
+                    if (std_deviation < STD_DEVIATION_THRESHOLD) {
```
horenmar: This doesn't look right -- you do this check after every iteration, which means that it will return after the first measurement iteration, and use whatever was measured in the first iteration as the resolution estimate.

Shouldn't it be on the outside of this loop?

horenmar: Open up a new one then.

hbina:
I messed up my original fork and had to create a new one
Note to self: Never merge your own changes to your master!

PR #1511, opened 2019-01-25
fix variant detection with non-clang compiler

robinmoussu:
I noticed that stringification support for `std::variant` isn't activated if your compiler is gcc or clang >= 8. This should correct it.

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -193,9 +193,11 @@
 #      include <ciso646>
 #      if defined(__GLIBCXX__) && defined(_GLIBCXX_RELEASE) && (_GLIBCXX_RELEASE < 9)
 #        define CATCH_CONFIG_NO_CPP17_VARIANT
-#     else
+#      else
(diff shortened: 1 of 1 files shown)


robinmoussu:
Using gcc9 (not tested with other compiler), when the C++17 mode is activated, the following test fails:

```
../projects/SelfTest/UsageTests/ToStringVariant.tests.cpp:37: FAILED:
  REQUIRE( value.valueless_by_exception() )
with expansion:
  false
```

I don't understand why however.

For reference:

```c++
 30     SECTION("valueless by exception") {
 31         struct sample {
 32             operator int() const { throw 42; }
 33         };
 34
 35         type value{1.5f};
 36         REQUIRE_THROWS_AS( value.emplace<int>(sample{}), int );
 37         REQUIRE( value.valueless_by_exception() );
 38         CHECK( "{valueless variant}" == ::Catch::Detail::stringify(value) );
 39     }
```

horenmar:
The problem with the test has been solved in a different PR.

PR #1513, opened 2019-01-25
fixing issue #1507 by using FSM

### LINKED ISSUE #1507: Approx is incorrect for small margins
The following test will incorrectly fail:
REQUIRE(Approx(1.0).margin(0.0000001) != 1.000001);
This is because of the default value for epsilon in the Approx class and the fact that Approx::equalityComparisonImpl will pass if the value is in the range of the margin OR the epsilon values.
Changing the default value of epsilon to zero fixes the bug but doesn't provide the desired default behaviour for Assert.

hbina:
## Description

**WHAT**
This PR attempts to fix Issue #1507 by introducing a finite state machine (FSM) to determine the type of approximation that the user wants to perform. The default is using both and then OR'ing them.

**WHY**
As per Issue #1507 the inequality check will fail even though it is obviously correct. More simply, the equality is supposed to be wrong but Catch2 is reporting them as correct, because we performed an OR of epsilon_check and margin_check.

The problem seems to occur when the margin that is set by the user is smaller than the default epsilon, therefore, while margin_check will fail, epsilon_check will not.

## GitHub Issues

Closes #1507

### DIFF
--- a/include/internal/catch_approx.cpp
+++ b/include/internal/catch_approx.cpp
@@ -29,7 +29,8 @@ namespace Detail {
     :   m_epsilon( std::numeric_limits<float>::epsilon()*100 ),
         m_margin( 0.0 ),
         m_scale( 0.0 ),
-        m_value( value )
+        m_value( value ),
+        current_calculation_type(DEFAULT)
     {}
 
     Approx Approx::custom() {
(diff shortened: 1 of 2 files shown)


hbina:
The original Issue #1507 was marked as Resolved since this behaviour is intended. Closing the PR.

PR #1516, opened 2019-01-27
WIP: Generators take 2

horenmar:
## Description

This PR reworks the generator interface to only require "Input" generators, as opposed to the old interface, that required "RandomAccess" generators. This means that the new interface can be easily fulfilled by generators that can only be iterated once per construction and only know whether they have more inputs when they try to prepare the next one, such as generators that read inputs from a file.

_I am planning to add some extra generator adapters and more custom generator examples before merging this, but I want review for the design now._

## Issues
N/A


include/internal/catch_generators.hpp
```diff
@@ -161,6 +175,94 @@ namespace Generators {
...
+#pragma warning(push)
+#pragma warning(disable:4706) // assignment within conditoinal
+#endif
+
+        bool next() override {
+            bool success = m_generator.next();
+            if (!success) {
+                return false;
+            }
+            while (!m_predicate(m_generator.get()) && (success = m_generator.next()));
```
JoeyGrajciar: why suppress if you can do `!!(success = m_generator.next())` ?

horenmar: That wouldn't suppress the warning.

JoeyGrajciar: Then use `(success = m_generator.next()) == true`. Might not look great but will fix it :)

horenmar: I've thought about that, but I also expect that construct to run afoul of various other static analysis and style checker tools (and honestly think it is even uglier).

I'll sleep on it and come back to this.

include/internal/catch_generators.hpp
```diff
@@ -29,218 +43,329 @@ namespace Generators {
...
     template<typename T>
     class SingleValueGenerator : public IGenerator<T> {
         T m_value;
     public:
-        SingleValueGenerator( T const& value ) : m_value( value ) {}
+        SingleValueGenerator(T const& value) : m_value( value ) {}
+        SingleValueGenerator(T&& value) : m_value(std::move(value)) {}
 
-        auto get( size_t ) const -> T override {
+        T const& get() const override {
```
JoeyGrajciar: I suppose that there is no reason to derive from `SingleValueGenerator` so `get()` can be marked as final, so devirtualization can take place.

include/internal/catch_generators.hpp
```diff
@@ -16,8 +16,22 @@
...
+class GeneratorException : public std::exception {
+    const char* const m_msg = "";
+
+public:
+    GeneratorException(const char* msg):
+        m_msg(msg)
+    {}
+
+
+    const char* what() const noexcept override;
```
JoeyGrajciar: I think this can final too unless there will be more exception types deriving from `GeneratorException`.

docs/generators.md
```diff
@@ -1,50 +1,125 @@
...
+
+* `GENERATE` macro,  that serves to integrate generator expression with
+a test case,
+* 2 fundamental generators
+  * `ValueGenerator<T>` -- contains only single element
+  * `ValuesGenerator<T>` -- contains multiple elements
+* 4 generic generators that modify other generators
+  * `FilterGenerator<T, Predicate>` -- filters out elements from a generator
+  for which the predicate returns "false"
+  * `TakeGenerator<T>` -- takes `n` elements from a generator
```
JoeyGrajciar: I think that description should be ```-- takes first `n` elements from a generator```

docs/generators.md
```diff
@@ -1,50 +1,125 @@
...
+  * `MapGenerator<T, U, Func>` -- returns the result of applying `Func`
+  on elements from a different generator
+
+The generators also have associated helper functions that infer their
+type, making their usage much nicer. These are
+
+* `value(T&&)` for `ValueGenerator<T>`
+* `values(std::initializer_list<T>)` for `ValuesGenerator<T>`
+* `filter(predicate, GeneratorWrapper<T>&&)` for `FilterGenerator<T, Predicate>`
+* `take(size, GeneratorWrapper<T>&&)` for `TakeGenerator<T>`
```
JoeyGrajciar: `size` can be renamed to `count`? What do you think?

include/internal/catch_generators.hpp
horenmar: Even if there are, we can always remove it for our use. The question then is, whether it should be a user-facing customization point and I think the answer is "no".

include/internal/catch_generators.hpp
horenmar: If we don't except people deriving from `SingleValueGenerator`, the entire class can be marked `final` instead.

docs/generators.md
horenmar: Yeah, that's probably better

docs/generators.md
horenmar: Yeah, also better.

PR #1521, opened 2019-01-31
junit reporter reports random seed (--rng-seed)

### LINKED ISSUE #1520: '--reporter junit' should contain --rng-seed value
**Description**
The junit reporter should also report the value supplied to `--rng-seed`, as a continuation of #1402, which was for the XML reporter only

dosena:
Fixes #1520

### DIFF
--- a/projects/SelfTest/Baselines/junit.sw.approved.txt
+++ b/projects/SelfTest/Baselines/junit.sw.approved.txt
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<testsuitesloose text artifact
->
+<testsuites>
+  <properties>
+    <property name="random-seed" value="1"/>
+  </properties>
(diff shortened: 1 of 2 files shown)


projects/SelfTest/Baselines/junit.sw.approved.txt
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<testsuitesloose text artifact
->
+<testsuites>
+  <properties>
+    <property name="random-seed" value="1"/>
+  </properties>
+loose text artifact
```
dosena: I don't understand if there is a "hard" expectation on what should happen with `loose text artifact` here

JoeyGrajciar: I guess the only expectation is that it will be there :)

Issue #983, opened 2017-08-03
Logging macros scoped to test case instead of current scope


rcdailey:
So I'm familiar with the `INFO` macro for logging. However, these logs get lost once the enclosing scope ends. I'd instead like logs to be scoped to a `TEST_CASE` or `SECTION`. I have a logging system (basically a singleton) that I can register logging sinks to. For only catch tests, I override the logger to sink all logs to `INFO`. My hope was, that for failed test cases, it would dump logs that occurred within its own test case across our whole code base (since they all use this logger). But this doesn't work since it's nested so deep, outside of actual test case scope (but still inside; just not explicitly).

Is there some way I can make all logs in the code base funnel into Catch? Maybe an INFO_WITHIN_TESTCASE macro could be added or something? It would basically keep a list of all logs created, until the test case or section ends and then it is erased.

ideas?

horenmar:
There are couple requests for the old `INFO` behaviour to come back (perhaps as `UNSCOPED_INFO` ?), so it might eventually get in. For now and for your usecase I recommend an alternative solution: Sink all logs to std::cerr (Catch 1.9.6 doesn't properly redirect std::clog) and modify your favourite reporter to request std stream redirection (see the junit and xml reporters for examples). This should give you the desired behaviour.

rcdailey:
If I'm using latest on master as of today, does your recommendation still stand? I noticed commit 7e4038d8488596e850a11c8fe236b59bca605444 which is similar to the `cerr` issue you brought up.

rcdailey:
Also is it intended that I implement `CumulativeReporterBase`? Or is everything in `reporters/catch_reporter_bases.hpp` implementation detail?

horenmar:
Yes, current master properly redirects `std::clog` writes.


IIRC you need to implement `IStreamingReporter`, but we provide two basic reporter bases to provide common bookkeeping - `CumulativeReporterBase`, used by junit and `StreamingReporterBase` used by the other reporters. `CumulativeReporterBase` is meant to be used by reporters that need global(ish) view of results before writing and `StreamingReporterBase` is meant to be used by reporters that can write output as soon as an assertion ends.

rcdailey:
So all I need to do is implement `IStreamingReporter` and stream out logs in the test case results variable in the test case ended overload? I looked at the junit one, and that's the gist of what it seems to be doing. But it's far more complex than what I need I think.

horenmar:
[...]

PR #1522, opened 2019-02-01
Unscoped info feature

ozars:
## Description
Add assertion scoped information messaging functionality. Newly introduced `ASSERTION_INFO` macro stores messages kept until next assertion expression, that is, they are cleared right until after `assertionEnd` every time. This enables passing info message across helper functions in assertion scope, which is not allowed currently due to local scope of `INFO` macro.

If this looks reasonable, I will go ahead and add complementary `SECTION_INFO` and `TEST_CASE_INFO` macros following the same logic for sections and test cases respectively.

I've added a few test cases and updated approval tests accordingly. Please verify correctness of approval tests. They looked good to me.

## GitHub Issues
Related to #415, #983.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,7 +1,7 @@
 <a id="top"></a>
 # Logging macros
 
-Additional messages can be logged during a test case. Note that the messages are scoped and thus will not be reported if failure occurs in scope preceding the message declaration. An example:
+Additional messages can be logged during a test case. Note that the messages logged with `INFO` are scoped and thus will not be reported if failure occurs in scope preceding the message declaration. An example:
 
 ```cpp
 TEST_CASE("Foo") {
@@ -28,6 +28,60 @@ The number is 1
 ```
 When the last `CHECK` fails in the "Bar" test case, then only one message will be printed: `Test case start`.
 
+## Logging without local scope
+
+`UNSCOPED_INFO` is similar to `INFO` with two key differences:
+
+- Lifetime of an unscoped message is not tied to its own scope.
+- An unscoped message can be reported by the first following assertion only, regardless of the result of that assertion.
+
+In other words, lifetime of `UNSCOPED_INFO` is limited by the following assertion (or by the end of test case/section, whichever comes first) whereas lifetime of `INFO` is limited by its own scope.
+
+These differences make this macro useful for reporting information from helper functions or inner scopes. An example:
+
+```cpp
+void print_some_info() {
+    UNSCOPED_INFO("Info from helper");
+}
+
+TEST_CASE("Baz") {
+    print_some_info();
+    for (int i = 0; i < 2; ++i) {
+        UNSCOPED_INFO("The number is " << i);
+    }
+    CHECK(false);
+}
+
+TEST_CASE("Qux") {
+    INFO("First info");
+    UNSCOPED_INFO("First unscoped info");
+    CHECK(false);
+
+    INFO("Second info");
+    UNSCOPED_INFO("Second unscoped info");
+    CHECK(false);
+}
+```
+
+"Baz" test case prints:
+```
+Info from helper
+The number is 0
+The number is 1
+```
+
+With "Qux" test case, two messages will be printed when the first `CHECK` fails:
+```
+First info
+First unscoped info
+```
+
+"First unscoped info" message will be cleared after the first `CHECK`, while "First info" message will persist until the end of the test case. Therefore, when the second `CHECK` fails, three messages will be printed:
+```
+First info
+Second info
+Second unscoped info
+```
 
 ## Streaming macros
 
@@ -43,10 +97,14 @@ These macros come in three forms:
 
 **INFO(** _message expression_ **)**
 
-The message is logged to a buffer, but only reported with the next assertion that is logged. This allows you to log contextual information in case of failures which is not shown during a successful test run (for the console reporter, without -s). Messages are removed from the buffer at the end of their scope, so may be used, for example, in loops.
+The message is logged to a buffer, but only reported with next assertions that are logged. This allows you to log contextual information in case of failures which is not shown during a successful test run (for the console reporter, without -s). Messages are removed from the buffer at the end of their scope, so may be used, for example, in loops.
 
 _Note that in Catch2 2.x.x `INFO` can be used without a trailing semicolon as there is a trailing semicolon inside macro.
 This semicolon will be removed with next major version. It is highly advised to use a trailing semicolon after `INFO` macro._
+
+**UNSCOPED_INFO(** _message expression_ **)**
+
+Similar to `INFO`, but messages are not limited to their own scope: They are removed from the buffer after each assertion, section or test case, whichever comes first.
 
 **WARN(** _message expression_ **)**
 
--- a/include/internal/catch_capture.hpp
+++ b/include/internal/catch_capture.hpp
@@ -131,6 +131,10 @@
     Catch::ScopedMessage INTERNAL_CATCH_UNIQUE_NAME( scopedMessage )( Catch::MessageBuilder( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, Catch::ResultWas::Info ) << log );
 
 ///////////////////////////////////////////////////////////////////////////////
+#define INTERNAL_CATCH_UNSCOPED_INFO( macroName, log ) \
+    Catch::getResultCapture().emplaceUnscopedMessage( Catch::MessageBuilder( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, Catch::ResultWas::Info ) << log )
+
+///////////////////////////////////////////////////////////////////////////////
 // Although this is matcher-based, it can be used with just a string
 #define INTERNAL_CATCH_THROWS_STR_MATCHES( macroName, resultDisposition, matcher, ... ) \
     do { \
--- a/include/internal/catch_run_context.cpp
+++ b/include/internal/catch_run_context.cpp
@@ -161,6 +161,9 @@ namespace Catch {
         // and should be let to clear themselves out.
         static_cast<void>(m_reporter->assertionEnded(AssertionStats(result, m_messages, m_totals)));
 
+        if (result.getResultType() != ResultWas::Warning)
+            m_messageScopes.clear();
+
         // Reset working state
         resetAssertionInfo();
         m_lastResult = result;
@@ -215,6 +218,7 @@ namespace Catch {
 
         m_reporter->sectionEnded(SectionStats(endInfo.sectionInfo, assertions, endInfo.durationInSeconds, missingAssertions));
         m_messages.clear();
+        m_messageScopes.clear();
     }
 
     void RunContext::sectionEndedEarly(SectionEndInfo const & endInfo) {
@@ -239,6 +243,10 @@ namespace Catch {
 
     void RunContext::popScopedMessage(MessageInfo const & message) {
         m_messages.erase(std::remove(m_messages.begin(), m_messages.end(), message), m_messages.end());
+    }
+
+    void RunContext::emplaceUnscopedMessage( MessageBuilder const& builder ) {
+        m_messageScopes.emplace_back( builder );
     }
 
     std::string RunContext::getCurrentTestName() const {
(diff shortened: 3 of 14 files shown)


ozars:
I modified PR per https://github.com/catchorg/Catch2/issues/983#issuecomment-459853405.

It seems to work as intended, however, WARN/FAIL/FAIL_CHECK/SUCCEED also clears unscoped info messages since they all eventually call `RunContext::assertionEnded`. I am not sure if this should be expected behavior. It makes sense to me to disable clearing on `WARN` and keep it for other three. Suggestions?

JoeyGrajciar:
`FAIL`, `FAIL_CHECK`, `SUCCEED` and `WARN` are basically assertions. First three are to explicitly end the test case. `WARN` is special one which does not explicitly end the run but continues on failure. I guess that `WARN` was made like that to display the message even if the test passes. Is it really needed to have those unscoped info messages kept even after `WARN`?

include/internal/catch_capture.hpp
```diff
@@ -130,6 +130,12 @@
 #define INTERNAL_CATCH_INFO( macroName, log ) \
     Catch::ScopedMessage INTERNAL_CATCH_UNIQUE_NAME( scopedMessage )( Catch::MessageBuilder( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, Catch::ResultWas::Info ) << log );
 
+///////////////////////////////////////////////////////////////////////////////
+#define INTERNAL_CATCH_UNSCOPED_INFO( macroName, log ) \
+    do { \
```
JoeyGrajciar: Is there a particular reason why it is in `do {...} while(false)` statement? Can be just the
```c++
Catch::getResultCapture().emplaceUnscopedMessage( Catch::MessageBuilder( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, Catch::ResultWas::Info ) << log )
```
without the semicolon to prevent latest clang warning `-Wextra-semi-stmt`

JoeyGrajciar:
Please add also documentation about this macro to [docs/logging.md](https://github.com/catchorg/Catch2/blob/master/docs/logging.md). Also it might be good to add an example to the [examples](https://github.com/catchorg/Catch2/tree/master/examples). If you add the example add link to it in the documentation.

include/internal/catch_capture.hpp
ozars: Yes, it can, but I tend to wrap all expression macros with `do..while(0)` to be on safe side about semicolon for later modifications. It eventually comes down to personal/project coding conventions though, pretty much like preference of using or not using brackets for one-liner `if`s, `while`s,  `for`s etc.

Shall I drop do-while?

ozars:
> Is it really needed to have those unscoped info messages kept even after `WARN`?

Not really. I mixed two things. I'd prefer to have a variation of ordinary info messages that are not locally scoped (a variation of unscoped info messages not reset at each assertion in other words) so that it would behave exactly like INFO, but it could be created from helper functions different from INFO. Unscoped info doesn't behave this way by design. I understand it may be cumbersome to maintain many variations of info, so I am not pushing for this to be maintained officially.

I think "unscoped info" is somewhat a counter-intuitive name for something scoped to assertions though. I see where you are coming from, that unscoped info is something existed before, but you may consider renaming before reintroducing it given that "info" isn't also referred as "scoped info" any more. Something like "assertive info" or "assertion info" sounds more intuitive to me, but then again I never used unscoped info when it existed.

> Please add also documentation about this macro to [docs/logging.md](https://github.com/catchorg/Catch2/blob/master/docs/logging.md). Also it might be good to add an example to the [examples](https://github.com/catchorg/Catch2/tree/master/examples). If you add the example add link to it in the documentation.

Sure. I will add some documentation. I didn't see any stand alone examples just for logging messages, so I plan to add some inline examples to the documentation quoting from test cases.

ozars:
@JoeyGrajciar Please check the documentation I added.

Do you want me to squash everything or leave it as-is?

docs/logging.md
```diff
@@ -28,6 +28,58 @@ The number is 1
...
 When the last `CHECK` fails in the "Bar" test case, then only one message will be printed: `Test case start`.
 
+## Logging without local scope
+
+`UNSCOPED_INFO` is similar to `INFO` with two key differences:
+
+- Lifetime of unscoped info is not tied to its own scope.
+- Unscoped info can be reported by the first following assertion only, regardless of the result of that assertion.
+
+This macro can be used for reporting information from helper functions or inner scopes because lifetime of `UNSCOPED_INFO` is limited by the assertion following whereas lifetime of `INFO` is limited by its own scope. An example:
```
JoeyGrajciar: Split this sentence into multiple lines. Lifetime of `UNSCOPED_INFO` is not limited just by the following assertion, but also section end and test case end. 
`is limited by the assertion following` should be 'is limited by the following assertion`

docs/logging.md
```diff
@@ -43,11 +95,17 @@ These macros come in three forms:
...
 
-The message is logged to a buffer, but only reported with the next assertion that is logged. This allows you to log contextual information in case of failures which is not shown during a successful test run (for the console reporter, without -s). Messages are removed from the buffer at the end of their scope, so may be used, for example, in loops.
+The message is logged to a buffer, but only reported with next assertions that are logged. This allows you to log contextual information in case of failures which is not shown during a successful test run (for the console reporter, without -s). Messages are removed from the buffer at the end of their scope, so may be used, for example, in loops.
 
 _Note that in Catch2 2.x.x `INFO` can be used without a trailing semicolon as there is a trailing semicolon inside macro.
 This semicolon will be removed with next major version. It is highly advised to use a trailing semicolon after `INFO` macro._
 
+**UNSCOPED_INFO(** _message expression_ **)**
+
+Similar to `INFO`, but messages are removed from the buffer after each assertion and they are not limited to their own scopes.
```
JoeyGrajciar: Here should be also mentioned that section end and testcase end are clearing stored messages.

include/internal/catch_capture.hpp
ozars: Done.

docs/logging.md
ozars: Done.

docs/logging.md
ozars: Done.

docs/logging.md
```diff
@@ -32,10 +32,12 @@ When the last `CHECK` fails in the "Bar" test case, then only one message will b
...
 
-- Lifetime of unscoped info is not tied to its own scope.
-- Unscoped info can be reported by the first following assertion only, regardless of the result of that assertion.
+- Lifetime of an unscoped message is not tied to its own scope.
+- An unscoped message can be reported by the first following assertion only, regardless of the result of that assertion.
 
-This macro can be used for reporting information from helper functions or inner scopes because lifetime of `UNSCOPED_INFO` is limited by the assertion following whereas lifetime of `INFO` is limited by its own scope. An example:
+In other words, lifetime of `UNSCOPED_INFO` is limited by the following assertion (or by the end of test case/section, whichever comes first) whereas lifetime of `INFO` is limited by its own scope.
+
+These make this macro useful for reporting information from helper functions or inner scopes. An example:
```
JoeyGrajciar: `These make this macro` is about differences to `INFO`? Probably it might be written like `These differences make this macro` or `That makes this macro`

ozars: Yes, it refers to properties explained just above. It sounds okay to me, but feel free to rephrase it and push.

horenmar:
I am back, gonna take a look on this.

horenmar:
I had a hell of a time reasoning through this, and, apart from some cosmetic things that I can just push by myself, I have two remarks:

* I am not sure that unscoped messages should be cleared at the end of a `SECTION`. IMO, an unscoped message that is unused by the end of a section is the result of an ad-hoc reimplementation of generators with sections:
```cpp
TEST_CASE("foo") {
    int how_many = 0;
    SECTION("3 elements") { how_many = 3; }
    SECTION("5 elements") { how_many = 5; }
    ...
}
```

As an example, our current codebase actually has multiple instances of this pattern.


* Under current implementation, `WARN` will also remove all `UNSCOPED_INFO`s preceding it, because it is internally implemented as an assertion with specific property set. I don't think this is optimal, and instead I would suggest resurrecting the logic using `IStreamingReporter::assertionEnded` return value, see `catch_interfaces_reporter.h` lines 193-194, and `catch_run_context.cpp` lines 146-168.

ozars:
> I am not sure that unscoped messages should be cleared at the end of a `SECTION`. IMO, an unscoped message that is unused by the end of a section is the result of an ad-hoc reimplementation of generators with sections:
> ```c++
> TEST_CASE("foo") {
>     int how_many = 0;
>     SECTION("3 elements") { how_many = 3; }
>     SECTION("5 elements") { how_many = 5; }
>     ...
> }
> ```

I'm not sure if I understood this use case correctly. What would be the use of `SECTION` without any assertions in it? Also wouldn't it be semantically more reasonable to put unscoped info in between two sections instead of just before the end of the first section if one wants some info to be printed for the second section?

I agree with `WARN` case. I will update accordingly.

include/internal/catch_run_context.cpp
```diff
@@ -162,14 +162,16 @@ namespace Catch {
         // and should be let to clear themselves out.
         static_cast<void>(m_reporter->assertionEnded(AssertionStats(result, m_messages, m_totals)));
 
+        if (result.getResultType() != ResultWas::Warning)
+            m_messageScopes.clear();
+
```
ozars: Using `IStreamingReporter::assertionEnded` as-is didn't work. An example assumption in the downstream:
https://github.com/catchorg/Catch2/blob/093b72416de31a804fea15a4055bb44cb523401f/include/reporters/catch_reporter_console.cpp#L364-L366

`UNSCOPED_INFO` messages are cleared in each assertion regardless of the result with its current semantics. I could change above implementation, but it requires shotgun surgery across all reporters inheriting from `StreamingReporterBase`. This also leads to an ambiguity: If `assertionEnded` returns true, will unscoped info be cleared while info will be kept, or will both be cleared?

So, I figured out it's easier to check result type against warning instead of dealing with all these. :)

horenmar:
Multiple `SECTION`s were a way to simulate generators before Catch's generators were finished. The idea is that only one sibling section is run at a time, so it will set a new value for the variable in each run, so the complicated assertion code can be shared. That's not overly important though.

After some deliberation, I think I agree with just fixing `WARN` for now, because there are even more places that need some rework, e.g. the logic in `RunContext::handleExpr` for short-circuiting would need to handle this properly -- in the end it is probably better to get the feature in now and rewrite it later.

ozars:
Okay, I got what you meant with sections above. I will try removing message scope clear at the end of each section, but this change alone probably won't be enough if test cases are ran multiple times to iterate over sections and there are other assertions on that path before reaching the next section, which is quite likely. In this case, it may be hard to implement this without serious refactoring or ugly workarounds.

> e.g. the logic in `RunContext::handleExpr` for short-circuiting

Good thing you brought this up. I think I need to add another check against `WARN` in `RunContext::assertionPassed` for the same reason. Also just so you know, I found out by chance that current tests don't cover the execution path for that short-circuit. See the comment in above commit https://github.com/catchorg/Catch2/pull/1522/commits/093b3b898d9e1ec74f28941553dd57faff45ff74.

I am going to make above changes, update docs and rebase it against master, hopefully sometime this weekend.

horenmar:
I added some cosmetic changes that you can just squash down.

ozars:
> I will try removing message scope clear at the end of each section

This solution didn't work out as I expected unfortunately. I don't have much idea about how to implement it with major changes either. So, I will depend this behavior being not overly important.

>  I think I need to add another check against `WARN` in `RunContext::assertionPassed` for the same reason.

It turns out this was being handled correctly already.

I splitted commits to three parts. If there are no other changes requested, it is ready.

ps. I rebased it against master. I've just realized some tests are failing on appveyor due to some unrelated changes inherited from the tip of the master.

horenmar:
Yeah, sorry about that. For some reason, the tests are run in a slightly different order on Windows than they are run on Linux. I'll need to look into it, but didn't have the time just yet.

ozars:
No problem at all. I doesn't affect me. I just wanted to note it since PR fails to pass appveyor checks.

horenmar:
Merged, thanks.

ozars:
Thanks.

Issue #1523, opened 2019-02-02
approvalTests.py cannot find executable because missing -.exe


hbina:
**Describe the bug**
approvalTests.py in Windows fails to find the executable.

**Expected behavior**
It should be able to find the executable and run the unit test

**Reproduction steps**
Run approvalTests.py in CLion in Windows

**Platform information:**

 - OS: **Windows NT**
 - Compiler+version: **gcc version 8.2.0 (MinGW.org GCC-8.2.0-3)**
 - Catch version: **v2.6.0**

horenmar:
Resolved by #1524

PR #1524, opened 2019-02-02
fix an issue where approvalTests.py cannot find selfTest.exe executable

hbina:
## Description

Currently, the script for finding the executable is not OS aware. Specifically, it will fail to find it in Windows because the executable have .exe post-fix to its filename.
## GitHub Issues

Issue #1523


horenmar:
Can you undo the irrelevant changes before merging?

Basically, everything in the diff that is not the change to `getBuildExecutable` is just noise.

hbina:
Yea, my bad, its my habit of auto-formatting. I have tried to resolve it by copying the original code from master and changing only the related function. I hope the effective diff is a lot smaller now.

Issue #1514, opened 2019-01-25
The output of failing tests is not redirected to the junit reporter


SimonChh:
**Describe the bug**
Given a test that fails (either via a failing assertion or an uncaught exception), the stdout and stderr output of the test is not redirected to the reporter. For instance, when invoking the test via _-r junit_, the resulting \<system-out\> and \<system-err\> tags are empty.

The output of successful tests is redirected correctly to the reporter, however.

**Expected behavior**
Personally I expected the output of the failing test to be redirected to the reporter, since having access to the test's output helps when trying to figure out what went wrong (especially when the test is non-deterministic and thus the error is not easy to reproduce). But maybe others have different expectations, so I'm not sure if this is really a bug.

**Reproduction steps**

```
#include <Catch/catch.hpp>
#include <iostream>

SCENARIO("A failing test")
{
  std::cout << "Cout!" << std::endl;
  std::cerr << "Cerr!" << std::endl;

  throw std::runtime_error("std::runtime_error generated from test"); // REQUIRE(false) also fails
}
```

Run the above test using the _-r junit_ command-line option. The \<system-out\> and \<system-err\> tags will be empty.

**Platform information:**

 - OS: **Windows 10 Pro**
 - Compiler+version: **Visual Studio Professional 2017, version 15.7.4**
 - Catch version: **v2.5.0**

SimonChh:
FYI, the problem can be fixed by slightly changing the _RunContext::runCurrentTest()_ function:

Instead of:
```
        CATCH_TRY {
            if (m_reporter->getPreferences().shouldRedirectStdOut) {
#if !defined(CATCH_CONFIG_EXPERIMENTAL_REDIRECT)
                RedirectedStdOut redirectedStdOut;
                RedirectedStdErr redirectedStdErr;

                timer.start();
                invokeActiveTestCase();
                redirectedCout += redirectedStdOut.str();
                redirectedCerr += redirectedStdErr.str();
#else
```

I used this:
```
        CATCH_TRY {
            if (m_reporter->getPreferences().shouldRedirectStdOut) {
#if !defined(CATCH_CONFIG_EXPERIMENTAL_REDIRECT)
                RedirectedStreams redirectedStreams(redirectedCout, redirectedCerr);

                timer.start();
                invokeActiveTestCase();
#else
```
[...]

PR #1525, opened 2019-02-03
Fix issue #1514 where the output of failing tests is not redirected to the reporter

SimonChh:
## Description
Fixes issue where the output of a failing test is not redirected to the reporter (but only when `CATCH_CONFIG_EXPERIMENTAL_REDIRECT` is not defined).

The issue is caused by `invokeActiveTestCase()` throwing an exception when the test fails, so the next lines aren't ran. I fixed the error by transferring this code to the destructor of the RAII-style class `RedirectedStreams` so that it is ran even when an exception occurs.

## GitHub Issues
Closes issue #1514 (the output of failing tests is not redirected to the junit reporter)
https://github.com/catchorg/Catch2/issues/1514


include/internal/catch_run_context.cpp
```diff
@@ -67,6 +67,23 @@ namespace Catch {
         GeneratorTracker::~GeneratorTracker() {}
     }
 
+    class RedirectedStreams {
```
JoeyGrajciar: this class semantically belongs to `include/internal/catch_stream.h` with implementation inside `include/internal/catch_stream.cpp`

SimonChh: Can I move the class to catch_output_redirect.h/.cpp instead?

I need to reference the `RedirectedStdOut `and `RedirectedStdErr` classes from the header file (unfortunately I cannot forward declare them), and these two classes are defined in catch_output_redirect.h (which already #includes catch_stream.h), so I want to avoid the circular include dependency. Plus the class is not really a stream, more a form of output redirection. I could rename the class to something like `RedirectOutputToStrings` if it's clearer.

horenmar:
Okay, looks good.

horenmar:
Thanks

SimonChh:
> Thanks

No problem. Thanks to you too Martin for maintaining this library.

PR #1526, opened 2019-02-03
Reimplementing RangeGenerator

Hedanito:
## Description
The RangeGenerator was lost during the refactor of the generator interface. I've re-implemented it using the new generator interface. I have also extended it by adding an optional step size.

If it is satisfactory then I will update the documentation as well.

## GitHub Issues
Disappeared in #1516

### DIFF
--- a/include/internal/catch_generators.hpp
+++ b/include/internal/catch_generators.hpp
@@ -349,6 +349,51 @@ namespace Generators {
         );
     }
 
+    template <typename T>
+    class RangeGenerator final : public IGenerator<T> {
+        T m_current;
+        T m_end;
+        T m_step;
+        bool m_positive;
+
+    public:
+        RangeGenerator(T const& start, T const& end, T const& step):
+            m_current(start),
+            m_end(end),
+            m_step(step),
+            m_positive(m_step > T(0))
+        {
+            assert(m_current != m_end && "Range start and end cannot be equal");
+            assert(m_step != T(0) && "Step size cannot be zero");
+            assert(((m_positive && m_current <= m_end) || (!m_positive && m_current >= m_end)) && "Step moves away from end");
+        }
+
+        RangeGenerator(T const& start, T const& end):
+            RangeGenerator(start, end, (start < end) ? T(1) : T(-1))
+        {}
+
+        T const& get() const override {
+            return m_current;
+        }
+
+        bool next() override {
+            m_current += m_step;
+            return (m_positive) ? (m_current < m_end) : (m_current > m_end);
+        }
+    };
+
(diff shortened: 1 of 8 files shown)


include/internal/catch_generators.hpp
```diff
@@ -349,6 +349,53 @@ namespace Generators {
...
+        bool m_positive;
+    public:
+        RangeGenerator(T const& start, T const& end, T const& step):
+            m_current(start),
+            m_end(end),
+            m_step(step),
+            m_positive(m_step > T(0))
+        {
+            assert(m_step != T(0) && "Step size cannot be zero");
+            assert((m_positive && m_current <= m_end) || (!m_positive && m_current >= m_end) && "Step moves away from end");
```
JoeyGrajciar: you might want to enclose the or condition in parentheses so the message is reported for both checks. If i read it correctly now it should be reported only when the second condition triggers.

include/internal/catch_generators.hpp
```diff
@@ -349,6 +349,53 @@ namespace Generators {
         );
     }
 
+    template <typename T>
+    class RangeGenerator final : public IGenerator<T> {
```
JoeyGrajciar: I think that it should be limited only for arithmetic types.

JoeyGrajciar:
Add also a documentation and example for this.

include/internal/catch_generators.hpp
Hedanito: I personally prefer to keep it open for anything that provides the correct interface. This way someone can still use it with their own classes, such as a `BigInteger`. A rare use case, but I think trying to prevent someone from putting a string in it isn't going to help anyone either.

But that's my opinion, if you really want me to I can always add a `static_assert` with `std::is_arithmetic`.

Hedanito:
I have removed the possibility to use anything other than integers as requested.
The documentation has been updated as well.

horenmar:
I am back.

Apart from the two lines with hard tabs, this looks good and definitely mergeable. However, I remember from Discord that you wanted the `RangeGenerator` to work with non-integral types, while I would really rather it didn't (because the step can be hard to define, or even plain wrong, for non-integral types, such as `float`).

I think we can compromise by forcing the _helper_ to do the checking for integralness of the types -- doing so would ensure that users wouldn't be bitten by something like `range(-2.0, 7.4, 0.1)`, while they would be able to say "trust me, I know what I am doing", and construct `RangeGenerator` directly, with an arbitrary type.

horenmar:
I made the mentioned changes and squashed the PR together, gonna merge it later after it goes through CI.

PR #1529, opened 2019-02-04
Fix TAP protocol reporter

berkus:
According to TAP protocol version 13, the comments after the test name
only may contain Directives - # TODO or # SKIP. We should put
the comment (aka suite name) on a separate line before the test.

See http://testanything.org/tap-version-13-specification.html#directives


berkus:
Fail is due to travis-ci valgring builder being broken? Not related to code.

PR #1539, opened 2019-02-13
Docs: missing hyphens and full-stops added

mattclarke:
## Description
In the docs I noticed some missing hyphens, so I added them. Then I noticed that some project descriptions had full-stops and others didn't. I decided to add them where missing rather than remove them if present.

## GitHub Issues
N/A


horenmar:
Thanks, but I'll admit that I would prefer proofreading for other parts of the documentation.

PR #1541, opened 2019-02-13
use "std" namespace for "sprintf"

girtsf:
Some compilers don't export C compatibility functions in `::`. Using
`std::sprintf` should work for those (and all other compilers).

Reference: https://en.cppreference.com/w/cpp/header#C_compatibility_headers

### DIFF
--- a/include/reporters/catch_reporter_bases.cpp
+++ b/include/reporters/catch_reporter_bases.cpp
@@ -36,7 +36,7 @@ namespace Catch {
 #ifdef _MSC_VER
         sprintf_s(buffer, "%.3f", duration);
 #else
-        sprintf(buffer, "%.3f", duration);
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks. We generally make an attempt at consistently using the `std::` prefix (and the `<cfoo>` form of C headers), but it sometimes creeps in. The only platform I know of that actually cares about the difference is QNX and it is kinda hard to get a CI using it.

girtsf:
Thanks, @horenmar!

What we are using isn't QNX, but isn't too far off. It's something that starts with "vx" and ends with "works".  :)

Catch2 works on it with two minor fixes:
1) this pull request
2) https://github.com/catchorg/Clara/pull/86  (just fixed it so that it now passes the appveyor windows builds too)

Thanks for the project!

PR #1544, opened 2019-02-16
Use correct type name instead of index for TEMPLATE_PRODUCT_TEST_CASE

upsj:
## Description

Previously, for a `TEMPLATE_PRODUCT_TEST_CASE("Test" ..., T, (P1, P2))`,
the generated test case names were

    Test - 0
    Test - 1

When trying to identify the exact type causing the fail, one previously needed to resort to work-arounds like `INFO(typeid(TestType).name())`

With this commit, the correct typename is used:

    Test - T<P1>
    Test - T<P2>

The macros I used to generate the type name arrays might need a look from somebody with more preprocessor experience than me. Also, the combination with the angle brackets could probably be integrated into the `INTERNAL_CATCH_STRINGIZE_WITHOUT_PARENS` to simplify the code a bit.


upsj:
It seems like my changes have caused the execution order of some `TEMPLATE_PRODUCT` test cases to be reversed. I don't really understand why that's the case, but everything still seems to be in order with the output.

upsj:
The AppVeyor checks are failing because `cl` (incorrectly?) doesn't expand the `INTERNAL_CATCH_NOINTERNAL_CATCH_DEF` macro before stringizing. As far as I can tell, that's the only difference to clang/gcc.

horenmar:
Yeah, the "traditional" MSVC preprocessor is definitely non-conforming. The PR looks okay*, but I need to test something first, before I will merge it.


* I actually loathe the preprocessor monstrosity that is being slowly created, but we have to work with what we have, not what we wish for. 😃

Issue #1553, opened 2019-02-27
matching a non-constable value


dvirtz:
**Description**
I'm trying to match a range-v3 view which cannot be const as it mutates its state during iteration. The problem is that matchers accept their input by const reference only.

Is it possible to add additional overload which will accept the parameter by value? Or do you have any other suggestion?

**Additional context**
Version 2.6.0

horenmar:
This is one of the cases where `const_cast` is well-defined and useful, so my suggestion would be to use it for now. Other short-term solution would be using a ref-wrapper I guess.

However, you should be really careful with things that are mutated during matching, because they can easily break if you try to chain multiple matchers in a single assertion.


As for the future, we will see what we can do about this -- there already is a problem with using `const&` in assertions, because it does not play nice with type traits and other values without linkage -- but I don't know of a good solution to the problem.

dvirtz:
Thanks. A reference_wrapper is a nice idea.
I started working on a PR which replaces virtual functions with CRTP and the matcher simply forwarding its input. Would you consider accepting that?

It already compiles, I'm just working on making sure the output stays the same from the builtin matchers.

horenmar:
Hard to say from that description alone, you might want to open up a PR that highlights the intended changes so you don't waste time if there are problems.

dvirtz:
PR #1554 opened. Is there a time estimation for 3.0?

dvirtz:
Now that v3.0 is a thing, I've rebased the PR on [dev-v3](https://github.com/catchorg/Catch2/tree/dev-v3). Can you please review it now?

horenmar:
Yeah, sorry for not returning to you sooner on the issue.

I had some misgivings about the fact that your proposed PR is a breaking change to all already existing matchers. Meanwhile, back in October @melak47 offered an alternative design seen in #1843, that allows keeping the old matchers while forward extending them with templated ones.

Would the design in #1843 work for what you want to do?

dvirtz:
It would work but the performance will be much worse because of virtual functions and type erasure in #1843 . 
I ran some benchmarks to illustrate this:
Using `EqualsRangeMatcher` defined in https://github.com/melak47/Catch2/blob/43b123f83f69c7a75918c80954a106e428d27a7d/tests/SelfTest/UsageTests/Matchers.tests.cpp#L561, the following test:
```
TEST_CASE("Combining templated matchers", "[matchers][templated]") {
    std::array<int, 3> container{{ 1,2,3 }};

[...]

PR #1554, opened 2019-02-27
CRTP matchers

dvirtz:
## Description

As discussed in #1553, the current matchers implementation forces passing values by const ref. Changing the implementation to be CRTP based instead of dynamic polymorphism, enables forwarding the checked value as it is.
A nice benefit is that it is no longer required to explicitly set the value type for the `Predicate` matcher.

## GitHub Issues

#1553


dvirtz:
I've renamed the new generic matcher to `MatcherBaseGeneric` and restored `MatcherBase` for keeping backward compatibility.
I also added more tests taken from #1843 (thanks @melak47).
All the custom matchers are implemented with `MatcherBaseGeneric` which should give them better performance as no virtual functions are used.

melak47:
In my PR I found I could get away without CRTP, just having a base class from the namespace where the non-member operator overloads are defined is enough to get them picked up via ADL.

Issue #1550, opened 2019-02-24
Suggestion: Report filter usage


zivsha:
**Description**
When running the tests with any filters enabled, print the filter as part of the tests results.

Instead of this:
![image](https://user-images.githubusercontent.com/22654243/53297464-d6322000-3827-11e9-9b05-097d5b6aec70.png)

Have something like this:
![image](https://user-images.githubusercontent.com/22654243/53297553-ba7b4980-3828-11e9-8ce5-2c79c2eb7978.png)

horenmar:
I like the idea.

zivsha:
@horenmar, if you want to give me some guiding points (since I never looked at the source code) maybe I can implement this and open a PR

horenmar:
Here are the basics:
* Reporters are implemented in files in `include/reporters/`
* There are two methods that can interest you from the reporter interface, `testRunEnded` and `testRunStarted`
* Various reporters follow different convention and need different treatment.
  * I would probably skip CompactReporter
  * XMLReporter just needs a reasonable name for the tag
  * JunitReporter follows the JUnit standard (or at least our understanding of it). You would have to check if there is similar field in the JUnit XML standard, or at least a "comment" field or similar
  * ConsoleReporter looks good the way you proposed
  * TeamCity reporter uses the TeamCity format -- I am fairly sure that the format allows for arbitrary comments, but you'd have to check
  * AutoMake and TAP reporters: 🤷‍♂️ 


As for the testing and stuff, `docs/contributing.md` should be a reasonable overview.

horenmar:
This has been added to master.

johnthagen:
@horenmar Just a idea, but should `Filters:` only be output when a filter is actually used?

After upgrading to `2.7.1` I was a little surprised to yellow output at the top when I didn't set filters and all tests passed. I think it might be a nicer UX if `Filters:` is only printed if the filters are used.

Thoughts?

horenmar:
Its already fixed in master, but the release will have to wait until weekend.

zivsha:
@horenmar can you give a reference the commit that fixed this?

horenmar:
e99f1efd289944142

PR #1555, opened 2019-03-01
Support for printing test filters

zivsha:
### What?
Adding printing of the filters being used by the tests as part of the report.

### Why?
When running the tests from a non terminal interface like any IDE, the arguments passed to main are hidden under a pile of GUI and usually you don't look at them before running the executable.
If you forger that you've added a filter to the tests, you might think that all your test have passed and commit new changes, but some of the tests did not even run, and you can miss it in case the number of tests that have passed is close enough to the total number of tests/assertions (or you simply don't know how many assertions should pass). 

Having a bright yellow indication of filters that were in use makes the report more complete and also helps users avoid scenarios like the one depicted above.

## GitHub Issues
Implementation of suggestion in issue #1550.

### DIFF
--- a/include/internal/catch_test_spec.h
+++ b/include/internal/catch_test_spec.h
@@ -22,10 +22,16 @@
 
 namespace Catch {
 
+	struct PatternDescription {
+		std::string name;
+		std::string pattern;
+	};
+
     class TestSpec {
         struct Pattern {
             virtual ~Pattern();
-            virtual bool matches( TestCaseInfo const& testCase ) const = 0;
+			virtual bool matches(TestCaseInfo const& testCase) const = 0;
+			virtual PatternDescription description() const = 0;
         };
         using PatternPtr = std::shared_ptr<Pattern>;
 
@@ -34,7 +40,8 @@ namespace Catch {
             NamePattern( std::string const& name );
             virtual ~NamePattern();
             virtual bool matches( TestCaseInfo const& testCase ) const override;
-        private:
+			virtual PatternDescription description() const override;
+		private:
             WildcardPattern m_wildcardPattern;
         };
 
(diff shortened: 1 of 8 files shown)


zivsha:
Partial implementation, closing PR for the moment

include/internal/catch_test_spec.h
```diff
@@ -22,10 +22,16 @@
...
+	struct PatternDescription {
+		std::string name;
+		std::string pattern;
+	};
+
     class TestSpec {
         struct Pattern {
             virtual ~Pattern();
-            virtual bool matches( TestCaseInfo const& testCase ) const = 0;
+			virtual bool matches(TestCaseInfo const& testCase) const = 0;
```
horenmar: We use spaces, not tabs.

zivsha: Fresh installation of VS, me too. I'll untabify it.

horenmar:
Also one more comment: The resulting filter string should preferably be roundtrippable. This means that user can just copy-paste the output string onto CLI and get the same behaviour.

zivsha:
> Also one more comment: The resulting filter string should preferably be roundtrippable. This means that user can just copy-paste the output string onto CLI and get the same behaviour.

Does it matter if the order is not preserved?

horenmar:
No, but tags should be inside square brackets, test names with spaces should probably be inside quotes and so on.

zivsha:
@horenmar , any suggestions for testing this?

horenmar:
Well, there are two parts to this that need testing

1) That the output makes sense
2) That the output roundtrips, in other words, I can copy the filter string as the argument to the test binary and get the same output

The first can be done via ApprovalTests in the same way all other output sanity checks are done. The second will need to be a separate test under CTest, like the one for #1492, and if you implement the feature, I should be able to implement it myself fairly quickly.

zivsha:
I hope I'll have some spare time soon to work on this 😉

zivsha:
@horenmar can you reopen this PR or should I open a new one?

horenmar:
I can't either.

PR #1559, opened 2019-03-07
 ParseAndAddCatchTests enhancements

drdanz:
A few enhancements to the `ParseAndAddCatchTests` CMake module:

#### Set the `ParseAndAddCatchTests_TESTS` property

After the script, the `ParseAndAddCatchTests_TESTS` property for the target, and for each source file in the target is set, and contains the list of the tests extracted from that target, or from that file.
This is useful, for example to add further labels or properties to the tests.

#### Fix tests not found

When using an optional launcher the target name is not interpreted as a target, therefore it is not replaced automatically with its path.

#### Do not change `CMAKE_MINIMUM_REQUIRED_VERSION` when included


horenmar:
Thanks

PR #1562, opened 2019-03-08
Update url of blog post

claremacrae:


horenmar:
Avoiding a superfluous redirect is always good, thanks.

PR #1563, opened 2019-03-08
Documentation improvements

claremacrae:
## Description

Minor changes to to the documentation, to update a URL that was redirecting, and other small things like:

* expanding an abbreviation that not all readers will know
* remove commas so that reading a sentence flows
* adding the odd missing word

## GitHub Issues

None


horenmar:
Thanks.

PR #1565, opened 2019-03-09
Added override keyword near IntRange::describe() method

inobelar:
## Description

Because `Catch::MatcherBase<T>::describe()` is virtual method, when someone copy-paste this example code for changing, typically code-analyzer produces warning about [`IntRange::describe()`](https://github.com/catchorg/Catch2/blob/master/docs/matchers.md#custom-matchers) `override`ing.


horenmar:
Sure, thanks.

PR #1574, opened 2019-03-18
raspigcd - opensource-users

pantadeusz:
Added raspigcd to the opensource-users that uses Catch2 for tests

## Description

I would like to include my little project (described here https://hackaday.com/2018/05/15/direct-cnc-control-with-the-raspberry-pi/) in the list of usages of Your awesome test lib. Raspigcd uses older version of Catch. The new version (not yet mentioned here) will use Catch2.


horenmar:
Thanks, on both the PR and the praise 😃

PR #1587, opened 2019-04-03
Add RenderDoc to list of open source users.

Jason2013:
## Description
Add RenderDoc to list of open source users.

RenderDoc  source code reference: https://github.com/baldurk/renderdoc/blob/v1.x/renderdoc/3rdparty/catch/official/catch.hpp

## GitHub Issues
Not related to GitHub Issues.


horenmar:
Thanks for the PR, but we strongly prefer if projects are added to the user list by their maintainers, which in this case is Baldurk.

PR #1592, opened 2019-04-07
Use correct filename in CMake integration docs

robinlinden:
The header at https://github.com/catchorg/Catch2/blob/master/docs/cmake-integration.md#catchcmake-and-addcatchtestscmake should be naming https://github.com/catchorg/Catch2/blob/master/contrib/CatchAddTests.cmake


horenmar:
Thanks

PR #1596, opened 2019-04-10
IndexOfHideLabel is never used

50ty:
## Description

Drop `list(FIND Labels "!hide" IndexOfHideLabel)` due to it's return value `IndexOfHideLabel` is never used. So the find in the list is unnecessary.


horenmar:
Fair

PR #1599, opened 2019-04-13
Allow 'make package' to create an installer

ThijsWithaar:
I'd like to be able to generate a .deb package from the latest source.
This allows me to both install and uninstall the package in a consistent way.

This change adds the CPack statements to allow 'make package' to generate a .deb on debian systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,9 +205,7 @@ if (NOT_SUBPROJECT)
         ${PKGCONFIG_INSTALL_DIR}
     )
 
-    if(EXISTS "/etc/debian_version")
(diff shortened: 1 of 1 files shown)


horenmar:
I am for adding CPack support, but I don't think `deb` should be assumed if `/etc/debian_version` is present in the system, because AFAIU it would also override the generator passed to CPack via the `-G` option flag.

ThijsWithaar:
Ok, I've remove the auto-setting, but kept it as (default-off) flag.

horenmar:
I redid the changes differently in c765381, there is no option in CMake and you need to use the `-G` option in CPack.

PR #1603, opened 2019-04-17
Adding a note about TEMPLATE_TEST_CASE

ScottHutchinson:
Adding a note that the TEMPLATE_TEST_CASE was added in version 2.5.0, so people with an older version might know why the TEMPLATE_TEST_CASE will not compile. Also corrected spelling mistakes.


horenmar:
Thanks, but I don't think that adding versioning to a single feature is a good way forward.

ScottHutchinson:
I see no harm in it. However, I might agree that providing a central summary of new features added in each version might be even better. There are release notes, but they appear to be out of date.

horenmar:
I am going to close this one in favour of #1695 which actually adds these notes for all features introduced in v2.3.0 or later.

PR #1605, opened 2019-04-21
TrackerContext: Remove misleading instance static method.

textshell:
TrackerContext is not used as singleton, so just remove this misleading method.

While working on an prove of concept for something in catch, I was confused by this no longer needed method leading me to think that TrackerContext was a singleton. But it's not.


horenmar:
Yeah, the code seems to be some evolutionary artifact from the past. 

Thanks for PR.

PR #1606, opened 2019-04-24
fix  example's mistake in  slow-compiles.md

phiysng:
the example  in docs/slow-compiles.md miss to link factorial.o file which lead to an undefined reference error.

I don't find relevent issue and hope I don't get it wrong with cpp's linker.


horenmar:
Thanks. The handling of the original file was indeed missing and this should fix it.

Just an FYI, you can compile to object file multiple files at once, like so `g++ factorial.cpp tests-main.cpp -c`.

PR #1611, opened 2019-04-24
Add UNSCOPED_INFO to prefix/disabled sections

cincodenada:
## Description
The new `UNSCOPED_INFO` feature was only added for the default usage. This adds defines so that it works under `#define CATCH_CONFIG_PREFIX_ALL` and/or `#define CATCH_CONFIG_DISABLE`.

The `X02-DisabledMacros.cpp` is pretty thin and doesn't test individual macros, so only the `CATCH_`-prefixed modification is accounted for in tests. Expanding that test seemed out scope for this PR, but I'm happy to be the first if that's useful.

## GitHub Issues

Amends #1522, which introduced `UNSCOPED_INFO`

### DIFF
--- a/include/catch.hpp
+++ b/include/catch.hpp
@@ -132,6 +132,7 @@
 #endif // CATCH_CONFIG_DISABLE_MATCHERS
 
 #define CATCH_INFO( msg ) INTERNAL_CATCH_INFO( "CATCH_INFO", msg )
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks for the fix. Also if you are willing, the tests could definitely use some expanding. There is always a trade-off in the amount of work that goes into them and apparently the X02 tests were on the wrong side.

horenmar:
Just so we are on the same page: should I merge this PR and wait for a new one, or leave this one open for the new tests?

PR #1617, opened 2019-04-28
Listener docs: overrides should call base class functions

jbytheway:
## Description
Documentation change only.

Overrides in event listeners should call the base class versions of the functions.  Failing to do so can result in strange errors.

Emphasize this by including such calls in the example code.


horenmar:
Can you give some examples of the errors? AFAIK there is no requirement to do that, and if there is, it shouldn't be there.

jbytheway:
If you override `sectionStarting` (without calling the base class version) but not `sectionEnded` then it tries to pop from the section stack without pushing to it, which crashes.

Observed [here](https://github.com/CleverRaven/Cataclysm-DDA/pull/30006).

PR #1621, opened 2019-05-04
Remove explicit setting of CXX_STANDARD for SelfTest target

jayeshbadwaik:
- The current setup tries to detect USE_CPP14/USE_CPP17 and sets the
    CXX_STANDARD property for the SelfTest target. This is not ideal, since
    CMAKE_CXX_STANDARD can be provided by the toolchain file or as command line
    option and should be used by the library internally correctly.  Hence, the
    whole set of the relevant lines from `projects/CMakeLists.txt` have been
    removed.

  - The above can also cause subtle issues where the user is expecting the tests
    to compile with C++17 after setting CMAKE_CXX_STANDARD and then getting
    results of compilation with C++11 as USE_CPP17 has not been set.

  - The current build matrix used the above code to run the tests. So, even
    though the it should not required anymore to build Catch2, it was still
    required to send correct options to build matrix. In that respect,
    .travis.yml has been modified to send correct options to the build command
    in the new setup.


jayeshbadwaik:
Only after submitting the PR did I realize that there was a AppVeyor test suite as well. I unfortunately am not well-versed in the `.bat` files, neither do I have a platform to test it, so I have not touched it. You might require to copy the changes [here](https://github.com/catchorg/Catch2/pull/1621/commits/2ce0c935efa8f0a19b01f03c7d477ead7abc8be2#diff-354f30a63fb0907d4ad57269548329e3L305) to the appveyor script.

PR #1651, opened 2019-06-05
Fix sectionEnded() not popping corresponding section from sectionStarted() in case where error was thrown in some section.

DanielPBak:
## Description

In a failing test, the sectionend() corresponding to a sectionstart() could possibly pop different data than was pushed. This is because sectionEndedEarly() would push the failing section onto another stack (m_unfinishedSections), and then the next sectionEnded() would close the hanging sectionStarted(). Then the m_unfinishedSections would be popped, filling up the remaining hanging sections.

This caused two issues:
	
One: any reporter relying on SectionEnd() having accurate section data was broken. SectionEnd() would present different data than SectionStarted(). This caused an issue for the junit reporter, which relies on that data matching to manage the parralel m_sectionStack 

However, this also applied to other reporters; having the xml reporter write the "name" attribute to the OverallResults written in sectionEnded revealed that it was not matching the section.

Two: I noticed where tests containing failing sections would pass through again, but not hit any nodes. Since a failing section would not be popped, it would remain on the stack, and this causes the test runner to believe that the test requires another pass because there are unresolved open sections (when all sections were actually closed).

This PR fixes both of these problems and does not seem to have any down side. Is there some edge case where it is important to pop the stack out of order? To me this looks like legacy code that is no longer useful.

The change I made was just to always call sectionEnded() and never sectionEndedEarly().

In my opinion, this needs a larger workaround to avoid introducing similar bugs. sectionEnded() shouldn't consume anything and should automatically pop the top.

Sorry about the verbose description.

## GitHub Issues

Closes #1650
Closes #417
It also looks related to #1210

Issue #1650, opened 2019-06-04
Junit Reporter breaks when BDD Sections fail


dbak-invn:
**Describe the bug**
When a failure occurs, all previously-occurring tests within the same top-level section have their names messed up in a particular way. For example, the following test:

```
SCENARIO("This is a fake test", "[fake]"){
    GIVEN("Prior given"){
        REQUIRE(true);
        THEN("Prior Given Inner"){
            REQUIRE(true);
        }
    }
    GIVEN("Given1"){
        THEN("Beep Boop"){
            REQUIRE(true);
        }
        WHEN("When1"){
            THEN("Success"){
                REQUIRE(true == true);
            }
            THEN("PRIOR WITH INNER"){
                AND_THEN("inner prior"){
                    REQUIRE(true);
                }
            }
            THEN("FAIL"){
                AND_THEN("SUCCEED INNER"){
                    REQUIRE(true);
                }
                AND_THEN("FAIL INNER") {
                    FAIL("FAILURE");
                }
            }
            THEN("THEN AFTER FAILURE"){
                AND_THEN("inner after") {
                    REQUIRE(true);
                }
            }
            THEN("LAST"){
                REQUIRE(true);
            }
        }
    }
    GIVEN("Given 2"){
        REQUIRE(true);
    }
}
  THEN("FAIL"){
                AND_THEN("FAIL INNER") {
                    FAIL("FAILURE");
                }
            }
            THEN("THEN AFTER FAILURE"){
                AND_THEN("inner after") {
                    REQUIRE(true);
                }
            }
            THEN("LAST"){
                REQUIRE(true);
            }
        }
    }
    GIVEN("Given 2"){
        REQUIRE(true);
    }
}
```

Produces the following junit xml output:

```
<testsuites>
<testsuite name="flamenco_algorithms_anomaly_stiction_tests" errors="0" failures="1" tests="10" hostname="tbd" time="0.000328" timestamp="2019-06-04T01:09:41Z">
<properties>
<property name="filters" value="[fake]"/>
</properties>
<testcase classname="flamenco_algorithms_anomaly_stiction_tests.global" name="Scenario: This is a fake test/Given: Prior given" time="0.000029"/>
<testcase classname="flamenco_algorithms_anomaly_stiction_tests.global" name="Scenario: This is a fake test/Given: Prior given/Then: Prior Given Inner" time="0.000002"/>
[...]

Issue #417, opened 2015-05-18
Junit reporter: sections are mixed


tomgeb:
When using the Junit reporter and nested sections, test case names are generated in the form "outer section/nested section level 1/..." a.s.o.

When any expression in the whole test case fails, this seems to be broken.

Minimal test case to reproduce the problem:

    TEST_CASE( "catch a bug" )
    {
        	SECTION( "outer section" )
	        {
		        SECTION( "inner section 1" ) { REQUIRE( 1 == 1 ); }
         		SECTION( "inner section 2" ) { REQUIRE( 2 == 2 ); }
        	}
    }

produces

    <testsuites>
      <testsuite name="all tests" errors="0" failures="0" tests="2" hostname="tbd" time="6.4e-05" timestamp="tbd">
        <testcase classname="catch a bug" name="outer section/inner section 1" time="0.000013"/>
        <testcase classname="catch a bug" name="outer section/inner section 2" time="0.000004"/>
        <system-out/>
        <system-err/>
      </testsuite>
    </testsuites>

However,

    TEST_CASE( "catch a bug" )
    {
        	SECTION( "outer section" )
	        {
		        SECTION( "inner section 1" ) { REQUIRE( 1 == 1 ); }
         		SECTION( "inner section 2" ) { REQUIRE( 2 == 3 ); }
        	}
    }

produces 

    <testsuites>
      <testsuite name="all tests" errors="0" failures="1" tests="2" hostname="tbd" time="0.000109" timestamp="tbd">
        <testcase classname="catch a bug" name="inner section 2/inner section 1" time="0.000014"/>
        <testcase classname="catch a bug" name="inner section 2/outer section" time="0.000042">
          <failure message="2 == 3" type="REQUIRE">at ../libs/adasmath/test/camera_models.cpp:81      </failure>
        </testcase>
        <system-out/>
        <system-err/>
      </testsuite>
    </testsuites>

Note the section names

**outer section/inner section 1**
**outer section/inner section 2**

vs. 

**inner section 2/inner section 1**
**inner section 2/outer section**.

PR #1654, opened 2019-06-06
#1650 Approval Tests

DanielPBak:
## Description
I may not have the time shortly to find a good solution for https://github.com/catchorg/Catch2/pull/1651

So this is a separate PR containing only the tests which reveal the erroneous behavior I am trying to fix.

I expect that this will fail CI as this PR is adding a test which is failing against master. However, I believe that it should be merged because the behavior in master is legitimately failing to produce expected results.
## GitHub Issues
Shows #1650
Shows #417


horenmar:
Ok, I see the issue. However, the tests will have to be part of #1651, as merging failing tests that will not be fixed within a known (and short) time-frame would provide too much impedance for further development.

Issue #1590, opened 2019-04-05
Make catch tags work with ctest labels


pterodragon:
```
include(Catch)
catch_discover_tests(all_tests)
```
The above is good to include all the test names into ctest so that `ctest -N` lists the test names. But it doesn't work with tags. ctest provides the LABEL property for a test so maybe the idea of tags in Catch is close to that.
`ctest --print-labels` lists all the test labels and `ctest -L <regex>` runs tests with labels matching the regex.

horenmar:
That sounds like an interesting and useful feature for the `contrib/` scripts.

horenmar:
This has been added by @MaciejPatro.

horenmar:
As it turns out, the original PR for this issue introduced bugs to the original functionality due to various factors, so it was reverted.

There are also problems with adding tags as labels to the tests because of [CMake upstream issue](https://gitlab.kitware.com/cmake/cmake/issues/19391)

Quincunx271:
The upstream CMake issue has been merged into CMake 3.18. As such, we can go ahead and pull this feature back online after verifying the CMake version. As I mentioned in #1658, the LABELS should be APPEND-ed, not just set. See the partial implementation here: https://github.com/catchorg/Catch2/pull/1658#issuecomment-502490071 . That implementation should work after some adjustments.

sourcedelica:
Hi @Quincunx271 - what's the status of this?  I tried using the `Catch.cmake` and `CatchAddTests.cmake` from your partial implementation (https://github.com/Quincunx271/Catch2/blob/3047ee81052b35f9867b61c15a81c9df2b539621/contrib/CatchAddTests.cmake) with CMake 3.21.0, but when I run `ctest` I get errors like:
```
$ ctest -N
Test project /home/eric/work/tw-base/cmake-build-debug
CMake Error at /home/eric/work/tw-base/cmake-build-debug/tw/utils/test/utils-test_tests-b858cb2.cmake:65 (set_property):
  set_property given TEST names that do not exist:

    utils:Binary Decoding
    utils:Binary Encoding
    utils:Binary Encoding Types
    utils:Binary Encoding sequence

Call Stack (most recent call first):
  /home/eric/work/tw-base/cmake-build-debug/tw/utils/test/utils-test_include-b858cb2.cmake:2 (include)
  /home/eric/work/tw-base/cmake-build-debug/tw/utils/test/CTestTestfile.cmake:7 (include)
[...]

PR #1658, opened 2019-06-16
Fix catch_discover_tests() from breaking on long test names

### LINKED ISSUE #1645: catch_discover_tests breaks when test description is long (wraps with --list-tests)
**Describe the bug**

Long test descriptions are mis-parsed by CatchAddTests.cmake and are parsed as multiple tests (of which neither is actually a test).

This is because `--list-tests`` wraps long test descriptions, and CatchAddTests.cmake assumes each test description is on a single line.

**Expected behavior**

Either `--list-tests` should have a way to forbid wrapping of test descriptions, or CatchAddTests.cmake needs more work

**Reproduction steps**

Running the test executable with `--list-tests` wraps the test description if the string is long enough:

https://github.com/catchorg/Catch2/blob/00347f1e79260e76d5072cca5b3636868397dda5/include/internal/catch_list.cpp#L45

However, CatchAddTests.cmake assumes the tests are newline delimited:

https://github.com/catchorg/Catch2/blob/00347f1e79260e76d5072cca5b3636868397dda5/contrib/CatchAddTests.cmake#L82

This means that a test case with a long enough name is parsed as two test cases.

**Platform information:**

 - Catch version: **v2.8.0**

Quincunx271:
Fixes #1645 by rolling back d4eec016a9df54a560d82e5ee12aef42822525e7 ; reopens #1590 .

Parsing `--list-tests` is broken, as Catch automatically line wraps the line when it gets too long, stripping any whitespace in the process. This means that it's impossible to reproduce the exact name of the
test if the test's name is long enough to line-wrap.

Furthermore, overwriting the `LABELS` property with the discovered labels breaks users who manually added custom ctest labels. However, this seems to not actually overwrite the labels due to buggy handling of our strange test names by CMake.

Rolling back to using `--list-test-names-only` for now, as it does not wrap lines even on very long test names.

We may be able parse the output of `--list-tags` to produce the ctest labels. However, the straightforward way of doing this is to use CMake's `get_property(TEST ...)` and `set_property(TEST ... APPEND ...)`, which don't work if the test name has spaces or other special characters. Such test names may be illegal test names anyway&mdash;the [documentation](https://cmake.org/cmake/help/latest/command/add_test.html) doesn't state whether it's allowed or disallowed for the variant of `add_test(...)` we are using&mdash;although they appear to work.

Upstream CMake issue on allowing custom strings as test names, or at least custom test descriptions: https://gitlab.kitware.com/cmake/cmake/issues/19391

### DIFF
--- a/contrib/CatchAddTests.cmake
+++ b/contrib/CatchAddTests.cmake
@@ -22,39 +22,6 @@ function(add_command NAME)
   set(script "${script}${NAME}(${_args})\n" PARENT_SCOPE)
 endfunction()
 
-macro(_add_catch_test_labels LINE)
-  # convert to list of tags
-  string(REPLACE "][" "]\\;[" tags ${line})
-
-  add_command(
-    set_tests_properties "${prefix}${test}${suffix}"
-      PROPERTIES
-        LABELS "${tags}"
-  )
-endmacro()
-
-macro(_add_catch_test LINE)
-  set(test ${line})
-  # use escape commas to handle properly test cases with commans inside the name
-  string(REPLACE "," "\\," test_name ${test})
-  # ...and add to script
-  add_command(
-    add_test "${prefix}${test}${suffix}"
-      ${TEST_EXECUTOR}
-       "${TEST_EXECUTABLE}"
-       "${test_name}"
-       ${extra_args}
-     )
-
-  add_command(
-    set_tests_properties "${prefix}${test}${suffix}"
-      PROPERTIES
-        WORKING_DIRECTORY "${TEST_WORKING_DIR}"
-        ${properties}
-  )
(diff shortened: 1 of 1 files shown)


Quincunx271:
I have a [partial implementation](https://github.com/Quincunx271/Catch2/tree/fix-catch-discover-tests-with-labels-from-tags) of making Catch labels work as CTest labels: https://github.com/Quincunx271/Catch2/blob/3047ee81052b35f9867b61c15a81c9df2b539621/contrib/CatchAddTests.cmake . But this is broken due to the handling of `get_property(TEST ...)` and `set_property(TEST ... APPEND ...)` as mentioned above.

horenmar:
Looks good and it is good to hear that the test name restriction is likely to be lifted.

PR #1673, opened 2019-06-29
Fix TrackerBase::close children completion check

### LINKED ISSUE #1670: Inconsistent behaviour of nested sections
**Describe the bug**
Suppose I have a test case with a structure like the following:
```
TEST_CASE("Nesting")
{
    SECTION("A") {
        SECTION("1") SUCCEED();
        SECTION("2") SUCCEED();
    }
    SECTION("B") {
        SECTION("1") SUCCEED();
        SECTION("2") SUCCEED();
    }
}
```

Upon selecting the test case and section on the command line, I get the following results:
* `./test Nesting`: All four assertions are executed;
* `./test Nesting -c A`: *Only A1 is executed*;
* `./test Nesting -c B`: B1 and B2 are executed.

**Expected behavior**
`./test Nesting -c A` should execute all subsections of section A.

**Reproduction steps**
See description.

**Platform information:**
 - Compiler+version: **GCC v9.1.1**
 - Catch version: **v2.8.0**, **v2.9.1**

sfranzen:
## Description
The function now verifies that all children have completed, instead of only the last. This fixes the section exeuction bug I reported the other day.

## GitHub Issues
Fixes #1670.

### DIFF
--- a/include/internal/catch_test_case_tracker.cpp
+++ b/include/internal/catch_test_case_tracker.cpp
@@ -139,7 +139,7 @@ namespace TestCaseTracking {
                 m_runState = CompletedSuccessfully;
                 break;
             case ExecutingChildren:
-                if( m_children.empty() || m_children.back()->isComplete() )
+                if( std::all_of(m_children.begin(), m_children.end(), [](ITrackerPtr const& t){ return t->isComplete(); }) )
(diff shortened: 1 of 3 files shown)


horenmar:
Thanks for the fix, I added a test for it to prevent later regressions.

PR #1680, opened 2019-07-05
Ignore ws

philsquared:
## Description
Whitespace at the start and end of a test case or section name should not be significant when matching from the command line. This is particularly significant with the BDD section names, since they hardcode in leading whitespace for alignment purposes - making it very hard to run specific sections.

So this change trims the names as they are being compared - and also trims the patterns coming into from the CLI/ config.

### DIFF
--- a/include/internal/catch_config.cpp
+++ b/include/internal/catch_config.cpp
@@ -15,11 +15,23 @@ namespace Catch {
     :   m_data( data ),
         m_stream( openStream() )
     {
+        // We need to trim filter specs to avoid trouble with superfluous
+        // whitespace (esp. important for bdd macros, as those are manually
+        // aligned with whitespace).
+
+        for (auto& elem : m_data.testsOrTags) {
+            elem = trim(elem);
+        }
+        for (auto& elem : m_data.sectionsToRun) {
+            elem = trim(elem);
+        }
+
         TestSpecParser parser(ITagAliasRegistry::get());
-        if (!data.testsOrTags.empty()) {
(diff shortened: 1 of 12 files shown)


include/internal/catch_config.cpp
```diff
@@ -9,17 +9,30 @@
...
+            &trim );
+
         TestSpecParser parser(ITagAliasRegistry::get());
-        if (!data.testsOrTags.empty()) {
+        if (!m_testsOrTags.empty()) {
             m_hasTestFilters = true;
-            for( auto const& testOrTags : data.testsOrTags )
-                parser.parse( testOrTags );
+            for( auto const& testOrTags : m_testsOrTags )
+                parser.parse( trim( testOrTags ) );
```
horenmar: Aren't these already trimmed?

Also now there are helpfully two fields named `testsOrTag` in closely related classes...

include/internal/catch_test_case_tracker.cpp
```diff
@@ -190,7 +191,7 @@ namespace TestCaseTracking {
 
         if ((m_filters.empty() || m_filters[0] == "") ||
              std::find(m_filters.begin(), m_filters.end(),
-                       m_nameAndLocation.name) != m_filters.end())
+                       trim(m_nameAndLocation.name)) != m_filters.end())
```
horenmar: Similarly, why isn't this trimmed at construction, rather than on each call to `isComplete`?

horenmar:
Merged manually

PR #1684, opened 2019-07-09
Improve reporting of unmatched filters

### LINKED ISSUE #1449: Catch2 does not report unmatched parts of a partially-matched test spec
## Description
Catch2's runTests function calls the reporter's noMatchingTestCases if *no* tests match the spec, but if some test matches some part of the spec, nothing reports that the rest of the spec matched nothing. In other words, if I pass "Test1 Test2" as command-line args, and "Test1" matches some test, nothing complains about the fact that "Test2" matched nothing. This can lead the user to believe that they're running all the tests in the spec when they are not.

### Steps to reproduce
Create a test file with one test. Run it with command-line args that specify two tests: the existing test and
some other, non-existent test. Everything will be green and happy, with no indication that the second test was not found.

### Extra information
* Catch version: **v2.4.1**
* Operating System: **Windows 10**
* Compiler+version: **MS VS 2015**

### LINKED ISSUE #1683: -w NoTests should give non-zero status
docs/command-line.md says "NoTests        // Return non-zero exit code when no test cases were run" but I see a status of zero being returned instead of non-zero

my_bug.cpp:
```#define CATCH_CONFIG_MAIN
#include "catch2/catch.hpp"

TEST_CASE("my case", "[foo]") {
  SECTION("my section") {
    CHECK(1);
  }
}
```
```
> g++ -o my_bug my_bug.cpp -I ...
> ./my_bug -w NoTests '[x]'
Filters: [x]
No test cases matched '[x]'
No tests ran

status=0
```

I expected a non-zero status.

**Platform information:**
 - OS: **Linux CentOS 6**
 - Compiler+version: **GCC v7.3.0**
 - Catch version: **v2.9.1**

sfranzen:
## Description
This PR is for two issues related to filtering tests, reported in issue #1449.

It modifies various existing structures to allow checking whether any filters passed to a test executable did not result in any matching test cases, giving the user better feedback about the input. It also adds small helper classes to aid readability of the code. Additionally, the option `-w NoTests` is now correctly checked and an exit code 2 returned if there were any unmatched filters.

## GitHub Issues
Fixes #1449. Also fixes #1683.

### DIFF
--- a/include/internal/catch_session.cpp
+++ b/include/internal/catch_session.cpp
@@ -59,13 +59,14 @@ namespace Catch {
             return ret;
         }
 
-        class TestSet {
+        class TestGroup {
         public:
-            explicit TestSet(IConfig const& config, RunContext& context)
-            : m_context{context}
+            explicit TestGroup(std::shared_ptr<Config> const& config)
+            : m_config{*config}
+            , m_context{config, makeReporter(config)}
             {
-                auto const& allTestCases = getAllTestCasesSorted(config);
-                m_matches = config.testSpec().matchesByFilter(allTestCases, config);
+                auto const& allTestCases = getAllTestCasesSorted(m_config);
+                m_matches = m_config.testSpec().matchesByFilter(allTestCases, m_config);
 
                 if (m_matches.empty()) {
                     for (auto const& test : allTestCases)
@@ -77,8 +78,9 @@ namespace Catch {
                 }
             }
 
-            Totals execute() const {
+            Totals execute() {
                 Totals totals;
+                m_context.testGroupStarting(m_config.name(), 1, 1);
                 for (auto const& testCase : m_tests) {
                     if (!m_context.aborting())
                         totals += m_context.runTest(*testCase);
@@ -92,28 +94,18 @@ namespace Catch {
                         totals.error = -1;
                     }
                 }
+                m_context.testGroupEnded(m_config.name(), totals, 1, 1);
                 return totals;
             }
 
         private:
             using Tests = std::set<TestCase const*>;
 
-            RunContext& m_context;
+            Config const& m_config;
+            RunContext m_context;
             Tests m_tests;
             TestSpec::Matches m_matches;
         };
-
-        Catch::Totals runTests(std::shared_ptr<Config> const& config) {
-            RunContext context(config, makeReporter(config));
-
-            context.testGroupStarting(config->name(), 1, 1);
-
-            TestSet testSet {*config, context};
-            auto const totals = testSet.execute();
-
-            context.testGroupEnded(config->name(), totals, 1, 1);
-            return totals;
-        }
 
         void applyFilenamesAsTags(Catch::IConfig const& config) {
             auto& tests = const_cast<std::vector<TestCase>&>(getAllTestCasesSorted(config));
@@ -290,7 +282,8 @@ namespace Catch {
             if( Option<std::size_t> listed = list( m_config ) )
                 return static_cast<int>( *listed );
 
-            auto totals = runTests( m_config );
+            TestGroup tests { m_config };
+            auto const totals = tests.execute();
 
             if( m_config->warnAboutNoTests() && totals.error == -1 )
                 return 2;
--- a/include/internal/catch_session.cpp
+++ b/include/internal/catch_session.cpp
@@ -25,6 +25,7 @@
 
 #include <cstdlib>
 #include <iomanip>
+#include <set>
 
 namespace Catch {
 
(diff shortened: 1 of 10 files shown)


sfranzen:
The remaining failure is with the [AppVeyor coverage build](https://ci.appveyor.com/project/CatchOrg/catch2/builds/25887374/job/iathytno44bpykih#L462) and is caused by OpenCppCoverage marking the exit code as an error where CTest does not (because it has `WILL_FAIL TRUE`). Is there an elegant way to ignore the exit code in this specific case?

horenmar:
I'll investigate the problem -- it is my understanding that both OpenCppCoverage and the wrapper we use for CTest should just pass-through the return codes back to CTest, so it should not fail.

horenmar:
After closer look, the helper executable does not actually pass the return code up the stack... I'll try to fix it tomorrow.

sfranzen:
In the meantime I've downloaded this coverage tool and played with it, and I could also obtain a coverage report for all tests by running `OpenCppCoverage [coverage options] -- ctest [ctest options]`. CTest then hosts the individual test commands and completes with return code 0 as long as all test cases return 0 or else have `WILL_FAIL TRUE`. This seems a lot simpler than the current setup, though I haven't studied it in full detail -- would it be significantly different?

horenmar:
Hmmm, I'll take a look -- it has honestly never occurred to me to run coverage over the top level ctest process and let it take coverage from the children processes...

sfranzen:
I found it via [this issue](https://github.com/OpenCppCoverage/OpenCppCoverage/issues/6) while googling about OpenCppCoverage. Oddly enough its documentation does not mention CMake or CTest at all, only the export + merge method that is currently used by Catch2.

horenmar:
Good news: I fixed the helper to pass through return codes, and will merge it in the evening.

Bad news: I also found out that the coverage merging fails in a way I am unable to reproduce locally, and will have to poke the CI to fix (second one this week, ugh)

What this means is that you should rebase the PR tomorrow/late evening today and I'll start reviewing the changes.

------------

I'll also open up an issue to investigate the different method of gathering coverage.

include/internal/catch_test_case_registry_impl.h
```diff
@@ -23,6 +23,8 @@ namespace Catch {
     struct IConfig;
 
     std::vector<TestCase> sortTests( IConfig const& config, std::vector<TestCase> const& unsortedTestCases );
+
+    bool isThrowSafe( TestCase const& testCase, IConfig const& config );
```
horenmar: I have two nitpicks with this function.

1) I don't like the fact that it is declared in two different headers. Does it need to be declared in both?
2) I don't like the name (but I currently don't have a better one either).

projects/CMakeLists.txt
```diff
@@ -392,8 +392,14 @@ set_tests_properties(ListTestNamesOnly PROPERTIES
...
-add_test(NAME NoTest COMMAND $<TARGET_FILE:SelfTest> -w NoTests "___nonexistent_test___")
-set_tests_properties(NoTest PROPERTIES PASS_REGULAR_EXPRESSION "No test cases matched")
+add_test(NAME NoTest COMMAND $<TARGET_FILE:SelfTest> Tracker, "___nonexistent_test___")
+set_tests_properties(NoTest PROPERTIES
+    PASS_REGULAR_EXPRESSION "No test cases matched '___nonexistent_test___'"
+    FAIL_REGULAR_EXPRESSION "No tests ran"
+)
+
+add_test(NAME WarnAboutNoTests COMMAND $<TARGET_FILE:SelfTest> -w NoTests "___nonexistent_test___")
+set_tests_properties(WarnAboutNoTests PROPERTIES WILL_FAIL TRUE)
```
horenmar: This should probably also have a `FAIL_REGULAR_EXPRESSION "Helper failed with"` so that if the underlying helper fails for some reason, the test does not pass.

include/internal/catch_test_spec.cpp
```diff
@@ -15,45 +16,81 @@
...
-        // All patterns in a filter must match for the filter to be a match
-        for( auto const& pattern : m_patterns ) {
-            if( !pattern->matches( testCase ) )
-                return false;
-        }
-        return true;
+        return std::all_of( m_patterns.begin(), m_patterns.end(), [&]( PatternPtr const& p ){ return p->matches( testCase ); } );
+    }
+
+    std::string TestSpec::Filter::name() const {
```
horenmar: The changes around name actually cause a regression, this is master:
```
$ ./projects/SelfTest [vector][fafa] -w NoTests
Filters: [vector][fafa]
No test cases matched '[vector][fafa]'
No tests ran
```

this is your branch
```
$ ./projects/SelfTest [vector][fafa]
Filters: [vector][fafa]
No test cases matched 'vector fafa'
No tests ran
```

include/internal/catch_session.cpp
```diff
@@ -58,46 +60,53 @@ namespace Catch {
...
+                auto const& allTestCases = getAllTestCasesSorted(*m_config);
+                m_matches = m_config->testSpec().matchesByFilter(allTestCases, *m_config);
+
+                if (m_matches.empty()) {
+                    for (auto const& test : allTestCases)
+                        if (!test.isHidden())
+                            m_tests.emplace(&test);
+                } else {
+                    for (auto const& match : m_matches)
+                        std::copy(match.tests.begin(), match.tests.end(), std::inserter(m_tests, m_tests.begin()));
```
horenmar: This seems to be an overly complicated variant of doing this:
```suggestion
                        m_tests.insert(match.tests.begin(), match.tests.end());
```

sfranzen: Entirely right, forgot about this possibility.

projects/CMakeLists.txt
sfranzen: Good idea, though apparently it should be a `PASS_REGULAR_EXPRESSION` to actually fail the test because a failure is already expected. I'll add that with a comment as it's confusing.

include/internal/catch_test_case_registry_impl.h
sfranzen: Hmm, I suppose it doesn't, but then wouldn't that also go for the other 3 functions declared in catch_test_case_registry.h? I added it just to reuse this part of the condition and figured it should have the same availability as `matchTest`.

Without changes to `#include`s in other files, my build succeeds with the declarations in a single location as follows:
1. catch_test_case_registry.h:
  Must have `isThrowSafe`, `filterTests`, `getAllTestCasesSorted`, may have `matchTest`;
2. catch_test_case_registry_impl.h:
  May have `matchTest`, doesn't need the other three.

So there are some locations that can only see the first header. What would be most appropriate? I'll wait before pushing another commit to resolve this.

horenmar: Hmmm, let's leave it be for now. I'll be doing a significant rewrite of those parts of the code for v3 anyway.

sfranzen: Alright. As for the name `isThrowSafe`, the only other idea that came to mind was something like `noIllegalThrows`, as `!(testCase.throws() && !config.allowThrows())` is a less readable equivalent of the condition. A slight downside is that it has a double negation, but that may be permissible since `!isThrowSafe` is not used.

I'm leaving it as is so you may decide whether to change it.

projects/CMakeLists.txt
```diff
@@ -400,7 +400,11 @@ set_tests_properties(NoTest PROPERTIES
 )
 
 add_test(NAME WarnAboutNoTests COMMAND $<TARGET_FILE:SelfTest> -w NoTests "___nonexistent_test___")
-set_tests_properties(WarnAboutNoTests PROPERTIES WILL_FAIL TRUE)
+set_tests_properties(WarnAboutNoTests PROPERTIES
+    WILL_FAIL TRUE
+    # Failure is now expected, so test errors must be indicated as passing it
+    PASS_REGULAR_EXPRESSION "Helper failed with"
```
horenmar: I had some doubts this would work so I tested it and it doesn't.

This is the CMakeLists.txt:
```cmake
cmake_minimum_required(VERSION 3.5)
project(runner-test LANGUAGES CXX)

add_executable(returns1 return1.cpp)

include(CTest)
add_test(NAME inverted COMMAND returns1)
set_tests_properties(inverted PROPERTIES
    WILL_FAIL TRUE
    PASS_REGULAR_EXPRESSION "Seeing this is an error"
)
```

this is the return1.cpp file:
```cpp
int main() {  return 0; }
```

and this is the result:
```
    Start 1: inverted
1/1 Test #1: inverted .........................   Passed    0.04 sec
```


As it turns out, the difference between the semantics of `PASS_REGULAR_EXPRESSION` and `FAIL_REGULAR_EXPRESSION` means that the test will now always pass as long as it does not write out the magic string, independently of the return value.

There are two solutions to this
1) Don't check for failure of the helper and hope for the best.
2) Break the logic out into its own script.

sfranzen: Huh, that's odd, but good that you checked, or it would have been a useless test! The documentation suggested (to me at least) that it would invert the interpretation of the return code as well as PASS/FAIL. It seems like this was [reported to CMake](https://gitlab.kitware.com/cmake/cmake/issues/16551) over 2 years ago and even had a [merge request](https://gitlab.kitware.com/cmake/cmake/merge_requests/416) with some discussion, but this was closed in favour of alternatives that were apparently never implemented.

Oh well. I've put the logic in a CMake script that is now used as the test command, this should work fine.

horenmar:
You will probably need to pass through the path to the test binary from the main script.

sfranzen:
I think you were right, here's to hoping this is the final change!

horenmar:
There is 1 more thing that needs fixing and one that I am not quite sure about.

Needs fixing: the filter string in `No test cases matched 'asdafs'` should roundtrip, so that it can be again used as a filter provided to Catch. Currently it does not:

```
$  ./projects/SelfTest [asdafs]
Filters: [asdafs]
No test cases matched 'asdafs'
No tests ran
```


The part I am not sure about is whether the extended reporting should be provided even if the `-w NoTests` flag was not specified.

sfranzen:
This was slightly trickier as I've now had to mess with the test spec parser, but it does the trick and passes the tests on my machine. About the reporting, I would prefer to always leave at least a textual trace, so the user can see where they mistyped or misconfigured test cases.

horenmar:
Okay, this looks good now.

Would you prefer if I squash-merge it, or if you rebase and cleanup the branch?

sfranzen:
I'm fine with squashing, it is just improvements/fixes and regression tests that don't warrant multiple commits IMO. Thanks for the careful review!

Issue #1685, opened 2019-07-10
std::byte should report as a number, not a character


studoot:
**Description**
Currently, `std::byte` is output as if it were a character, which it's not. This is similar to #1144, except that `std::byte` is defined as a scoped enumeration in the standard, so there shouldn't be any issue with it possibly being a typedef, unlike `gsl::byte`.

**Additional context**
This can be worked around in the usual way, by providing a custom `ostream` inserter or `StringMaker`, as in #1144.

I have a PR prepared that adds this.

PR #1686, opened 2019-07-10
Add `std::byte` support

studoot:
## Description
This addresses issue #1685, adding a `std::byte` specialisation for `Catch::Detail::StringMaker`, which makes `std::byte` print as an integer rather than as a character.

This has three parts:

1. **Add byte detection in Catch config**
    Following the pattern used for `std::optional`, look for the include file `cstddef` and the macro `CATCH_CPP17_OR_GREATER`. This defines `CATCH_INTERNAL_CONFIG_CPP17_BYTE`, which may then cause `CATCH_CONFIG_CPP17_BYTE` to be defined.

2. **Add a `std::byte`specialisation for Catch2::Detail::StringMaker**
    Added a declaration (gated by `CATCH_CONFIG_CPP17_BYTE`) of `struct StringMaker<std::byte>` in `catch_tostring.h`, and a definition in `catch_tostring.cpp`.

3. **Added a self-test (only compiled & executed for c++17)**
    Added new file `projects/SelfTest/UsageTests/ToStringByte.tests.cpp` and added it to the relevant place in `projects/CMakeLists.txt`

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -211,7 +211,6 @@ namespace Catch {
     };
 
 #if defined(CATCH_CONFIG_CPP17_BYTE)
-#include <cstddef>
     template<>
     struct StringMaker<std::byte> {
         static std::string convert(std::byte value);
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -210,6 +210,13 @@ namespace Catch {
         }
     };
 
+#if defined(CATCH_CONFIG_CPP17_BYTE)
+#include <cstddef>
+    template<>
+    struct StringMaker<std::byte> {
+        static std::string convert(std::byte value);
+    };
+#endif // defined(CATCH_CONFIG_CPP17_BYTE)
     template<>
     struct StringMaker<int> {
         static std::string convert(int value);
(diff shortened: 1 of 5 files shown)


include/internal/catch_tostring.h
```diff
@@ -210,6 +210,13 @@ namespace Catch {
         }
     };
 
+#if defined(CATCH_CONFIG_CPP17_BYTE)
+#include <cstddef>
```
horenmar: You are including a standard library header inside a namespace... this is not a good idea.

Because this file already uses `<cstddef>` by default, it is possible to just remove this. Alternatively, it should be included in a manner similar to `<string_view>` around line 20.

horenmar:
Apart from the small nitpick, this looks good and ready for merging 👍

include/internal/catch_tostring.h
studoot: Doh! I'll fix that...

studoot:
`#include <cstddef>` removed...

claremacrae:
I noticed that CATCH_CONFIG_CPP17_BYTE has been added, but as far as I can see, it is not included in the docs:

https://github.com/catchorg/Catch2/blob/master/docs/configuration.md#c17-toggles

Should it be?

studoot:
Yes, it probably  should be. CATCH_CONFIG_CPP17_OPTIONAL also appears to be missing from that documentation.

horenmar:
It has been added now.

PR #1697, opened 2019-07-21
Fix non-default-constructible type lists used in TEMPLATE_LIST_TEST_CASE

BenjaminW3:
std::tuple is not default constructible when the first type is not
default-constuctible. Therefore it can not be instantiated.
To circumvent this, we have to use std::declval in the unevaluate decltype
context.


horenmar:
Looks good, thanks.

PR #1706, opened 2019-07-29
Fix xml reporter crashing when a benchmark failed.

fodinabor:
## Description
When running a benchmark, that throws an exception or a optimized away exception is thrown, the xml reporter closes the wrong tag.
This was fixed by opening a BenchmarkResults tag on the preparing message. This way it is also reproducible, which benchmark failed.

### DIFF
--- a/include/reporters/catch_reporter_xml.cpp
+++ b/include/reporters/catch_reporter_xml.cpp
@@ -220,10 +220,13 @@ namespace Catch {
     }
 
 #if defined(CATCH_CONFIG_ENABLE_BENCHMARKING)
-    void XmlReporter::benchmarkStarting(BenchmarkInfo const &info) {
+    void XmlReporter::benchmarkPreparing(std::string const& name) {
         m_xml.startElement("BenchmarkResults")
-            .writeAttribute("name", info.name)
-            .writeAttribute("samples", info.samples)
+            .writeAttribute("name", name);
+    }
+
+    void XmlReporter::benchmarkStarting(BenchmarkInfo const &info) {
+        m_xml.writeAttribute("samples", info.samples)
             .writeAttribute("resamples", info.resamples)
             .writeAttribute("iterations", info.iterations)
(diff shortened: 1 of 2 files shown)


horenmar:
So did it crash or did it write out wrong xml?

fodinabor:
Both ;-)
It crashes because it tries to pop more tags than are opened..
```
<Catch name="test-gpgpuevaluation-performance.exe" filters="[polarization][copy]">
  <Group name="test-gpgpuevaluation-performance.exe">
    <TestCase name="CPU Memory Transfers - Types&lt;cl_uchar, cl_uchar3, cl_uchar4>" tags="[copy][cpu][polarization]" filename="D:\Dokumente\Uni\Bachelorarbeit\Project\test\performance\src\test_cpu_performance.cpp" line="241">
      <Section name="Given: CPU Provider" filename="D:\Dokumente\Uni\Bachelorarbeit\Project\test\performance\src\test_cpu_performance.cpp" line="247">
        <Info>
          work_group_size := 1
        </Info>
        <Section name="And given: real polarization image" filename="D:\Dokumente\Uni\Bachelorarbeit\Project\test\performance\src\test_cpu_performance.cpp" line="256">
          <Info>
            work_group_size := 1
          </Info>
          <Expression success="true" type="REQUIRE" filename="D:\Dokumente\Uni\Bachelorarbeit\Project\test\performance\src\test_cpu_performance.cpp" line="262">
            <Original>
              data
            </Original>
            <Expanded>
              0x00000262a4f2e040
            </Expanded>
          </Expression>
          <failed message="could not measure benchmark, maybe it was optimized away"/>
        </Section>
        <failed message="could not measure benchmark, maybe it was optimized away"/>
      </Section>
      <failed message="could not measure benchmark, maybe it was optimized away"/>
    </TestCase>
    <OverallResults successes="1" failures="0" expectedFailures="0"/>
  </Group>
  <OverallResults successes="2" failures="0" expectedFailures="0"/>
</Catch>
</>
`

horenmar:
Ok, I see.

PR #1709, opened 2019-07-30
Fix typo in the tutorial

aytekinar:
## Description
Changed `#include "catch.hpp"` to `#include <catch2/catch.hpp>`.

## GitHub Issues
N/A

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -44,7 +44,7 @@ To keep things simple we'll put everything in a single file (<a href="#scaling-u
 
 ```c++
 #define CATCH_CONFIG_MAIN  // This tells Catch to provide a main() - only do this in one cpp file
-#include "catch.hpp"
+#include <catch2/catch.hpp>
 
 unsigned int Factorial( unsigned int number ) {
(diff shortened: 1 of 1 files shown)


horenmar:
From the very same tutorial

> The rest of this tutorial will assume that the Catch2 single-include header (or the include folder) is available unqualified - but you may need to prefix it with a folder name if necessary.

aytekinar:
My bad, we can close this PR.

When I read that description you are quoting, I thought of having a directory structure similar to the below:
```
 ~/project> tree .
.
├── main.cpp
└── third-party
    └── catch2
        └── catch.hpp

2 directories, 2 files
```
where I assumed I would be simply copying `catch.hpp` (while preserving the directory structure) to a non-standard directory, which would result in my

1. either prepending the `catch2/catch.hpp` with the parent directory, or,
2. adding `-I` switch to the compiler options to point out to the non-standard directory.

Thanks for clarification, anyways.

horenmar:
Fair, but the tutorial is targeted at people just starting out with trivial structure, where the files are all next to each other 😃

PR #1710, opened 2019-07-31
Add Android implementation of writeToDebugConsole that prints to Logcat

michaelcowan:
## Description
Adds an implementation of `writeToDebugConsole` for Android that logs directly to Logcat.

In my opinion, this provides a really good out-of-the-box experience for Android NDK developers.

### Notes
I refrained from adding `CATCH_PLATFORM_ANDROID` and a `catch_android_h_proxy.h` as this felt a little heavy considering that the platform is really more Linux-like and the change is very focused. Happy to make any changes suggested though 😄

### DIFF
--- a/include/internal/catch_debug_console.cpp
+++ b/include/internal/catch_debug_console.cpp
@@ -20,7 +20,7 @@
         }
     }
 
-#elif CATCH_PLATFORM_WINDOWS
+#elif defined(CATCH_PLATFORM_WINDOWS)
 
     namespace Catch {
         void writeToDebugConsole( std::string const& text ) {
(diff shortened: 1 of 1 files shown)


michaelcowan:
The Travis job failed because of the [Python configuration job](https://travis-ci.org/catchorg/Catch2/jobs/565923116) which I think looks to be unrelated to this change.

The coverage decrease is annoying though. Not really sure how I can improve that without adding an Android test configuration.

horenmar:
If you rebase onto master, travis should work now. As for AppVeyor, you should use `elif defined(CATCH_PLATFORM_WINDOWS)`, because it does not have integral value.

horenmar:
Okay, thanks.

PR #1715, opened 2019-08-06
Fix infinite loop in compact reporter printer

sfranzen:
## Description
I was playing with different reporters and found an infinite loop when using the compact reporter with no further options. This is fixed by ensuring that the message iterator is always advanced in the `printRemainingMessages` function.


horenmar:
I believe I've ran into this one recently, but did not have time to dig in back then.

PR #1717, opened 2019-08-07
static constexpr: storage class first

ax3l:
## Description

Seen with PGI compiler for `static constexpr`:
```
warning: storage class is not first
```

## GitHub Issues

N/A


horenmar:
Sure, why not.

ax3l:
Thanks for merging, it's also the only location where the order was that way.

Issue #1347, opened 2018-07-26
JUnit failures lack detail


rcdailey:
Using the latest version of Catch2 on master (I think it's pretty close to v2.3), failures in JUnit XML output is very bare. You basically get this:

```
REQUIRE: "hello" == "world"
at /build/source/Core/UI/Test/Source/DataLoaders/TestCSVLoader.cpp:52
```

For the following code:

```
std::string dumbdata = "hello";
REQUIRE(dumbdata == "world");
```

I'd like to see a level of detail similar to console reporter, where it shows a lot more information, such as the verbatim parameter names passed into the assertion macro, the expanded form, etc.

horenmar:
Should be addressed by #1719

PR #1719, opened 2019-08-07
Improve JUnit failure report

AlexanderStein:
The JUnit report is improved in that:
* The message shows the testing condition, not the result
* The actual message has similar output than the console one

My reason is that using catch2 with junit reporter in a gitlab testing environment, failures report next to nothing, just the line where the failure happened. So i modified the output to be similar to the console one.
I think this also addresses #1347


horenmar:
Thanks, I rebased the changes to current master to make it ready for merging.

PR #1723, opened 2019-08-09
Rename template parameter names to avoid name conflict on Solaris.

### LINKED ISSUE #1722: Catch2 does not compile on Solaris.
Current master does not compile on Solaris due to a name conflict with a #define in the <sys/regset.h> header.

I was able to reproduce this bug in an isolated directory containing a main.cpp (below) and the single_include/catch2/catch.hpp header from commit e2d863b090f54b648147df1ffd084babc94a5e99. This is the simple main.cpp file that doesn't compile:

```cpp
#include "catch.hpp"
int main() {}
```

Compilation command: **g++ -std=c++11 -I. main.cpp**

This bug originated in commit 004228efb2ccbd2f69eb443933507012fa82304c where the template parameter name "SS" was used. The conflicting define in the Solaris <sys/regset.h> header is this:

```cpp
#define SS              18      /* only stored on a privilege transition */
```

Naturally, g++ errors out when it encounters a numeric constant instead of a template parameter name: 

```
catch.hpp:1542:31: error: expected nested-name-specifier before **numeric constant**
template<typename SS, typename TT>
                   ^
```

Platform Info:
OS: Solaris 11.4.0.15.0 i386
Compiler: GCC 7.3.0
Catch Version: e2d863b090f54b648147df1ffd084babc94a5e99

A PR will be created to fix this issue.

jealas:
Closes #1722

This PR fixes a compilation issue on Solaris 11.4 with GCC 7.3.0 on the latest Catch2 master branch. A #define in the Solaris 11.4 <sys/regset.h> header conflicts with a template parameter name used in catch_tostring.h and it gets renamed to an integer constant. This causes an obvious compilation issue. Please refer to the issue for more details about the bug.

The bug is simply fixed by re-naming the 'SS' template parameter name to something more descriptive like 'Stream'. I also took the liberty of renaming the template parameter name 'TT' to 'U' because I saw the template parameter name 'U' used throughout the codebase and couldn't find any other "same double letter" names used.

After making this change, I can confirm that Catch2 compiles and builds with the main.cpp file detailed in the issue, with a non-trivial project and with the Catch2 SelfTest target on Solaris 11.4.

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -44,9 +44,9 @@ namespace Catch {
 
         template<typename T>
         class IsStreamInsertable {
-            template<typename SS, typename TT>
+            template<typename Stream, typename U>
             static auto test(int)
-                -> decltype(std::declval<SS&>() << std::declval<TT>(), std::true_type());
+                -> decltype(std::declval<Stream&>() << std::declval<U>(), std::true_type());
 
             template<typename, typename>
             static auto test(...)->std::false_type;


jealas:
SelfTest ran successfully on Solaris 11.4 with: **All tests passed (1395 assertions in 219 test cases)**.

horenmar:
Yet another example of why the C macros were a terrible idea.

Thanks for the fix.

PR #1726, opened 2019-08-12
Small updates to include directives

matt77hias:
## Description
Updated some include directives to improve consistency and add more flexibility when integrating Catch2 with regard to include directories (e.g., allows setting the root as include directory).


horenmar:
Thanks.

PR #1729, opened 2019-08-17
Allow to use non-copyable and non-movable types in TEMPLATE_LIST_TEST_CASE

BenjaminW3:
This is a follow-up to #1697.

The parameter given to `convert` may not be copyable therefore it has to be
captured by const reference. For example an `std::tuple` that contains a
non-copyable type is itself non-copyable.

The NonDefaultConstructible test-case was reduced by one example type
because it did not add any value.

### DIFF
--- a/include/internal/catch_preprocessor.hpp
+++ b/include/internal/catch_preprocessor.hpp
@@ -118,7 +118,7 @@
     template<template <typename...> class Final, template< typename...> class...Containers, typename...Types>\
     constexpr auto create(TypeList<Types...>) noexcept -> decltype(append(Final<>{}, rewrap<Containers>(Types{}...)...)) { return {}; }\
     template<template <typename...> class Final, template <typename...> class List, typename...Ts>\
-    constexpr auto convert(List<Ts...>) noexcept -> decltype(append(Final<>{},TypeList<Ts>{}...)) { return {}; }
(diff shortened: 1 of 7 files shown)


BenjaminW3:
Independent from my changes CI seems to be broken:
```
0.78s$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install clang-6.0 libstdc++-8-dev
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 clang-6.0 : Depends: libjsoncpp0 (>= 0.6.0~rc2) but it is not installable
E: Unable to correct problems, you have held broken packages.
apt-get.diagnostics
apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install clang-6.0 libstdc++-8-dev" failed and exited with 100 during .
Your build has been stopped.
```

horenmar:
Thanks, the CI should be fixed soon(ish) and I'll merge this afterwards.

PR #1733, opened 2019-08-22
Enable variadic macros support for MSVS 2005 as it has them

vadz:
Fix the test for MSVC version: variadic macros support is available
since MSVS 2005, a.k.a. MSVC 8, using _MSC_VER=1400.

---

I know nobody cares about this ancient compiler any more, but unfortunately I still have to build some projects using it, and I'd still like to use CATCH with it and it does support variadic macros.


horenmar:
Not gonna lie, I feel sorry for you.

Issue #1230, opened 2018-03-21
Warning from Clang Static Analyzer about returning null reference


nlohmann:
## Description

I ran Clang Static Analyzer over [nlohmann/json](https://github.com/nlohmann/json/tree/feature/catch2) which uses Catch 2.2.1 in a feature branch. The analyzer reported a warning in Catch's source code:

```
/Users/niels/Documents/repositories/json/test/thirdparty/catch/catch.hpp:4620:9: warning: Returning null reference
        return *IMutableContext::currentContext;
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/niels/Documents/repositories/json/test/thirdparty/catch/catch.hpp:9341:17: warning: Potential leak of memory pointed to by 'filename.m_start'
                CATCH_ENFORCE( !m_ofs.fail(), "Unable to open file: '" << filename << "'" );
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/niels/Documents/repositories/json/test/thirdparty/catch/catch.hpp:6620:56: note: expanded from macro 'CATCH_ENFORCE'
    do{ if( !(condition) ) CATCH_ERROR( msg ); } while(false)
                                                       ^~~~~
```

I attached a ZIP file of the HTML report where the analyzer's trace can be followed: [report.zip](https://github.com/catchorg/Catch2/files/1835392/report.zip)

This may be a false positive, but I though this about all warnings of Clang Static Analyzer that eventually turned out to be real issues :-)

### Steps to reproduce

- check out the [`feature/catch2` branch of `nlohmann/json`](https://github.com/nlohmann/json/tree/feature/catch2)
- run `make clang_analyze`

(possibly adjust the path to Clang Static Analyzer).

### Extra information

* Catch version: **v2.2.1**
* Operating System: **macOS High Sierra 10.13.3 (17D102)**
* Compiler+version: **clang version 5.0.0 (trunk 292523)**

horenmar:
The second warning is definitely bogus, but I'll take a look at the null deref.

Btw, looking through the changes, you might be interested in `REQUIRE_THROWS_MATCHES(expr, exception-type, matcher)`.

horenmar:
Finally got around this and it is definitely a false positive.

[...]

PR #1735, opened 2019-08-24
Suppress false positive from clang-analyzer

grimmn:
Fixes issue #1230

## Description
Suppress a false positive warning from the Clang static analysis tools which thought that a null reference was potentially being returned here.


horenmar:
Thanks, sorry for taking so long.

Issue #1737, opened 2019-08-26
Escaping special characters in test name no longer works in v2.9.2


JohnnyHendriks:
**Describe the bug**
When calling a test via the command line by its full name, and this name contains a special character such as `,`, `[`, `\` or `"`, it used to be possible to escape these characters and run the specific test. This no longer seems to work as of v2.9.2 of Catch2. Works just fine in all previous Catch2 versions. The [Test Adapter for Catch2](https://github.com/JohnnyHendriks/TestAdapter_Catch2) extension for Visual Studio relies on this functionality to call individual tests.

**Expected behavior**
It should be possible to call specific tests with special characters in their name via the command line by escaping the special characters, as was the case in all versions of Catch2 prior to v2.9.2.

**Reproduction steps**
Create a test with a name that contains a special character, _e.g._ "Test, special". Call the test by its full name via the commandline. Make sure you make use of Catch2 v2.9.2

```
Test.exe "Test\, special"
```
This will result in an `invalid string position` error.

The same test with any of the previous version of Catch2 are able to run the test this way.

**Platform information:**
 - OS: Windows 10
 - Compiler+version: Visual Studio 2019
 - Catch version: v2.9.2

amitherman95:
I'm looking into it

horenmar:
This should now work

JoeyGrajciar:
@JohnnyHendriks was the issue resolved for you?
I am going to close this for now. If it is not resolved please reopen it.

JohnnyHendriks:
Thank you. I checked and it looks like the issue is indeed resolved.

PR #1752, opened 2019-09-14
Escape characters bug patch(v2.9.2)

amitherman95:
## Description
In Catch v2.9.2 escape characters are not working properly
I modified the finite state machine implemented in include/catch_test_spec_parser.cpp
to correct this problem

## GitHub Issues
Issue #1737


horenmar:
Thanks for the PR. I added some tests and will probably merge it right after CI is done.

amitherman95:
What about the single include file? was it updated too?

horenmar:
That will happen during next release.

PR #1758, opened 2019-09-21
Refactor CMake for more modern usage

slurps-mad-rips:
## Description

Hello! I've run into a few small issues with Catch2 that prevent me from `FetchContent` downloading it directly, and just using `add_subdirectory`. This PR also fixes a few issues regarding file install locations, CPack source packages, and removes an unnecessary `if()` when compiling for Android

Use `cmake_dependent_option` to disable a few options if it's being used as a subdirectory

Reorganize file so that properties, includes, options are all at the top

Catch2 will now link to Android's log library via a generator expression.

Left a note if the minimum version for Catch2 is ever 3.14

Simplified install locations

`CATCH_INSTALL_HELPERS` will now use pattern matching for files instead of listing them out each time.

Building a CPack source package will no longer include a build, .git, or other "garbage" files

### DIFF
--- a/CMake/Catch2Config.cmake.in
+++ b/CMake/Catch2Config.cmake.in
@@ -1,10 +1,8 @@
 @PACKAGE_INIT@
 
-
 # Avoid repeatedly including the targets
 if(NOT TARGET Catch2::Catch2)
     # Provide path for scripts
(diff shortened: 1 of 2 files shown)


horenmar:
With no followup I am going to drop this.

slurps-mad-rips:
I'll be breaking this up into smaller PRs over the next few months, so no worries :)

PR #1764, opened 2019-10-04
update name of Value Generators in doc to match class names

WimLeflere:
update name of Value Generators in doc to match class names


horenmar:
Thanks

Issue #1767, opened 2019-10-07
Cannot pass a test name with special characters via an input file


igor-akhmetov:
I can't seem to pass the name of a test with special characters via an input file, specifically the one from CmdLine.tests.cpp:
```
TEST_CASE("Test with special, characters \"in name", "[cli][regression]") {
    // This test case succeeds if we can invoke it from the CLI
    SUCCEED();
}
```
`--list-test-names-only` outputs `Test with special, characters "in name`, which does not allow to run the test:
```
C:\src\Catch2\build\projects\Debug>SelfTest.exe -f tests.txt
Filters: "Test with special, characters "in name",
No test cases matched '"Test with special'
No test cases matched 'characters "in name"'
===============================================================================
No tests ran
```
I'd expect
```
Test with special\, characters \"in name
```
or
```
"Test with special\, characters \"in name"
```
to work, but they don't:
```
C:\src\Catch2\build\projects\Debug>SelfTest.exe -f tests.txt
Filters: "Test with special\, characters \"in name",
No test cases matched '"Test with special\, characters \"in name"'
===============================================================================
No tests ran


C:\src\Catch2\build\projects\Debug>SelfTest.exe -f tests.txt
Filters: "Test with special\, characters \"in name",
No test cases matched '"Test with special\, characters \"in name"'
===============================================================================
No tests ran
```

amitherman95:
Try updating the single header file by executing <Catch2 path>/scripts/generateSingleHeader.py

igor-akhmetov:
Same with names like `vector<int,allocator> -> toString`, escaping a comma in the input file does not seem to help.

igor-akhmetov:
@amitherman95 How is the single header relevant here? This is with the latest version of Catch2, I just grab the latest source, build, and run SelfTest.exe.

amitherman95:
Oh I see, I'll check it out

echuber2:
I'm also experiencing this. Commas in test case names break stuff. This used to work in older versions of Catch (at least, it worked in v2.2.3). Now, with v2.9.2, it's causing issues. I do note that when I search through the v.2.9.2 catch.hpp file, there is one part that says:

> // Applies the function macro `f` to each of the remaining parameters, inserts commas between the results,
// and passes userdata as the first parameter to each invocation,

[...]

PR #1769, opened 2019-10-08
Patch:Failure to pass special chars through input file

amitherman95:
## Description
When Catch parses quoted test name, it may insert wrong positions of escape characters into
TestSpecParser::m_escapeChars, because the quotes are not added to the pattern itself
I added a simple data member to TestSpecParser that will track the right location of those characters in the pattern.

## GitHub Issues
This PR resolves Issue #1767


horenmar:
Thanks, I fixed up the branch history and added tests for the changes.

Issue #1766, opened 2019-10-07
Catch2 terminates when an invalid test name is passed


igor-akhmetov:
Put the following test name in a file, say tests.in:
```
Test with special, characters \"in name
```
Run a Catch binary with `-f tests.in`. The program terminates with a runtime error thrown from
```
 	msvcp140d.dll!00007ffc846a4ef9()	Unknown
>	SelfTest.exe!std::_String_val<std::_Simple_types<char>>::_Xran() Line 2141	C++
 	SelfTest.exe!std::_String_val<std::_Simple_types<char>>::_Check_offset(const unsigned __int64 _Off) Line 2131	C++
 	SelfTest.exe!std::string::assign(const std::string & _Right, const unsigned __int64 _Roff, unsigned __int64 _Count) Line 2824	C++
 	SelfTest.exe!std::string::basic_string<char,std::char_traits<char>,std::allocator<char>>(const std::string & _Right, const unsigned __int64 _Roff, const unsigned __int64 _Count, const std::allocator<char> & _Al) Line 2278	C++
 	SelfTest.exe!std::string::substr(const unsigned __int64 _Off, const unsigned __int64 _Count) Line 3949	C++
 	SelfTest.exe!Catch::TestSpecParser::addPattern<Catch::TestSpec::NamePattern>() Line 58	C++
 	SelfTest.exe!Catch::TestSpecParser::endMode() Line 108	C++
 	SelfTest.exe!Catch::TestSpecParser::visitChar(char c) Line 38	C++
 	SelfTest.exe!Catch::TestSpecParser::parse(const std::string & arg) Line 22	C++
 	SelfTest.exe!Catch::Config::Config(const Catch::ConfigData & data) Line 34	C++
```
I'd expect Catch to ignore the invalid test name.

horenmar:
@amitherman95 You've worked on the parser recently, can you take a look at this?

amitherman95:
Yes I will

amitherman95:
Can the input file also include tags? or test names only?

horenmar:
Let's say test names only for now

amitherman95:
Very well, I pretty much solved the problem, I just need to add more robust tests for this feature and by tomorrow a PR will be ready.

PR #1770, opened 2019-10-10
Patch: Catch terminates when parsing invalid test name from input file

amitherman95:
## Description
Catch 2.9.2 lacks the means to detect invalid test names and ignore them when needed.

## GitHub Issues
This PR resolves Issue #1766

### DIFF
--- a/include/internal/catch_test_spec_parser.cpp
+++ b/include/internal/catch_test_spec_parser.cpp
@@ -20,8 +20,13 @@ namespace Catch {
         m_substring.reserve(m_arg.size());
         m_patternName.reserve(m_arg.size());
         m_realPatternPos = 0;
+        
         for( m_pos = 0; m_pos < m_arg.size(); ++m_pos )
-            visitChar( m_arg[m_pos] );
+          //if visitChar fails
+           if( !visitChar( m_arg[m_pos] ) ){ 
+               m_testSpec.m_invalidArgs.push_back(arg);
+               break;
+           }
         endMode();
         return *this;
     }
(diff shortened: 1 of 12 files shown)


horenmar:
I played around with this a bit, and I have 2 notes:
1) Silently ignoring a test spec is probably not a good idea, so when the parser decides to 
2) I am not sure that all of these patterns are actually incorrect (part of this is probably that the input format is not 100% specified). For example I wouldn't expect this to happen:

```
./projects/SelfTest "\"asdf,ffss\""

SelfTest is a Catch v2.9.2 host application.
Run with -? for options

... snip ...

All tests passed (1457 assertions in 227 test cases)
```

The "faulty" input is removed and suddenly all non-hidden tests are run.

amitherman95:
Thanks for the feedback, I'll work on it

amitherman95:
Done

amitherman95:
Hmm that's unfortunate

include/internal/catch_test_spec_parser.cpp
```diff
@@ -20,8 +20,12 @@ namespace Catch {
         m_substring.reserve(m_arg.size());
         m_patternName.reserve(m_arg.size());
         m_realPatternPos = 0;
-        for( m_pos = 0; m_pos < m_arg.size(); ++m_pos )
-            visitChar( m_arg[m_pos] );
+        try{
+            for( m_pos = 0; m_pos < m_arg.size(); ++m_pos )
+              visitChar( m_arg[m_pos] );
+        }catch(...){
```
JoeyGrajciar: technically here you should use `CATCH_TRY` and `CATCH_CATCH_ALL` to avoid problems with disabled exceptions.
But with disabled exceptions `CATCH_CATCH_ALL` expands to `if ((false))` so that would not behave as you want.
Would it be possible to change `visitchar` to return something what would indicate error?

amitherman95: Yes, I thought about it, perhaps this is what I'll do

JoeyGrajciar:
Tried your changes locally and it look fine except for two things:
```
Filters: "Test with special, characters \"in name" , "T*" ,
All tests passed (1 assertion in 1 test case)
```
* the extra comma after last filter
* I do not see that the first one was actually ignored (adding something like `Ignored Filters:` would be nice)

amitherman95:
Whoops, I missed some files in the commit that address exactly what you mentioned

JoeyGrajciar:
i still see the last comma in the filters :(
```
Filters: "Test with special, characters \"in name" , "T*" ,
Invalid Filter: "Test with special, characters \"in name"
All tests passed (1 assertion in 1 test case)
```

amitherman95:
I'll look into it

PR #1776, opened 2019-10-16
Enable floating point ranges

mvd4:
## Description
The generator helper `range(start, end, step)` currently only supports integer ranges.
This seems to be an artificial limitation, as the underlying `RangeGenerator<T>` works fine also for floating point types.

This PR changes the `static_assert` that limits the types for `range(start, end, step)` to also support floating point types. It also introduces additional tests to cover the extended functionality.


horenmar:
Actually it does not support floating point numbers intentionally. The original PR had no limits on the generator, but I wanted to avoid bug reports from people who'd use the generator for floating point numbers, without taking proper care of edge conditions caused by floating point rounding.

This is incidentally also why the static assert is placed in the helper function and not in the generator class itself -- it allows advanced users to create the generator manually if needed.

-------

I'll think about lifting the restriction.

PR #1779, opened 2019-10-17
add a note how to run selftests using multiconfig generators

arnemertz:
## Description
Going through the documentation to get started with some bug fixes, I noticed that with using Visual Studio as CMake generator leads to errors when trying to run ctest exactly as described in the documentation.


horenmar:
We could probably recommend just passing `-c Debug` for all generators -- Make generator happily ignores it.

arnemertz:
I just looked into the `-C` argument for CTest - sadly the documentation does not specify whether it's _always_ ignored by single configuration generators.

PR #1783, opened 2019-10-21
docs: command-line: Add example to specify tests by file name.

textshell:
## Description
I've been searching for a way to run all tests from a file for some time. To avoid other users having to connect the dots of tags and the `-#` option add a simple example to the "Specifying which tests to run" section.


horenmar:
Looks reasonable, thanks.

Issue #1759, opened 2019-09-23
CHECK at TEST_CASE level ignores logging?


sd-x:
Given this program:

```
#define CATCH_CONFIG_MAIN
#include "catch2/catch.hpp"

int x = 0;

TEST_CASE("my case", "[foo]") {
  auto y = 1-x;
  CAPTURE(y);
  SECTION("my section") {
    // CHECKfalse);  // If uncomment this, I see the CAPTURE message for it
  }
  CHECK(false); // I expect the message for this line to show 'y := 1', but it's missing
}
```

I do not see the
```
with message:
  y := 1
```
in the output (when run default options).  INFO behaves similarly.

my g++ is 7.3.0 and my Catch2 is v2.9.2

amitherman95:
I'm on it

Ext3h:
`RunContext::sectionEnded(SectionEndInfo const & endInfo)` appears to behave faulty. It is clearing *both* `m_messageScopes` (un-scoped messages) and `m_messages` (messages wich scope is strictly handled by RAII style scope objects).

Clearing `m_messageScopes` is in line with the documentation. Clearing `m_messages` goes against it.

There is also a second, similar bug in `RunContext::popScopedMessage(MessageInfo const & message)`, if a message with an identical text has been also added by an inner scope, both the inner *and* outer captures are removed. `RunContext::popScopedMessage(MessageInfo const & message)` should only remove a single instance of the message in this case, searching from the *top* of the scoped message stack.

PR #1785, opened 2019-10-23
Fixes 1759: Messages in Run Context removed after section ends

amitherman95:
## Description
When we use Capture, it inserts the variable name and value into RunContext::m_messages
When we apply an assertion, it copies the messages into its assertion results object to be printed later
by one of the reporters.

However, when we destroy a Catch::Section it calls RunContext::sectionEnded and this clears RunContext::m_messages and thus every assertion after the section will not copy anything from m_messages as it has been cleared.

I don't understand what was the purpose of clearing the messages in the first place

## GitHub Issues
This PR resolves Issue #1759

### DIFF
--- a/include/internal/catch_run_context.cpp
+++ b/include/internal/catch_run_context.cpp
@@ -217,8 +217,6 @@ namespace Catch {
         }
 
         m_reporter->sectionEnded(SectionStats(endInfo.sectionInfo, assertions, endInfo.durationInSeconds, missingAssertions));
-        m_messages.clear();
-        m_messageScopes.clear();
     }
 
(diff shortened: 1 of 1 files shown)


JoeyGrajciar:
I think with your changes even captures from inside section will survive the section end and that is probably something we do not want to have.

amitherman95:
I just checked it, any capture doesn't survive outside the section

torarnv:
Isn't this the same of `INFO` vs `UNSCOPED_INFO`? Perhaps we need an `UNSCOPED_CAPTURE`?

torarnv:
Sorry, I just saw the test case in the bug report. The capture/info should survive a section begin/end like that, that's what I'd expect at least.

PR #1794, opened 2019-11-01
Fix error when period of steady_clock is not nano

cericks0n:
## Description

On systems where std::chrono::steady_clock::period is not std::nano, benchmark tests fail to compile due to trying to convert analysis.samples from a vector of duration<double, clock::period> to a vector of std::chrono::duration<double, std::nano>.

To fix this, we instead construct a BenchmarkStats that matches the period of the clock. Once it's constructed, BenchmarkStats's conversion operator will handle converting to other durations.


horenmar:
Thanks. Out of curiosity, what platform did you run into this on?

cericks0n:
I hit this on PS4.

PR #1802, opened 2019-11-11
Eliminate copying of TestCaseInfo and reduce allocations during test case registration

horenmar:
## Description

This PR refactors handling of `TEST_CASE`s so that the test case name and tags are not copied around all the time. The tag handling is also refactored to decrease the total number of allocations when registering test cases.


horenmar:
Assuming I used `massif` right, doing `SelfTest --list-tests` went from 929 allocations (binary built from the `de-v3` branch), to 614 allocations on this branch.

horenmar:
FF-Merged to `dev-3` outside of github.

Issue #1800, opened 2019-11-07
storage_for missing from single include


Tridacnid:
**Describe the bug**
single_include/catch2/catch.hpp appears to be missing some benchmarking symbols that appear in the [docs](https://github.com/catchorg/Catch2/blob/master/docs/benchmarks.md#constructors-and-destructors). Others may also be missing.

**Expected behavior**
Compiling the example code should work:
```
TEST_CASE("BLAH")
{
    BENCHMARK_ADVANCED("construct")(Catch::Benchmark::Chronometer meter)
    {
        std::vector<Catch::Benchmark::storage_for<std::string>> storage(meter.runs());
        meter.measure([&](int i) { storage[i].construct("thing"); });
   }
}
```

**Observed behavior**
`error: storage_for is not a  member of Catch::Benchmark`

**Reproduction steps**
Compile the following:
```
#define CATCH_CONFIG_MAIN
#define CATCH_CONFIG_ENABLE_BENCHMARKING

#include <catch2/catch.hpp>

#include <vector>
#include <string>

TEST_CASE("BLAH")
{
    BENCHMARK_ADVANCED("construct")(Catch::Benchmark::Chronometer meter)
    {
        std::vector<Catch::Benchmark::storage_for<std::string>> storage(meter.runs());
        meter.measure([&](int i) { storage[i].construct("thing"); });
   }
}
```

**Platform information:**
 - Catch version: **v2.10.2**

Tridacnid:
Looks like include/internal/benchmark/catch_constructor.hpp isn't ever referenced in the source code, so the generateSingleHeader.py script doesn't include it in the final product. I'm going to try to submit a patch for both the missing inclusion as well as add a warning to the script so that in the future this sort of problem will be more visible to the developer.

horenmar:
Yeah, this sort of missing functionality has happened couple of times before.

Improvements to the stitching script are always welcome, but for the benchmarking specifically I would suggest creating file named something like `catch_benchmarking_all.hpp` inside the benchmarking folder, and including that for stitching. I suspect that our users will want such a header for the v3 version where we intend to provide Catch2 as a regular library made from multiple headers/TUs.

Tridacnid:
Thanks! I'll try to put together a patch to warn about this sort of thing when running the script. Is it intentional that SelfTest/UsageTests/Benchmark.test.cpp doesn't always compile with CATCH_CONFIG_ENABLE_BENCHMARKING?

PR #1808, opened 2019-11-19
Update single header generation script to warn about unused headers

Tridacnid:
## Description
This PR addresses a couple of things from #1800 
1. Add a warning to generateSingleHeaders.py to indicate that a header wasn't parsed as part of generating the single header. This would have alerted devs to the fact that catch_constructor.hpp was not part of the single header.
2. Correct docs/benchmarks.md usage of BENCHMARK_ADVANCED for the "construct" and "destroy" examples. The original documentation would not have compiled as written.
3. Correct the namespacing on helpers and constructs in catch_constructor.hpp. The docs indicated that these were intended to be found in the Catch::Benchmark namespace.

## GitHub Issues
Motivated by #1800 even though it is already closed.

### DIFF
--- a/docs/benchmarks.md
+++ b/docs/benchmarks.md
@@ -168,7 +168,7 @@ Note that it is not possible to simply use the same instance for different runs
 and resetting it between each run since that would pollute the measurements with
 the resetting code.
 
-It is also possible to just provide an argument name to the simple `BENCHMARK` macro to get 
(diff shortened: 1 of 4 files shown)


horenmar:
Thanks for the changes, but I'll have to fix them up to work on both Linux and Windows -- right now the missing header check only works on Linux.

PR #1815, opened 2019-12-15
Fix issue #1809.

### LINKED ISSUE #1809: Can't compile SingleValueGenerator inside GENERATE_COPY
**Describe the bug**
Attempting to invoke the `SingleValueGenerator` (using external variables) inside `GENERATE_COPY` like so...
```C++
int a = 1;
int b = 2;
int c = GENERATE_COPY( a,b );
```
results in a compilation error:
```C++
error: no matching function for call to 'Catch::Generators::GeneratorWrapper<const int&>::GeneratorWrapper(Catch::Generators::GeneratorWrapper<const int>)'
  { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }
```

**Expected behavior**
The above example code should compile fine, since `GENERATE_COPY` should allow the use of external variables (`a` and `b`) in any generator. For example, the following code compiles totally fine (using `RangeGenerator`):
```C++
int a = 1;
int b = 2;
int c = GENERATE_COPY( range(a,b) );
```

**Reproduction steps**
self-contained code **`bugdemo.cpp`:**
```C++
#include "catch.hpp"

TEST_CASE( "bug " ) {
    
    int a = 1;
    int b = 2;
    int c = GENERATE_COPY(a, b);
    REQUIRE( true );
}
```
and compiled simply with
```
g++ -std=c++14 -c bugdemo.cpp
```

**Platform information:**

 - OS: **MacOS 10.14** and **Ubuntu 18.04**
 - Compiler+version: **GCC v8.2.0** and **GCC v7.4.0** and **clang v10.0.0** (and all other versions on system)
 - Catch version: **v2.10.0**

Tridacnid:
Fix SingleValueGenerator CTor clash. Correct some misuses of move when forward was needed. Correct some missing universal references in method parameter lists. Add IntrospectiveTest to verify fix for issue #1809.

## Description
SingleValueGenerator had 2 clashing CTors thanks to the semantic differences in universal references and rvalue references when used in templates as parameters. 

## GitHub Issues
Closes #1809

### DIFF
--- a/include/internal/catch_generators.hpp
+++ b/include/internal/catch_generators.hpp
@@ -57,7 +57,6 @@ namespace Generators {
     class SingleValueGenerator final : public IGenerator<T> {
         T m_value;
     public:
-        SingleValueGenerator(T const& value) : m_value( value ) {}
         SingleValueGenerator(T&& value) : m_value(std::move(value)) {}
 
         T const& get() const override {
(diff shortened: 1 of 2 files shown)


Tridacnid:
Did most of my work in Godbolt, hopefully all the CI passes.
https://godbolt.org/z/AL89HN

Tridacnid:
Looks like Travis is having some issues downloading packages.

horenmar:
Thanks, I moved the tests around a little bit.

Issue #1830, opened 2020-01-13
v3.0.0 Link order


offa:
**Describe the bug**
Does the link order of _Catch2_ / _Catch2Main_ matter? Linking `Catch2Main Catch2` works, but using the opposite order results in undefined references:

```
[…]/lib/libCatch2Main.a(catch_main.cpp.o): in function `main':
catch_main.cpp:(.text.startup+0x34): undefined reference to `Catch::Session::Session()'
/usr/bin/ld: catch_main.cpp:(.text.startup+0x46): undefined reference to `Catch::Session::~Session()'
/usr/bin/ld: catch_main.cpp:(.text.startup+0x79): undefined reference to `Catch::Session::applyCommandLine(int, char const* const*)'
/usr/bin/ld: catch_main.cpp:(.text.startup+0x88): undefined reference to `Catch::Session::run()'
```

**Expected behavior**
Either _(a)_ this is intentional or _(b)_ a should be fixed (both are OK to me).

**Reproduction steps**

1. Build _v3.0.0-preview1_ as a static library (eg. using CMake)
1. Compile minimal program (see below) and change order of Catch2 / Catch2Main

```cpp
// minimal program
#define CATCH_CONFIG_MAIN
#include <catch2/catch.hpp>
```

```sh
# Fails (undefined references):
$ g++ 000-CatchMain.cpp -I[…]/include -L[…]/lib -lCatch2 -lCatch2Main

# Ok
$ g++ 000-CatchMain.cpp -I[…]/include -L[…]/lib -lCatch2Main -lCatch2
```

**Platform information:**

 - OS: **Linux**
 - Compiler+version: **GCC v9.2.0**
 - Catch version: **v3.0.0-preview1**

**Additional context**
#1823

maybug:
I met this

horenmar:
Right, I forgot about people who don't use CMake. If you use the CMake targets, linking just `Catch2Main` should work, because it declares dependency on `Catch2`.

I'll have to think about whether I want the two libraries to be link-order dependent, or whether they should both include all the required symbols.

horenmar:
I am going to go with the option of keeping the libraries as they are, and documenting that if you don't use CMake-provided integration, you have to link them both, and in the correct order (`Catch2Main` and then `Catch2`).

offa:
Thanks!

Issue #1823, opened 2019-12-30
v3.0.0 preview available through conan?


offa:
Will the v3.0.0 preview version become available through conan?

horenmar:
I intend to make v3 available through conan, but from the tasks in backlog it has a very low priority, similarly to providing single header version.

If you have experience with conan, you can contribute the required changes to speed it up.

offa:
What's still open respectively where can I help? I'm not an expert but a casual Conan user, maybe I can give you a hand.

horenmar:
I think there are 2 things that need work for v3 to show up on Conan:

1) Catch's Conan recipe likely needs to be updated to properly use the static library build in v3
2) The Conan publishing script needs to be updated to publish to `testing` (or whatever the usual non-stable name is) under some circumstances (currently it only publishes to `stable` when a stable tag is made).

horenmar:
Catch2 is also included in Conan-center by the Conan folks, so you might also try asking them to provide a recipe for v3.

offa:
I'm currently working on this.

offa:
Conan center ships only stable releases, so I have updated the conanfile for v3.

horenmar:
Bintray should now have v3.0.0 preview uploaded
https://bintray.com/catchorg/Catch2/catch2%3Acatchorg

gegles:
Now that bintray is gone, could we get it through another conan remote?

Much appreciated!

horenmar:
@gegles Yes, if someone takes care of it.

I don't use Conan anymore, nor have an idea what is a good free remote. However, if someone sets up something like existed for bintray in a GitHub action, I am fine with merging it again.

PR #1832, opened 2020-01-15
Conanfile updated for v3

offa:
## Description

Updated conanfile for v3. It uses the static library build, not the single header.

_:question: Is the target branch (`dev-v3`) the right one for *v3* contributions?_

## GitHub Issues
#1823, #1830

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,18 +10,28 @@ class CatchConan(ConanFile):
     homepage = url
     license = "BSL-1.0"
     exports = "LICENSE.txt"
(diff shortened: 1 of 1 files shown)


Minimonium:
Looks great! I guess we also could remove the `header-only` from topics, since it's a library now. 🙂

horenmar:
Yeah, until v3 becomes the main version, changes to it should target the `dev-v3` branch.

offa:
> Looks great! I guess we also could remove the `header-only` from topics, since it's a library now. slightly_smiling_face

Yeah, that makes sense.

PR #1837, opened 2020-01-21
Removed unneeded 'using uchar = unsigned char'

dtgriscom:
## Description
Catch2 was conflicting with my code that had a "#define uchar unsigned char" in its code. I found that Catch2's use of "uchar" was extremely limited (one "using uchar = unsigned char", two subsequent uses of "uchar"). Under the principle that Catch2 should be as light on the land as possible, I edited the three lines involved, re-ran the Catch2 SelfTests (it not surprisingly passed) and created this PR.

## GitHub Issues
None

### DIFF
--- a/include/internal/catch_xmlwriter.cpp
+++ b/include/internal/catch_xmlwriter.cpp
@@ -12,8 +12,6 @@
 #include <iomanip>
 #include <type_traits>
 
-using uchar = unsigned char;
-
 namespace Catch {
 
 namespace {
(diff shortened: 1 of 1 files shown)


dtgriscom:
Five consecutive Travis checks failed, all with the message "Error: retrieving gpg key timed out". I'm guessing these errors were collateral damage from some other issue.

horenmar:
The failures were on Travis's end. Anyway, there is not much of a reason to not merge the changes, but you really should not define lower case macros, especially to serve as typedefs...

dtgriscom:
Sorry; dealing with historical code. And thanks for the merge.

PR #1843, opened 2020-01-25
Feature: generic matchers

melak47:
## Description

Currently, composable matchers are based on `MatcherBase<T>`, like `EqualsMatcher<T>`.
`EqualsMatcher<T>` can compare two `std::vector<T>`s for equality, but not a `std::vector` and a `std::array`, for example.
This can make writing tests cumbersome somtimes, as you either need to write your own matcher
for whatever containers you need to work with, or convert everything to `std::vector`.

This PR aims to implement support for matchers that can be generic in their argument types, while also:

- not breaking compatibility with existing matchers
- allowing combining generic and non-generic matchers via &&, ||, !
- (ideally) not impacting compile time of code that only uses non-generic matchers

Here is a motivating example:
```cpp
TEST_CASE("example") {
    std::array<int, 3> container{1,2,3};

    std::array<int, 3> a{1,2,3};
    std::vector<int> b{0,1,2};
    std::list<int> c{4,5,6};

    CHECK_THAT(container, EqualsRange(a) || EqualsRange(b) || EqualsRange(c));
}
```
While it's easy to write your own custom `EqualsRange` matcher, making it composable with other matchers using &&, ||, ! is not, since that requires deriving from a shared `MatcherBase<T>`.

With the changes from this PR, an implementation of `EqualsRange` could look like this:
```cpp
template<typename Range>
struct EqualsRangeMatcher : Catch::MatcherGenericBase {

    EqualsRangeMatcher(Range const& range) : range{range} {}

    template<typename OtherRange>
    bool match(OtherRange const& other) const {
        using std::begin;
        using std::end;

        return std::equal(begin(range), end(range), begin(other), end(other));
    }

    std::string describe() const override {
        return "Equals: " + Catch::rangeToString(range);
    }

private:
    Range const& range;
};

template<typename Range>
auto EqualsRange(const Range& range) -> EqualsRangeMatcher<Range> {
    return EqualsRangeMatcher<Range>{range};
}
```


horenmar:
I fixed it up a bit.

src/catch2/catch_matchers_templates.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+
+        template<typename... MatcherTs, std::size_t... Idx>
+        std::string describe_multi_matcher(std::string const& combine, std::vector<void const*> const& matchers, index_sequence<Idx...>) {
+            std::array<std::string, sizeof...(MatcherTs)> descriptions {
+                static_cast<MatcherTs const*>(matchers[Idx])->toString()...
+            };
+
+            std::string description;
+            description.reserve( 4 + descriptions.size() * 32 );
```
horenmar: At this point we should already know the final size from combination of the lengths above, the `combine` size and the leading and finishing noise.

src/catch2/catch_matchers_templates.hpp
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        template<typename T, typename... MatcherTs, typename Arg, std::size_t Idx, std::size_t... Indices>
+        bool match_any_of(Arg const& arg, std::vector<void const*> const& matchers, index_sequence<Idx, Indices...>) {
+            return static_cast<T const*>(matchers[Idx])->match(arg) || match_any_of<MatcherTs...>(arg, matchers, index_sequence<Indices...>{});
+        }
+
+
+        template<typename... MatcherTs, std::size_t... Idx>
+        std::string describe_multi_matcher(std::string const& combine, std::vector<void const*> const& matchers, index_sequence<Idx...>) {
```
horenmar: I would also consider using `StringRef` together with its UDL for the `combine` arguments, as we then don't have to rely on good implementation of `std::string` to avoid allocations.

src/catch2/catch_matchers_templates.hpp
melak47: good point

src/catch2/catch_matchers_templates.hpp
melak47: will do

tests/SelfTest/UsageTests/Matchers.tests.cpp
```diff
@@ -556,65 +556,218 @@ namespace { namespace MatchersTests {
...
+        }
+
+        bool match(int i) const {
+            return i == 45;
+        }
+
+        EvilMatcher const* operator& () const {
+            throw EvilAddressOfOperatorUsed();
+        }
+    };
```
horenmar: How about `operator,`? 😈

tests/SelfTest/UsageTests/Matchers.tests.cpp
```diff
@@ -556,65 +556,218 @@ namespace { namespace MatchersTests {
...
+
+    TEST_CASE("Overloaded address-of operator is not used", "[matchers][templated]") {
+        REQUIRE_NOTHROW(EvilMatcher() || EvilMatcher());
+        REQUIRE_NOTHROW(EvilMatcher() && EvilMatcher());
+    }
```
horenmar: I would add `!` just to be thorough.

tests/SelfTest/UsageTests/Matchers.tests.cpp
melak47: I don't think anything here is vulnerable to that *right now*, but it can't hurt :)

horenmar:
Cool, thanks.

src/catch2/catch_matchers_templates.hpp
```diff
@@ -0,0 +1,269 @@
...
+#include <catch2/catch_matchers.h>
+#include <catch2/catch_stringref.h>
+
+#include <array>
+#include <algorithm>
+#include <memory>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
```
horenmar: This include should not be needed (I am also not sure what <memory> is used for, but I did not look hard)

melak47: I forgot to remove the <vector> include.

melak47:
Sorry, rebased again.

PR #1844, opened 2020-01-27
Make warmup_time be configurable

khyperia:
There's a constant, `warmup_time`, that is used for both timer statistics discovery, as well as test warmup. The timer statistics usage is fine, but, the default constant of 100ms used for each and every test warmup makes running a large number of benchmarks take an unreasonable amount of time (the project I'm working on has many very small/fast benchmarks).

This PR makes that constant be configurable with a command line switch, instead of a `const`.

Thanks! :sparkles:

### DIFF
--- a/docs/command-line.md
+++ b/docs/command-line.md
@@ -24,6 +24,7 @@
 [Specify the number of resamples for bootstrapping](#specify-the-number-of-resamples-for-bootstrapping)<br>
 [Specify the confidence-interval for bootstrapping](#specify-the-confidence-interval-for-bootstrapping)<br>
 [Disable statistical analysis of collected benchmark samples](#disable-statistical-analysis-of-collected-benchmark-samples)<br>
+[Specify the amount of time in milliseconds spent on warming up each test](#specify-the-amount-of-time-in-milliseconds-spent-on-warming-up-each-test)<br>
 [Usage](#usage)<br>
 [Specify the section to run](#specify-the-section-to-run)<br>
 [Filenames as tags](#filenames-as-tags)<br>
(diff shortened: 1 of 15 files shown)


include/internal/catch_commandline.cpp
```diff
@@ -213,6 +213,9 @@ namespace Catch {
             | Opt( config.benchmarkNoAnalysis )
                 ["--benchmark-no-analysis"]
                 ( "perform only measurements; do not perform any analysis" )
+            | Opt( config.benchmarkWarmupTime, "benchmarkWarmupTime" )
+                ["--benchmark-warmup-time"]
+                ( "amount of time in milliseconds spent on warming up each test (default: 500)" )
```
horenmar: This does not track with the default in code.

docs/command-line.md
```diff
@@ -317,6 +318,12 @@ Must be between 0 and 1 and defaults to 0.95.
 When this flag is specified no bootstrapping or any other statistical analysis is performed.
 Instead the user code is only measured and the plain mean from the samples is reported.
 
+<a id="benchmark-warmup-time"></a>
+## Specify the amount of time in milliseconds spent on warming up each test
+<pre>--benchmark-warmup-time</pre>
+
+Configure the amount of time spent warming up each test.
+
```
horenmar: This should get an "introduced in" placeholder.

horenmar:
I'll have proper look later this week, but two things jumped out to me.

khyperia:
Great! Also, the constant `warmup_time` is still used in the timer resolution/cost calculation warmup, let me know if you think that should also use this flag, or if it should use the `warmup_time` constant. (I have a local change that makes it also use the flag, but, yeah)

horenmar:
IIRC those are done only once, so they shouldn't cause a problem for users with a large numbers of benchmarks.

Anyway, I fixed a missing ToC entry in docs and rebased the approvals, I'll merge it after CI passes.

khyperia:
thanks so much! :heart:

PR #1846, opened 2020-01-27
Make CATCH_BREAK_INTO_DEBUGGER be user-configurable

khyperia:
Unfortunately, the project I work on supports platforms under NDA, and so I can't upstream the fixes here.

(It'd be really nice if all platform configuration macros like this were externally configurable! But just this one is the one we need right now.)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,6 +16,7 @@
 [Windows header clutter](#windows-header-clutter)<br>
 [Enabling stringification](#enabling-stringification)<br>
 [Disabling exceptions](#disabling-exceptions)<br>
+[Overriding Catch's debug break (`-b`)](#overriding-catchs-debug-break--b)<br>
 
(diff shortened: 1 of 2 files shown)


khyperia:
The test failure is `Error: retrieving gpg key timed out.`, not sure how to retrigger tests

horenmar:
AFAIK only accounts with access to the repo are allowed to rerun Travis tests.

Anyway, the change looks reasonable enough, but I'd like to also see it added to the documentation. (See `docs/contributing.md` how to properly add a "introduced in" placeholder)

khyperia:
Let me know if that's what you were thinking of with documentation!

horenmar:
Thanks. The documentation was good enough but I noticed that it was a subheading of "Disabling Exceptions", so I pulled it out and rewrote slightly.

PR #1847, opened 2020-01-27
Make hidden tags behave identically

dmsteck:
## Description

There are multiple tags to hide tests: `[.]`, `[.something]` and `[!hide]`. The Catch2 documentation (see [here](https://github.com/catchorg/Catch2/blob/master/docs/test-cases-and-sections.md#special-tags)) states that, for compatibility and consistency purposes, if one of these is used for a test case, then *all* of them are added implicitly.

However, this isn't actually how Catch2 behaves. As it turns out, if Catch2 encounters a hidden test, it will always make sure there is a `[.]` tag, but it will not add a `[!hide]` tag. This means that `TEST_CASE("Something", "[!hide]")` will result in a `[.][!hide]` signature, whereas `TEST_CASE("Something2", "[.]")` will produce a `[.]` signature.

A simple way to observe this behaviour is to compile `SelfTest` and observe that `SelfTest [.]` and `SelfTest [!hide]` produce different results.

## Changes

This PR enables `[.][!hide]` for all hidden tests by modifying the corresponding code in `internal/catch_test_case_info.cpp`. I have also updated the approval test results.


horenmar:
Thanks, I forgot `[!hide]` is a thing. However, the PR is missing a "real" unit test for the behaviour.

PR #1854, opened 2020-02-01
catch_compiler_capabilities.h: use proper math define

neheb:
C++11 math requires _GLIBCXX_USE_C99_MATH_TR1 to be true with gcc/clang.

Also fixes an issue with uClibc-ng where __UCLIBC__ is defined in features.h but
that is not included here and is thus no-op.

Similar to how boost does it: https://github.com/boostorg/math/blob/5f8561bcfd4f1784e729acd333d3532e30adc477/include/boost/math/tools/roots.hpp#L863

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -203,7 +203,7 @@
     #define CATCH_CONFIG_COLOUR_NONE
 #endif
 
-#if defined(__UCLIBC__)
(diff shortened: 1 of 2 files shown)


neheb:
Actually this has the same problem. hmmm...

horenmar:
Either I can fix the missing decl warnings when I get around to it, or you will have to :-)

PR #1858, opened 2020-02-04
Refactoring

offa:
# Description

- stdioh. replaced with cstdio
- Local variables made const

## GitHub Issues

*- none -*


horenmar:
I appreciate the effort, but e.g. the `testSpec` variables would be better served with making them a `const&`, to avoid the copy altogether, rather than just adding `const`.

offa:
Thanks for that suggestion, I have updated my PR accordingly and changed to east-const to match the existing style.

PR #1864, opened 2020-02-10
Switch back to uncolored output after printing filters

morxa:
## Description
After printing the list of filters, switch back from yellow to black before printing a newline to avoid the remaining output to be colored in yellow.

This is especially useful when the output is prefixed. In [Fawkes](https://github.com/fawkesrobotics/fawkes/), our build system prepends the test output with `[TEST]`, which is colored incorrectly:

old:
image

with this PR:
image

This is a purely aesthetic change, but I found it annoying enough to want it to be fixed.

### DIFF
--- a/include/reporters/catch_reporter_console.cpp
+++ b/include/reporters/catch_reporter_console.cpp
@@ -685,9 +685,8 @@ void ConsoleReporter::printSummaryDivider() {
 
 void ConsoleReporter::printTestFilters() {
     if (m_config->testSpec().hasFilters()) {
-        stream << Colour(Colour::BrightYellow)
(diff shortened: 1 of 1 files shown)


horenmar:
Interestingly, I cannot reproduce this, but I am fairly sure I saw this behaviour recently, but did not have time to investigate a small visual bug.

Can you check if just using Colour as a kind of a ScopeGuard works for your?

```cpp
    if (m_config->testSpec().hasFilters()) {
        Colour c(Colour::BrightYellow);
        stream << "Filters: " << serializeFilters(m_config->getTestsOrTags()) << '\n';
    }
```

horenmar:
@morxa I modified the branch somewhat. Can you check if it works for you?

PR #1866, opened 2020-02-13
Add command line option 'never' to --wait-for-keypress

tstenner:
I want my end-users on Windows to be able to run the tests without navigating to the folder on the command-line and view the results before the Window closes, so I set the default value for the option in the config runner:

``` cpp
session.configData().waitForKeypress = Catch::WaitForKeypress::BeforeExit;
```

On the CI systems, I can't wait for a keypress so I'd like to disabled the prompt by adding `--wait-for-keypress never` to the command line.

### DIFF
--- a/docs/command-line.md
+++ b/docs/command-line.md
@@ -271,7 +271,7 @@ See [The LibIdentify repo for more information and examples](https://github.com/
 
 <a id="wait-for-keypress"></a>
 ## Wait for key before continuing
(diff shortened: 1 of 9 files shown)


horenmar:
Looks reasonable enough. You might want to add a simple test that the CLI option is accepted.

horenmar:
Thanks, I modified the tests a tiny bit before merging.

PR #1868, opened 2020-02-17
Create a SYSTEM library in case Catch2 is included as a subproject.

nicber:
Define the CMake target as a SYSTEM library if Catch2 is included as a subproject. This is necessary to make sure that linting tools (such as `clang-tidy`) do not report errors about Catch2's headers.

In case the library is not included as a subproject keep it as a normal library, since it is not a good idea to silence linters when developing the framework itself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,11 +90,19 @@ target_compile_features(Catch2
     cxx_variadic_macros
 )
 
-target_include_directories(Catch2
-  INTERFACE
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/single_include>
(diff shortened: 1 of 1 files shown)


horenmar:
Hey, sorry for not getting back to you sooner, but I won't be merging this.

The problem is a combination of factors
* MSVC supports system headers with subtly different semantics than GCC/Clang
* Because of the above, CMake does not support `SYSTEM` for MSVC
* If you want, there is a fairly reasonable workaround by doing `target_include_directories(${target} SYSTEM PRIVATE $<TARGET_PROPERTY:Catch2,INTERFACE_INCLUDE_DIRECTORIES>)`.

nicber:
Would you consider merging this if the test included checking if the compiler is MSVC and not using `SYSTEM` then?

PR #1874, opened 2020-02-24
include <iterator> in catch_stats.hpp

khyperia:
Needed for `std::back_inserter` on some platforms (in particular, windows) when building `catch_stats.cpp`.


horenmar:
Interesting, I changed the v3 branch to always build benchmarking tests and I do not remember having to fix this.

Either way, thanks.

PR #1876, opened 2020-02-26
Don't cast nanoseconds to integers when writing output

khyperia:
I mostly care about `catch_reporter_xml.cpp`, but, figured I'd also do `catch_reporter_console.cpp` while I'm in the area.

Also, some context, since it's not plainly visible in the diff - stdev (just below) is already being written as doubles, without casting to int, so I don't think there's an issue with mean being written as a double as well


horenmar:
I had some misgivings about the resulting formatting for `ConsoleReporter` (I don't want to tell users that something took `1.0000002213` ns), but it looks fine and there probably should be a more significant cleanup of how the times are written if I want it to be human readable.

PR #1881, opened 2020-03-10
Add king as commercial user

magnus-ornebring:
## Description
Several teams here at (game development company) King make use of this excellent testing library and we would want to be included in the list of commercial users.

## GitHub Issues
N/A


horenmar:
Thanks

PR #1886, opened 2020-03-19
Fixed typo in "benchmark name" column width calculation.

### LINKED ISSUE #1885: "benchmark name" has overly wide column width
**Describe the bug**
The "benchmark name" column is too wide for the default `CATCH_CONFIG_CONSOLE_WIDTH` in full analysis mode, which prevents output from appearing as intended on a single line in a terminal which is 80 columns wide.

Example of current output:
```benchmark name                                  samples       iterations    estimated
                                                mean          low mean      high mean
                                                std dev       low std dev   high std dev
to_hex_old 0                                              100            17     2.0876 ms 
                                                     1.188 us      1.177 us      1.215 us 
                                                        84 ns         36 ns        152 ns 
```

**Expected behavior**

```benchmark name                        samples       iterations    estimated
                                      mean          low mean      high mean
                                      std dev       low std dev   high std dev
to_hex_old 0                                    100            17       2.04 ms 
                                           1.112 us      1.103 us      1.132 us 
                                              64 ns         34 ns        121 ns 
```

**Reproduction steps**
Use `BENCHMARK` in a test. See the output.

**Additional context**
The problem is caused by setting the "benchmark name" column width to `CATCH_CONFIG_CONSOLE_WIDTH - 32` in `ConsoleReporter::ConsoleReporter()`, when the other three columns are each 14 characters wide. (14*3 = 42 > 32).

uckelman-sf:
## Description
The width for the "benchmark name" column in full analysis mode is too large, which makes the output spill beyond `CATCH_CONFIG_CONSOLE_WIDTH`. 

The width was set to `CATCH_CONFIG_CONSOLE_WIDTH - 32`, but the other three columns are each 14 wide, for a total of 42, not 32, thus giving a total line length of `CATCH_CONFIG_CONSOLE_WIDTH + 10`, when it should obviously not exceed `CATCH_CONFIG_CONSOLE_WIDTH`. (My guess is that `32` was a typo for `42`.)

The best solution I see is setting the width of the "benchmark name" column to `CATCH_CONFIG_CONSOLE_WIDTH - 43`, the same as it is in the no analysis case. This PR does that.

## GitHub Issues
Fixes #1885.

### DIFF
--- a/include/reporters/catch_reporter_console.cpp
+++ b/include/reporters/catch_reporter_console.cpp
@@ -362,7 +362,7 @@ ConsoleReporter::ConsoleReporter(ReporterConfig const& config)
         else
         {
             return{
-                { "benchmark name", CATCH_CONFIG_CONSOLE_WIDTH - 32, ColumnInfo::Left },
+                { "benchmark name", CATCH_CONFIG_CONSOLE_WIDTH - 43, ColumnInfo::Left },
                 { "samples      mean       std dev", 14, ColumnInfo::Right },
                 { "iterations   low mean   low std dev", 14, ColumnInfo::Right },
                 { "estimated    high mean  high std dev", 14, ColumnInfo::Right }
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks, the changes, and the analysis seems reasonable.

The failing CI was because of a test that expected the specific look of the header, including whitespace. I updated it so that the new, shortened, header is checked instead.

PR #1891, opened 2020-03-23
Replace Deprecated C-Headers with the C++ ones

pi1024e:
## Description
 I switched out the deprecated C headers for the STL, and replaced them for the C++ ones.

### DIFF
--- a/include/internal/catch_debugger.cpp
+++ b/include/internal/catch_debugger.cpp
@@ -14,8 +14,7 @@
 
 #if defined(CATCH_PLATFORM_MAC) || defined(CATCH_PLATFORM_IPHONE)
 
-#  include <assert.h>
-#  include <stdbool.h>
+#  include <cassert>
 #  include <sys/types.h>
 #  include <unistd.h>
 #  include <cstddef>


horenmar:
I like the changes in theory, _but_ there are some problem with the PR as-is.

1. `<cstdbool>` has actually been removed from C++20, as it does not provide anything in C++. I think the reason it is even used is copy paste from somewhere, so that specific include can be removed completely.
2. Not all platforms support using functions from global namespace if you include the C++ variant of the header. This means that if you e.g. change `#include <signal.h>` to `#include <csignal>`, you also need to check that all uses of `signal` or `raise` are now `std::` qualified.

pi1024e:
However, they are std:: qualified.

horenmar:
First thing first, please only keep one PR open for one topic, and just update the originating branch accordingly 😃 If you need to make further changes because of CI, just add commits -- it is easier to review new commits than one perpetually rebased one, and you can just rebase and squash the final code.

include/internal/catch_compiler_capabilities.h
```diff
@@ -235,7 +235,6 @@
   #    if defined(__clang__) && (__clang_major__ < 8)
          // work around clang bug with libstdc++ https://bugs.llvm.org/show_bug.cgi?id=31852
          // fix should be in clang 8, workaround in libstdc++ 8.2
-  #      include <ciso646>
```
horenmar: This needs to be kept in. The trick to this header is that it is basically free to include on any/all platforms, and is required to provide you with the version macros of the standard library (such as `__GLIBCXX__` used below).

In C++20, `<version>` has been standardized for this, but Catch2 v2 targets C++11 and v3 targets C++14.

include/internal/catch_debugger.h
```diff
@@ -39,7 +39,7 @@ namespace Catch {
     #if defined(__GNUC__) && (defined(__i386) || defined(__x86_64))
         #define CATCH_TRAP() asm volatile ("int $3") /* NOLINT */
     #else // Fall back to the generic way.
-        #include <signal.h>
+        #include <csignal>
 
         #define CATCH_TRAP() raise(SIGTRAP)
```
horenmar: This `raise` now needs to be namespaced.

include/internal/catch_fatal_condition.h
```diff
@@ -35,7 +35,7 @@ namespace Catch {
 
 #elif defined ( CATCH_CONFIG_POSIX_SIGNALS )
 
-#include <signal.h>
+#include <csignal>
```
horenmar: This should also remain `<signal.h>`. This file relies on POSIX-provided `<signal.h>`, rather than the C-standard `<signal.h>`.

horenmar:
lgtm

PR #1898, opened 2020-03-31
Add vcpkg installation instructions

PhoebeHui:
Catch2 is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build Catch2, ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for Catch2 and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.


horenmar:
Thanks

PR #1899, opened 2020-03-31
catch_reporter_junit added xml attribute status

joda01:
## Description

Added "status" xml attribute to testcase section. 
This is done to be compatible with google test junit xml output.


horenmar:
What does the status mean?

joda01:
@horenmar The status tells if this test case has been executed or not. In case of catch this should always be run I think. GoogleTest has the possibility to skip tests.

horenmar:
Catch2 actually does sometimes skip tests (you can mark tests as throwing and ask Catch2 to skip throwing tests, and arguably you could say that tests not selected to be run are also skipped), and there is a reporter interface member function for that (`skipTest`), but I am willing to leave that to a different PR.

PR #1907, opened 2020-04-10
Remove usage of __builtin_constant_p under IBM XL

AndrewGaspar:
# Description

This change eliminates use of the `__builtin_constant_p` intrinsic for IBM XL compilers that use the clang front end. Under XL, code is generated to call destructors for temporaries in expressions passed to `__builtin_constant_p`, but the code to initialize those temporaries is not emitted. For example, IBM XL will generate a call to `std::string::~string` for the following code:

```c++
__builtin_constant_p(std::string("12") + "34");
```

This often results in a segfault in the case of `std::string::~string`, as `free` will likely be called on an invalid address.

Concretely, this code would result in a segfault:
```c++
REQUIRE(std::string("12") + "34" == "1234");
```

Previously, this the tests would fail when building with IBM XL 16.1.1.7. Now they all pass.

The bug has been reported to IBM. I will re-enable this for future versions of IBM XL when the problem is fixed.

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -58,7 +58,20 @@
 #    define CATCH_INTERNAL_START_WARNINGS_SUPPRESSION _Pragma( "clang diagnostic push" )
 #    define CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION  _Pragma( "clang diagnostic pop" )
 
+// As of this writing, IBM XL's implementation of __builtin_constant_p has a bug
+// which results in calls to destructors being emitted for each temporary,
+// without a matching initialization. In practice, this can result in something
+// like `std::string::~string` being called on an uninitialized value.
+//
+// For example, this code will likely segfault under IBM XL:
+// ```
+// REQUIRE(std::string("12") + "34" == "1234")
+// ```
+//
+// Therefore, `CATCH_INTERNAL_IGNORE_BUT_WARN` is not implemented.
+#  if !defined(__ibmxl__)
 #    define CATCH_INTERNAL_IGNORE_BUT_WARN(...) (void)__builtin_constant_p(__VA_ARGS__) /* NOLINT(cppcoreguidelines-pro-type-vararg) */
+#  endif
 
 
 #    define CATCH_INTERNAL_SUPPRESS_GLOBALS_WARNINGS \


horenmar:
> Under XL, code is generated to call destructors for temporaries in expressions passed to `__builtin_constant_p`, but the code to initialize those temporaries is not emitted.

🤦 

It shouldn't evaluate the expression at all, it is just an unevaluated context that allows lambdas (unlike `sizeof`). Anyway, thanks for the fix.

AndrewGaspar:
Don't touch your face! 🙂

I didn't actually realize [how recently this issue was introduced](https://github.com/catchorg/Catch2/commit/b7b346c3e56030d19ac58d33326b69bbaf226552).
Thanks for merging quickly!

AndrewGaspar:
Oh, I had assumed that this had been tested in clang. I should have checked.

AndrewGaspar:
@cebowler I can't repro this on either Apple's Clang distribution or Clang 9.0. Using this code:

```c++
#include <iostream>

class Foo {
public:
    Foo() {
        std::cout << "Foo::Foo" << std::endl;
    }

    ~Foo() {
        std::cout << "Foo::~Foo" << std::endl;
    }
};

int main(int argc, char *argv[]) {
    __builtin_constant_p(Foo());
}
```

If clang exhibited the bug, I would expect to see the following output:
```
Foo::~Foo
```

AndrewGaspar:
Tried it on a Power9 system with clang 7 and clang 8 and couldn't reproduce the issue. Only with an XL compiler.

hubert-reinterpretcast:
Thanks for checking. The similar Clang issues that have been reported have also been fixed:
llvm/llvm-project@4b03dd7b849e8f5068dc8d72c6eab724c22a2805
llvm/llvm-project@e128f710ea871bab5ed14b1944caa935ed61b003

AndrewGaspar:
Must have been introduced fairly recently - didn't have a clang 10.0 or nightly build lying around to test against. Our IBM XL compiler is fairly recently, so not surprising it would have picked up that issue.

horenmar:
As far as I can see, the linked issues are about _when_ `__builtin_constant_p` evaluates to true, but not about what happens with the code inside.

hubert-reinterpretcast:
> As far as I can see, the linked issues are about _when_ `__builtin_constant_p` evaluates to true, but not about what happens with the code inside.

That's not what I see. Please refer to the linked bug report: https://llvm.org/PR45535.

hubert-reinterpretcast:
That said, the implementation in XL does not act exactly the same as in Clang.

horenmar:
@hubert-reinterpretcast Well, I stand corrected.

Anyway, I would welcome a PR for specific version when IBM XL fixes this.

-----

As for Clang, it currently seems to work properly, so until bug reports with specific version come in, I am going to keep the `__builtin_constant_p`

AndrewGaspar:
btw, IBM has fixed this in a recent compiler release, at least to us. At some point this could be re-enabled for certain IBM XL versions if somebody is so inclined.

horenmar:
Yeah, I am open to a follow-up PR, but I don't intend to get hold of IBM XL to test it myself.

PR #1908, opened 2020-04-10
Change random test shuffling technique

jbytheway:
## Description
Previously a random test ordering was obtained by applying `std::shuffle` to the tests in declaration order.  This has two problems:

- It depends on the declaration order, so the order in which the tests will be run will be platform-specific.
- When trying to debug accidental inter-test dependencies, it is helpful to be able to find a minimal subset of tests which exhibits the issue. However, any change to the set of tests being run will completely change the test ordering, making it difficult or impossible to reduce the set of tests being run in any reasonably efficient manner.

Therefore, change the randomization approach to resolve both these issues.

Generate a random value based on the user-provided RNG seed.  Convert every test case to an integer by hashing a combination of that value with the test name.  Sort the test cases by this integer.

The test names and RNG are platform-independent, so this should be consistent across platforms.  Also, removing one test does not change the integer value associated with the remaining tests, so they remain in the same order.

To hash, use the FNV-1a hash, except with the basis being our randomly selected value rather than the fixed basis set in the algorithm.  Cannot use `std::hash`, because it is important that the result be platform-independent.

### DIFF
--- a/include/internal/catch_test_case_registry_impl.cpp
+++ b/include/internal/catch_test_case_registry_impl.cpp
@@ -15,27 +15,66 @@
 #include "catch_string_manip.h"
 #include "catch_test_case_info.h"
 
+#include <algorithm>
+#include <iterator>
+#include <random>
 #include <sstream>
 
 namespace Catch {
 
+    struct HashTest {
+        explicit HashTest( Catch::SimplePcg32& rng ) {
+            basis = rng();
+            basis <<= 32;
+            basis |= rng();
+        }
+
+        uint64_t basis;
+
+        uint64_t operator()( TestCase const& t ) const {
+            // Modified FNV-1a hash
+            static constexpr uint64_t prime = 1099511628211;
+            uint64_t hash = basis;
+            for( const char c : t.name ) {
+                hash ^= c;
+                hash *= prime;
+            }
+            return hash;
+        }
+    };
+
     std::vector<TestCase> sortTests( IConfig const& config, std::vector<TestCase> const& unsortedTestCases ) {
-
-        std::vector<TestCase> sorted = unsortedTestCases;
-
         switch( config.runOrder() ) {
-            case RunTests::InLexicographicalOrder:
+            case RunTests::InLexicographicalOrder: {
+                std::vector<TestCase> sorted = unsortedTestCases;
                 std::sort( sorted.begin(), sorted.end() );
-                break;
-            case RunTests::InRandomOrder:
+                return sorted;
+            }
+            case RunTests::InRandomOrder: {
                 seedRng( config );
-                std::shuffle( sorted.begin(), sorted.end(), rng() );
-                break;
+                HashTest h( rng() );
+                std::vector<std::tuple<uint64_t, std::string, TestCase const*>> indexed_tests;
+                indexed_tests.reserve( unsortedTestCases.size() );
+                std::transform( unsortedTestCases.begin(), unsortedTestCases.end(),
+                    std::back_inserter( indexed_tests ),
+                    [&]( TestCase const& t ) {
+                      return std::make_tuple( h(t), t.name, &t );
+                    } );
+                std::sort( indexed_tests.begin(), indexed_tests.end() );
+                std::vector<TestCase> sorted;
+                sorted.reserve( unsortedTestCases.size() );
+                std::transform( indexed_tests.begin(), indexed_tests.end(),
+                    std::back_inserter( sorted ),
+                    []( std::tuple<uint64_t, std::string, TestCase const*> const& t ) {
+                        return *std::get<2>( t );
+                    } );
+                return sorted;
+            }
             case RunTests::InDeclarationOrder:
                 // already in declaration order
                 break;
         }
-        return sorted;
+        return unsortedTestCases;
     }
 
     bool isThrowSafe( TestCase const& testCase, IConfig const& config ) {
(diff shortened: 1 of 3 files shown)


jbytheway:
After reading some of the commit history in this area (specifically 535da5c51) I realise that this is not *entirely* platform independent yet, because `std::uniform_int_distribution` may vary.  But at least the linker no longer affects it; it's a step in the right direction.

jbytheway:
I'd be happy to add a test for this, but after a brief glance at your testing infrastructure it wasn't clear to me where a test of command-line parameter like this belongs.

include/internal/catch_test_case_registry_impl.cpp
```diff
@@ -15,10 +15,34 @@
...
+    struct CompareTestsUsingRandom {
+        explicit CompareTestsUsingRandom( Catch::SimplePcg32& rng ) {
+            std::uniform_int_distribution<uint64_t> dist;
+            basis = dist( rng );
+        }
+
+        uint64_t basis;
+
+        uint64_t fnv_1a_hash( std::string const& s ) const {
+            static constexpr uint64_t prime = 1099511628211;
```
horenmar, line 32: I am not sure why the compiler doesn't warn on this, but AFAIK the constant will be truncated. To generate `uint64_t` constants, you need `ULL` suffix.

horenmar:
I like the general idea, but I will have to think about the approach a bit.

I have two misgivings right now:
* calculating two hashes on every comparison can prove slow when used for codebase with significant number of test cases
* I am not sure how to check how much randomness there will be in the order of the tests.

The first one should be easy enough to fix -- we can compute each hash once, sort those and then project the order back to test cases -- or we can benchmark say 100k test cases and see what it does.

The second one is obviously a bit harder and I will have to think about it first.

---------------

As to tests, generally everything possible should be tested via CTest. I do not see how to easily test this specific feature using just the facilities provided by CTest, so it will likely require a Python script that performs the actual test, and reports the result back.

The way I see it, the test should work something like
1) Invoke the `SelfTest` executable with `--list-tests --order rand --rng-seed <magic-value>`, record the order of listed tests
2) Invoke the `SelfTest` executable again, but limit the listing to a subset of tests, e.g. using `[approx]` and `[matchers]` tags. Check that the listed tests are in the same relative order as in step 1.
3) Do this once again, with just a single tag, and check once again.


The script will also need some sanity checks to avoid false negatives. The ones I can think of on top of my head are that the number of tests returned in step 1 should more than step 2, and step 2 should return more tests than step 3, and that there should always be at least 2 returned tests.

include/internal/catch_test_case_registry_impl.cpp
jbytheway: No.  The compiler will automatically pick the smallest integer type sufficient to represent the constant.  See e.g. N3337 [lex.icon] paragraph 2.

jbytheway:
I've added a test and rewritten the sorting code to only compute O(n) hashes rather than O(n log(n)).

Regarding your other concern: I see no reason to believe there will be any less randomness than there was with `std::shuffle`.  They're both derived from the same seed.  The limited entropy in that seed will be the limiting factor.

Should I move `struct HashTest` into an internal namespace of some kind?

horenmar:
My misgiving with this approach is that `std::shuffle` guarantees that every permutation of the input is equally likely (at least given true URBG and properly implemented uniform distribution), while I do not know the statistical properties of randomly-seeded FNV1a 😃 

That is why I want to play with it a bit first -- generate some random seeds, check if there are suspicious results.

On a related note, to ensure stability properly, the code needs to handle hash collisions. I suggest just using test case name as the fallback comparison if hashes happen to be the same.

include/internal/catch_test_case_registry_impl.cpp
```diff
@@ -15,27 +15,63 @@
...
+#include <iterator>
+#include <random>
 #include <sstream>
 
 namespace Catch {
 
-    std::vector<TestCase> sortTests( IConfig const& config, std::vector<TestCase> const& unsortedTestCases ) {
+    struct HashTest {
+        explicit HashTest( Catch::SimplePcg32& rng ) {
+            std::uniform_int_distribution<uint64_t> dist;
```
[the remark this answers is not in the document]
jbytheway: Good idea.  Done.

jbytheway:
> My misgiving with this approach is that `std::shuffle` guarantees that every permutation of the input is equally likely (at least given true URBG and properly implemented uniform distribution), while I do not know the statistical properties of randomly-seeded FNV1a smiley
> 
> That is why I want to play with it a bit first -- generate some random seeds, check if there are suspicious results.

FWIW, I have added a test which verifies that the seeds from 1 to 100 generate different orderings.  Obviously that's not proof of much, but it does at least demonstrate that the distribution isn't utterly awful.

> On a related note, to ensure stability properly, the code needs to handle hash collisions. I suggest just using test case name as the fallback comparison if hashes happen to be the same.

Done.

include/internal/catch_test_case_registry_impl.cpp
horenmar: [Oh right](https://eel.is/c++draft/lex.icon#3). Nevermind.

horenmar:
As I said before, thanks for this -- the idea is quite useful. I am going to do a little further refactoring later, but it is nothing that stops this from being merged.

PR #1909, opened 2020-04-11
support custom allocators in vector Matchers

schallerr:
## Description

This PR allows to use `std::vector` with custom allocator (e.g. `Eigen::aligned_allocator`) in vector Matchers.

For `VectorContains`, the allocator must be specified explicitly since it cannot be deduced from the comparator. For the other Matchers, it's deduced from the comparator, but may also be specified explicitly in case the allocator differs between comparator and matched object.

### DIFF
--- a/include/internal/catch_matchers_vector.h
+++ b/include/internal/catch_matchers_vector.h
@@ -17,12 +17,12 @@ namespace Catch {
 namespace Matchers {
 
     namespace Vector {
-        template<typename T>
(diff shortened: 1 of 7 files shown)


horenmar:
The changes look reasonable, and I intend to rebase and merge the changes later, when I have some time.

Fair warning though, the v3 branch will deprecate all vector matchers in favour of generic range matchers.

PR #1910, opened 2020-04-13
Add new command-line option to specify a threshold for test duration printing

jbytheway:
## Description
A test runner already has a --durations option to print durations.  However, this isn't entirely satisfactory.

When there are many tests, this produces output spam which makes it hard to find the test failure output.  Nevertheless, it is helpful to be informed of tests which are unusually slow.

Therefore, introduce a new option --min-duration that causes all durations above a certain threshold to be printed.  This allows slow tests to be visible without mentioning every test.

In our CI we're currently emulating this with [some ugly grep](https://github.com/CleverRaven/Cataclysm-DDA/blob/ef0197a5fb40f3937a0ad764ba097ff229cdbc4e/build-scripts/build.sh#L12) on the output, but this is a bit limited, and harder to use when running the tests outside of CI.

I've made a bunch of choices in this implementation that could have gone differently:
* `--min-duration` is independent of `--durations`.  Specifying either will cause things to be printed.  I could have rolled them into a single option that takes either a string or numeric argument, or made `--min-duration` do nothing if `--durations` was not also given.
* I made `IConfig::shouldShowDuration` a member function.  It could have been a free function, but I wouldn't know where to put it.
* `--min-duration` only affects the console and compact reporters, not e.g. the XML reporter, because I figured output spam is not an issue there.
* It's tricky to test this change, because every test in `SelfTest` is so fast that it is printed as `0.000 s`.  So for now the test just verifies that no durations are printed if a minimum of 1s is specified.  But it's written in such a way that it should do something more interesting if slower tests are ever added.

### DIFF
--- a/include/internal/catch_commandline.cpp
+++ b/include/internal/catch_commandline.cpp
@@ -170,6 +170,9 @@ namespace Catch {
             | Opt( [&]( bool flag ) { config.showDurations = flag ? ShowDurations::Always : ShowDurations::Never; }, "yes|no" )
                 ["-d"]["--durations"]
                 ( "show test durations" )
+            | Opt( config.minDuration, "seconds" )
+                ["-D"]["--min-duration"]
+                ( "show test durations for tests taking at least the given number of seconds" )
             | Opt( loadTestNamesFromFile, "filename" )
                 ["-f"]["--input-file"]
                 ( "load test names to run from a file" )
--- a/include/internal/catch_config.hpp
+++ b/include/internal/catch_config.hpp
@@ -52,6 +52,7 @@ namespace Catch {
         Verbosity verbosity = Verbosity::Normal;
         WarnAbout::What warnings = WarnAbout::Nothing;
         ShowDurations::OrNot showDurations = ShowDurations::DefaultForReporter;
+        double minDuration = -1;
         RunTests::InWhatOrder runOrder = RunTests::InDeclarationOrder;
         UseColour::YesOrNo useColour = UseColour::Auto;
         WaitForKeypress::When waitForKeypress = WaitForKeypress::Never;
@@ -103,6 +104,7 @@ namespace Catch {
         bool warnAboutMissingAssertions() const override;
         bool warnAboutNoTests() const override;
         ShowDurations::OrNot showDurations() const override;
+        double minDuration() const override;
         RunTests::InWhatOrder runOrder() const override;
         unsigned int rngSeed() const override;
         UseColour::YesOrNo useColour() const override;
--- a/include/reporters/catch_reporter_compact.cpp
+++ b/include/reporters/catch_reporter_compact.cpp
@@ -275,8 +275,9 @@ private:
         }
 
         void CompactReporter::sectionEnded(SectionStats const& _sectionStats) {
-            if (m_config->showDurations() == ShowDurations::Always) {
-                stream << getFormattedDuration(_sectionStats.durationInSeconds) << " s: " << _sectionStats.sectionInfo.name << std::endl;
+            double dur = _sectionStats.durationInSeconds;
+            if ( shouldShowDuration( *m_config, dur ) ) {
+                stream << getFormattedDuration( dur ) << " s: " << _sectionStats.sectionInfo.name << std::endl;
             }
         }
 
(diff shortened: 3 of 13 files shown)


include/internal/catch_config.hpp
```diff
@@ -52,6 +52,7 @@ namespace Catch {
         Verbosity verbosity = Verbosity::Normal;
         WarnAbout::What warnings = WarnAbout::Nothing;
         ShowDurations::OrNot showDurations = ShowDurations::DefaultForReporter;
+        double minDuration = -1;
```
horenmar, line 55: I think having `0.` mean "turned off" is fine here. If an user wants all durations to be shown, they can set up `showDurations::Always`.

include/internal/catch_commandline.cpp
```diff
@@ -170,6 +170,9 @@ namespace Catch {
             | Opt( [&]( bool flag ) { config.showDurations = flag ? ShowDurations::Always : ShowDurations::Never; }, "yes|no" )
                 ["-d"]["--durations"]
                 ( "show test durations" )
+            | Opt( config.minDuration, "seconds" )
+                ["-D"]["--min-duration"]
+                ( "show test durations when at least the given number of seconds" )
```
horenmar, line 175:
```suggestion
                ( "show test durations for tests taking at least seconds" )
```
?

`when` definitely does not fit, but I am ok with changing it to just `with`.

horenmar:
I think `IConfig::shouldShowDuration` should be a free function somewhere in reporter interfaces/bases -- there is already precedent with some cross-reporter helpers being there.

horenmar:
Some extra thoughts on combining `-d` and `-D`:
* `ShowDurations::Always` should take priority over values set by `-D`.
* `ShowDurations::Never` should suppress printing any durations. (I also see only one place it is used, and I have some doubts about the usefulness there)
* This implies that setting `-D` is only valid if `-d` is not set. The question is, should Catch2 stop you from specifying both?

jbytheway:
> For testing, I would suggest adding a hidden test non-approval test that sleeps for, say, 100 ms, and another that sleeps for 500 ms and then check that they are (not) reported as appropriate...

To verify I've understood you: by "hidden" do you mean "marked `[.]`"?  So I could use e.g. `"[.min_duration_test]" for them and run them just when testing this feature?

jbytheway:
> I think `IConfig::shouldShowDuration` should be a free function somewhere in reporter interfaces/bases -- there is already precedent with some cross-reporter helpers being there.

In `include/reporters/catch_reporter_bases.hpp` I see `getFormatterDuration`.  Is that the sort of place you were thinking?

include/internal/catch_config.hpp
jbytheway: I'm always leery of using a value that might possibly have been intended as valid for a sentinel value.  You never know when there might be extra layers between the user and the value (e.g. a GUI or a wrapper script) which makes it less easy to pass other arguments.

include/reporters/catch_reporter_compact.cpp
```diff
@@ -275,8 +275,9 @@ class AssertionPrinter {
         }
 
         void CompactReporter::sectionEnded(SectionStats const& _sectionStats) {
-            if (m_config->showDurations() == ShowDurations::Always) {
-                stream << getFormattedDuration(_sectionStats.durationInSeconds) << " s: " << _sectionStats.sectionInfo.name << std::endl;
+            double dur = _sectionStats.durationInSeconds;
+            if (m_config->shouldShowDuration( dur ) == ShowDurations::Always) {
```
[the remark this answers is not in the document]
jbytheway: Fixed.

include/internal/catch_commandline.cpp
jbytheway: I went with your suggested change.

jbytheway:
>     * This implies that setting `-D` is only valid if `-d` is not set. The question is, should Catch2 stop you from specifying both?

My instinct is not to prevent it.  You might e.g. be calling the test runner through a wrapper script that uses `-D` and it would be annoying if you couldn't add `-d` to override.

horenmar:
> In include/reporters/catch_reporter_bases.hpp I see getFormatterDuration. Is that the sort of place you were thinking?

Yes.

> To verify I've understood you: by "hidden" do you mean "marked [.]"? So I could use e.g. `"[.min_duration_test]" for them and run them just when testing this feature?

And yes.

jbytheway:
I have refactored and updated the tests as suggested.  Over to you.

horenmar:
There are some things I am still not quite happy with, but it will be fastest if I merge the PR as is and fix it later.

nandanvasudevan:
Would it be possible to group the output according to the test case and tag?

horenmar:
@nandanvasudevan Everything is possible with a custom reporter. 🤷

PR #1911, opened 2020-04-13
Use macro to determine whether std::uncaught_exceptions is available

johnelse:
Catch assumes std::uncaught_exceptions is available whenever C++17 is
available, but for macOS versions older than 10.12 this is not the case.

Instead of checking the C++ version, use a macro to check whether the
feature is available.


horenmar:
Thanks.

Issue #1462, opened 2018-11-29
With CATCH_CONFIG_DISABLE_EXCEPTIONS the single-header still references exception handling functionality


nagisa:
## Description

When Catch2 (v2.5.0 tag) is used with `CATCH_CONFIG_DISABLE_EXCEPTIONS` define set, the single-file header will still refer to exception-handling specific functionality from the standard library.

Namely, in the `ExceptionTranslatorRegistrar` it directly references `std::rethrow_exception` and `std::current_exception()` which will make the library fail to compile when exceptions are disabled.

### Steps to reproduce

1. Use v.2.5.0 library with exceptions disabled and `CATCH_CONFIG_DISABLE_EXCEPTIONS` define set.

### Extra information

* Catch version: **v2.5.0**
* Operating System: **none** (bare-metal)
* Compiler+version: **IAR 8.11.3**

horenmar:
Interesting, all of our test builds on TravisCI passed (the ones with `EXTRAS=1` build without exceptions and `CATCH_CONFIG_DISABLE_EXCEPTIONS` enabled, e.g. [this one](https://travis-ci.org/catchorg/Catch2/jobs/459919751))

horenmar:
Ok, I am going to merge it to master and you can reopen the issue if it still comes up.

PR #1916, opened 2020-04-21
Fix build with CATCH_CONFIG_DISABLE_EXCEPTIONS enabled

khyperia:
Tested by building our program that consumes Catch2 with `-fno-exceptions`.

The #include change is a little abstract, so to be clear, this is the line that was failing to build (with `CATCH_ERROR not found`) https://github.com/catchorg/Catch2/blob/37cbf4a4fe7a2963f8ae6d6fe998f8ac9360cbc0/include/internal/catch_assertionhandler.cpp#L88

### DIFF
--- a/include/internal/catch_assertionhandler.cpp
+++ b/include/internal/catch_assertionhandler.cpp
@@ -15,6 +15,7 @@
 #include "catch_interfaces_registry_hub.h"
 #include "catch_capture_matchers.h"
 #include "catch_run_context.h"
+#include "catch_enforce.h"
 
 namespace Catch {
 
--- a/include/internal/catch_interfaces_exception.h
+++ b/include/internal/catch_interfaces_exception.h
@@ -46,6 +46,9 @@ namespace Catch {
             {}
 
             std::string translate( ExceptionTranslators::const_iterator it, ExceptionTranslators::const_iterator itEnd ) const override {
+#if defined(CATCH_CONFIG_DISABLE_EXCEPTIONS)
+                return "";
+#else
                 try {
                     if( it == itEnd )
                         std::rethrow_exception(std::current_exception());
@@ -55,6 +58,7 @@ namespace Catch {
                 catch( T& ex ) {
                     return m_translateFunction( ex );
                 }
+#endif
             }
 
         protected:


khyperia:
#1462 may be related?

horenmar:
Is this for the non-public platform you work on? Because I am still unsure on why our own CI using libstdc++/MSSTL passes, but people in the wild run into compilation errors with this toggle.

Anyway, these changes make sense, I should be able to merge them later today.

khyperia:
I'm also really confused why your CI passes - this particular thing isn't a non-public platform, I developed this PR on linux, compiling for linux. (Actual reason I'm doing this is for WASI, which doesn't support exceptions right now)

So compiling the below code with `clang++ -fno-exceptions main.cpp`

```c++
int main()
{
    try { }
    catch (...) { }
}
```

results in

```
main.cpp:3:5: error: cannot use 'try' with exceptions disabled
    try { }
    ^
1 error generated.
```

and this is unconditionally declaring a try block, so I have no idea why this compiles on CI in your no-exceptions test.
https://github.com/catchorg/Catch2/blob/37cbf4a4fe7a2963f8ae6d6fe998f8ac9360cbc0/include/internal/catch_interfaces_exception.h#L49

horenmar:
So, I had a theory that maybe it is OK in noninstantiated templates, but that is not true in small test case:

```
$ cat noexcept.cpp
class SomeClass {
    template <typename T>
    class InnerClass {
        void foo() {
            try {} catch(T const&) {}
        }
    };
};
$ clang++ -fno-exceptions -c noexcept.cpp
noexcept.cpp:5:13: error: cannot use 'try' with exceptions disabled
            try {} catch(T const&) {}
            ^
1 error generated.
```

After this I rechecked the CI test to make sure that the compilation flags did not get quietly broken, but this is the relevant part of the build output:

```
$ make DisabledExceptions-DefaultHandler VERBOSE=1
/usr/bin/clang++   -I/mnt/c/ubuntu/Catch2/single_include  -g   -fno-exceptions -std=c++11 -o CMakeFiles/DisabledExceptions-DefaultHandler.dir/X03-DisabledExceptions-DefaultHandler.cpp.o -c /mnt/c/ubuntu/Catch2/projects/ExtraTests/X03-DisabledExceptions-DefaultHandler.cpp

cd /mnt/c/ubuntu/Catch2/clang-v2-all/projects/ExtraTests && /usr/bin/clang++  -g   CMakeFiles/DisabledExceptions-DefaultHandler.dir/X03-DisabledExceptions-DefaultHandler.cpp.o  -o DisabledExceptions-DefaultHandler
```

As far as I can see, these should be correct. A fairly strong signal that it is that the behavioral tests outside pass, which means that Catch detects exceptions being disabled through compiler flags... but the invocations above compile fine.

I even tried invoking the compilation separately and without the compile-to-object files step, and it still "works for me"™️:

```
$ clang++ -I/mnt/c/ubuntu/Catch2/single_include -g -fno-exceptions /mnt/c/ubuntu/Catch2/projects/ExtraTests/X03-DisabledExceptions-DefaultHandler.cpp && ./a.out
<snip>
Catch will terminate because it needed to throw an exception.
The message was: Test failure requires aborting test!
```

horenmar:
And if I just preprocess the file, I get this
```
    class ExceptionTranslatorRegistrar {
        template<typename T>
        class ExceptionTranslator : public IExceptionTranslator {
        public:

            ExceptionTranslator( std::string(*translateFunction)( T& ) )
            : m_translateFunction( translateFunction )
            {}

            std::string translate( ExceptionTranslators::const_iterator it, ExceptionTranslators::const_iterator itEnd ) const override {
                try {
                    if( it == itEnd )
                        std::rethrow_exception(std::current_exception());
                    else
                        return (*it)->translate( it+1, itEnd );
                }
                catch( T& ex ) {
                    return m_translateFunction( ex );
                }
            }
```
inside the (massive, 2.5MB) output, so the compiler should see it as well. 😕

khyperia:
Got it. It's `#pragma clang system_header` in `single_include/catch2/catch.hpp` that somehow silences the error.

```
$ clang++ -fno-exceptions single_include/catch2/catch.hpp
single_include/catch2/catch.hpp:21:19: warning: #pragma system_header ignored in main file [-Wpragma-system-header-outside-header]
#    pragma clang system_header
                  ^
single_include/catch2/catch.hpp:3003:17: error: cannot use 'try' with exceptions disabled
                try {
                ^
1 warning and 1 error generated.

$ cat blah.cpp
#include "single_include/catch2/catch.hpp"
int main() { return 0; }

$ clang++ -fno-exceptions blah.cpp
(no output, success)

$ vim single_include/catch2/catch.hpp
(delete "#pragma clang system_header")

$ clang++ -fno-exceptions blah.cpp
In file included from blah.cpp:1:
./single_include/catch2/catch.hpp:2997:17: error: cannot use 'try' with exceptions disabled
                try {
                ^
1 error generated.
```

We've not been using the single_include file because we want to use our patches we submit straight away, instead of waiting for a release. (Turns out it's easier to not use the single_include header, than maintaining a fork or locally running `scripts/generateSingleHeader.py`)

That means that `include/internal/catch_interfaces_exception.h` is not under `#pragma clang system_header`, so building something random fails:

```
clang++ -fno-exceptions "/home/khyperia/me/baselib/External/Catch2/include/internal/catch_run_context.cpp"
In file included from /home/khyperia/me/baselib/External/Catch2/include/internal/catch_run_context.cpp:1:
In file included from /home/khyperia/me/baselib/External/Catch2/include/internal/catch_run_context.h:14:
/home/khyperia/me/baselib/External/Catch2/include/internal/catch_interfaces_exception.h:49:17: error: cannot use 'try' with exceptions disabled
                try {
                ^
1 error generated.
```

horenmar:
That's... definitely a thing.

And apparently I've fixed this back in December for the v3 branch https://github.com/catchorg/Catch2/commit/8b42acc328fc7b6c4526e649a30c4b3c198c974d#diff-750d9be38fbd1dba525187f947915c7eR50-R62. 🤷

Issue #1182, opened 2018-02-09
Provide a clang-format to keep code formatting mostly consistent


horenmar:
## Description
We don't have the code formatting rules written down anywhere, so contributors have to look around at other code and guess. We should provide them with `.clang-format` instead (and also hash out some formatting disagreements first).

PR #1920, opened 2020-04-29
Clang-format configuration

offa:
##  Description

Initial `.clang-format` file, open for discussion. The config doesn't contain much as the defaults fit most cases already. Documentation of all options is available [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

## Settings to discuss / change

- [x] :question:  `NamespaceIndentation`: Indention of namespaces – `All`, `Inner` or `None`?
- [x] :question: `IndentPPDirectives`: Indention of proprocessor directives

## GitHub Issues

#1182

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,25 @@
+---
+AccessModifierOffset: '-4'
+AlignEscapedNewlines: Left
+AllowAllConstructorInitializersOnNextLine: 'true'
+BinPackArguments: 'false'
+BinPackParameters: 'false'
+BreakConstructorInitializers: AfterColon
+ConstructorInitializerAllOnOneLineOrOnePerLine: 'true'
+DerivePointerAlignment: 'false'
+FixNamespaceComments: 'true'
+IncludeBlocks: Regroup
+IndentCaseLabels: 'false'
+IndentPPDirectives: AfterHash
+IndentWidth: '4'
+Language: Cpp
+NamespaceIndentation: All
+PointerAlignment: Left
+SpaceBeforeCtorInitializerColon: 'false'
+SpaceInEmptyParentheses: 'false'
+SpacesInParentheses: 'true'
+Standard: Cpp11
+TabWidth: '4'
+UseTab: Never
+
+...


horenmar:
Right, I'll try to find some time to test the configuration out.

However, since opening #1182, I started having some misgivings about clang-format. From my experience with other projects, it doesn't handle some constructs well, and the actual formatting depends on the specific _version_ of clang-format, making inconsistencies likely.

offa:
Please let me know if there are some issues. The result might not be perfect, but maybe we can get some consistency.

horenmar:
So, I spent some time playing with the config and made some further changes (pushed them as a separate commit). There is still at least one problem that needs fixing before I am willing to say "use clang-format now". See below

Some notes on the current settings
* `NamespaceIndentation` - I would love a hybrid between `Inner` and `All`, where the most-nested namespace is indented, but the less nested ones aren't. That doesn't exist, so I ended up going with `All`.
* `IndentPPDirectives` - This is currently inconsistent in the codebase, but `AfterHash` is the style I prefer for new code.
* `AllowShortEnumsOnASingleLine` - preferably would be set to `False`, but that seems to be a clang-format-11 thing.

The big problem is with `SpacesInParentheses`. I don't care for it that much personally, but the absolute majority of current code is written in style that is closely resembled by using it, but not quite. The problems come from parenthesized expressions (such as defensively written macro argument expansion), which aren't either function decl or function call.

I think I will end up turning it on and trying it out a bit more, because I don't want to generate mostly pointless diff against all code.

offa:
> AllowShortEnumsOnASingleLine - preferably would be set to False, but that seems to be a clang-format-11 thing.

I've just check the clang docs, you are right, clang-format 10 doesn't mention it.

> The problems come from parenthesized expressions (such as defensively written macro argument expansion), which aren't either function decl or function call.

I see, in the worst case it might be an option to disable clang format for particular sections.

> I think I will end up turning it on and trying it out a bit more, because I don't want to generate mostly pointless diff against all code.

:+1:

PR #1923, opened 2020-05-08
Create a BUILD file for compatibility with bazel

nitnelave:
With this change, it's much easier for bazel projects to depend on
Catch. They just need to add:
  - In the workspace:
  ```
http_archive(
    name = "com_github_catchorg_catch2",
    urls = ["https://github.com/catchorg/Catch2/archive/v2.12.1.tar.gz"],
    strip_prefix = "Catch2-2.12.1",
    sha256 = "e5635c082282ea518a8dd7ee89796c8026af8ea9068cd7402fb1615deacd91c3",
)
  ```
  Or the appropriate version/sha256.
  - For the tests, assuming that `test_main.cc` contains the
  `CATCH_CONFIG_MAIN`:
  ```
cc_library(
    name = "test_main",
    srcs = ["test_main.cc"],
    deps = ["@com_github_catchorg_catch2//:catch2"],
)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ Build
 cmake-build-*
 benchmark-dir
 .conan/test_package/build
+bazel-*
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,10 @@
+# Load the cc_library rule.
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+# Header-only rule to export catch2/catch.hpp.
+cc_library(
+  name = "catch2",
+  hdrs = ["single_include/catch2/catch.hpp"],
+  visibility = ["//visibility:public"],
+  includes = ["single_include/"],
+)

### CODE REVIEW

**Comment by nitnelave** [comment]
Note that several people have adapted their own solution for bazel:
 - https://github.com/evanmoran/catch2-bazel -- bazel overlay by @evanmoran 
 - https://martinsosic.com/cpp/bazel/catch2/2018/08/28/cpp-bazel-catch2.html -- download a copy of catch.hpp and check it in
 - https://stackoverflow.com/questions/52621760/how-to-set-up-catch2-in-bazel-project -- recommendation of the above on SO

**Comment by horenmar** [comment]
Eh, sure.

**Comment by nitnelave** [comment]
That's the kind of enthusiasm I like to see :smile: 

Thanks!

**Comment by horenmar** [comment]
Someone will have to provide the support for v3 branch or Bazel will be dropped again though.

**Comment by nitnelave** [comment]
I'm just a random user of catch, so I don't know much about v3, but I can help with support if needed. Can you point me to docs about the v3 branch, how to use it?

**Comment by nitnelave** [comment]
So, I had a look at the dev-v3 branch, and from what I understand, instead of `#include <catch2/catch.hpp>`, the user would include just what they need from `catch2/*.hpp`, right? And it's `.hpp` all the way, that's the only kind of file.

If that's the case, we can get by easily by providing a target that globs all the hpp files recursively. We might also want to exclude the `internal` folder, and place it in a separate, non-visible target so users don't depend on it. WDYT?

It would look something like:

```bazel
cc_library(
  name = "catch2_internal",
  hdrs = glob(["src/catch2/internal/**/*.hpp"]),
  visibility = ["//visibility:private"],
  includes = ["src/"],
)

cc_library(
  name = "catch2",
  hdrs = glob(["src/catch2/**/*.hpp"], exclude=["src/catch2/internal/**"]),
  visibility = ["//visibility:public"],
  deps = [":catch2_internal"],
  includes = ["src/"],
)
```

(Not tested, just a design proposition)

horenmar:
v3 is also moving to be a standard statically compiled library, so you need proper sources in the build specification.

nitnelave:
Do you have docs on how to use the v3 branch?

horenmar:
Basically all .cpp files (except for the one that provides main) are compiled into one static library. The main is compiled into a different static library, that depends on the first one.

User then can link either just the implementation and provide main themselves, or opt into the default main by linking the second one (or both, but both CMake and pkg-config express dependency between the one with main and the implementation one, so users can just link to the one with main).

nitnelave:
Created https://github.com/catchorg/Catch2/pull/1936

PR #1929, opened 2020-05-14
Fix compilation failure when using libstdc++10

akemimadoka:
## Description

Sorry for my poor English, I'm not a native speaker.
The failure case is published on https://wandbox.org/permlink/E3IuG3dbqTTDAwxv
Which can be fixed by this patch as https://wandbox.org/permlink/jDqVGe2TlJ8F3Q9r
This bug is caused by std::__detail::begin declared by bits/iterator_concepts.h, the function may be lookup by adl, like the case above, so here will be a matched but deleted begin, not the begin returning not_this_one
This patch fixed the bug by detecting whether expression begin(std::declval<T>()) is well-formed directly
The void_type is introduced because of issue mentioned [here](https://en.cppreference.com/w/cpp/types/void_t#Notes)

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -469,20 +469,27 @@ namespace Catch {
 #endif // CATCH_CONFIG_ENABLE_VARIANT_STRINGMAKER
 
 namespace Catch {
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks.

PR #1936, opened 2020-05-25
Add Bazel support for the v3 branch

nitnelave:
This PR adds support for easily depending on catch2 v3 as a Bazel dependency.

It adds 2 targets: `catch2` and `catch2_main`, with the `_main` version depending on the `catch2` library.

I didn't try to hide the `internal` headers (well, actually I tried and failed), but I guess that's the same as CMake (they're available by default).

To create a unit test, you just have to write a `BUILD` file like:

```bazel
cc_test(
    name = "test",
    srcs = ["test.cc"],
    deps = ["@catch2//:catch2_main"],
    copts = ["-std=c++14"],
)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ Build
 cmake-build-*
 benchmark-dir
 .conan/test_package/build
+bazel-*
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,25 @@
+# Load the cc_library rule.
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+# Static library, without main.
+cc_library(
+    name = "catch2",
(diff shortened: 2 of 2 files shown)


nitnelave:
Note that it's the v3 version of https://github.com/catchorg/Catch2/pull/1923

horenmar:
> I didn't try to hide the internal headers (well, actually I tried and failed), but I guess that's the same as CMake (they're available by default).

Right, a header being internal means more of a "user shouldn't use this in common usage" than "user shouldn't use this ever". The number of truly internal-the-user-will-never-need-them headers is currently low enough that I don't see a reason to keep them separate.

The build looks correct.

horenmar:
Thanks

PR #1950, opened 2020-06-04
Hide std::exception_ptr and friends if exceptions disabled

mblankertz-cetitec:
## Description
Some compilers, e.g. the Green Hills C++ compiler, react badly to the
appearance of std::exception_ptr, std::current_exception,
std::rethrow_exception and std::uncaught_exception(s). To allow usage of
Catch2 with these compilers when exceptions are disabled, hide the usage
of std::exception_ptr etc. when compiling with
CATCH_CONFIG_DISABLE_EXCEPTIONS.


horenmar:
I would prefer if all platforms provided stubbed out exception-related functions, but I _do_ support `-fno-except`. Oh well.

Thanks.

PR #1958, opened 2020-06-11
Add OverallResultsCases element to XML reporter

rpavlik:
## Description

This adds a small additional output to the XML reporter: the number of passed/failed/expected-failed test *cases* and not just assertions. I am working with a team that is building somewhat of a system test, rather than unit test, with Catch2, and some assertions are called in a loop making the total assertion count not as helpful as the total test case count.

The approval tests appear to fail but for apparently unrelated reasons?


horenmar:
The approval tests fail for a reason that is very much related. They serve to check that the reporter output looks the way it should, which obviously does not match with the new changes. The reason the failure looks weird is that the checking script then attempts to print out diff of the incongruities, runs into a character it cannot print to the console and then dies.

horenmar:
The changes themselves look reasonable.

rpavlik:
Ah, thanks for interpreting the test results for me. I'll see if I can add that change to the expected output.

horenmar:
To update the baselines, run `scripts/approve.py` and check that the resulting diff makes sense.

PR #1960, opened 2020-06-18
Make scripts/updateDocumentToC.py executable.

richardash1981:
## Description
On systems where the file system has excute permissions, this script was not marked as executable in a clean git checkout and so could not be run without first changing the permissions.

Fixed by setting the relevant git flag.


horenmar:
Thanks. I work primarily of Windows or WSL, both of which happily execute files without x bit.

PR #1961, opened 2020-06-19
Add notes on compiling the examples.

richardash1981:
## Description
This took me some time to figure out, so document for others.
Adding a section to the contributor guide seems like the correct place to me.


horenmar:
Apart from the spurious tab at the start of the file, lgtm.

I will be reorganizing and rewriting the contributing docs a bit, but this is a good addition.

PR #1963, opened 2020-06-19
Add an example of using GENERATE(table())

richardash1981:
## Description
There are some examples on issue #850 of using this feature, but they are not easily found from the documentation. Adding them here as an example makes them more findable and ensures they keep working if the API changes.

### DIFF
--- a/docs/list-of-examples.md
+++ b/docs/list-of-examples.md
@@ -16,6 +16,7 @@
 - Configuration: [Provide your own output streams](../examples/231-Cfg-OutputStreams.cpp)
 - Generators: [Create your own generator](../examples/300-Gen-OwnGenerator.cpp)
 - Generators: [Use map to convert types in GENERATE expression](../examples/301-Gen-MapTypeConversion.cpp)
+- Generators: [Run test with a table of input values](../examples/302-Gen-Table.cpp)
 - Generators: [Use variables in generator expressions](../examples/310-Gen-VariablesInGenerators.cpp)
 - Generators: [Use custom variable capture in generator expressions](../examples/311-Gen-CustomCapture.cpp)
 
--- a/examples/302-Gen-Table.cpp
+++ b/examples/302-Gen-Table.cpp
@@ -1,0 +1,54 @@
+// 302-Gen-Table.cpp
+// Shows how to use table to run a test many times with different inputs. Lifted from examples on
+// issue #850.
+
+#include <catch2/catch.hpp>
+#include <string>
+
+struct TestSubject {
+    // this is the method we are going to test. It returns the length of the
+    // input string.
+    size_t GetLength( const std::string& input ) const { return input.size(); }
+};
+
+
+TEST_CASE("Table allows pre-computed test inputs and outputs", "[example][generator]") {
+    using std::make_tuple;
+    // do setup here as normal
+    TestSubject subj;
+
+    SECTION("This section is run for each row in the table") {
+        std::string test_input;
+        size_t expected_output;
+        std::tie( test_input, expected_output ) =
+            GENERATE( table<std::string, size_t>(
+                { /* In this case one of the parameters to our test case is the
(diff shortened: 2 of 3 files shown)


richardash1981:
That's weird. It failed on the AMD 64 / Linux / Xenial COMPILER='clang++-3.8' EXAMPLES=1 COVERAGE=1 EXTRAS=1 run in Travis.
That's fair enough in that my tests were AMD 64 / Linux / GCC 9.3.0.
Clang versions are confusing me, because the one I have available is clang++ 9.0.1. Clang 3.8 is a really old version by now?

This is the error compiling the example
```
[ 26%] Building CXX object examples/CMakeFiles/302-Gen-Table.dir/302-Gen-Table.cpp.o
/home/travis/build/catchorg/Catch2/examples/302-Gen-Table.cpp:31:59: error: no
      matching function for call to 'table'
  ...expected_output) = GENERATE( table<std::string, size_t>({
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/build/catchorg/Catch2/single_include/catch2/catch.hpp:4107:97: note: 
      expanded from macro 'GENERATE'
  ...namespace Catch::Generators; return makeGenerators( __VA_ARGS__ ); } ) /...
                                                         ^~~~~~~~~~~
/home/travis/build/catchorg/Catch2/single_include/catch2/catch.hpp:4058:41: note: 
      candidate function [with Ts = <std::__cxx11::basic_string<char>, unsigned
      long>] not viable: cannot convert initializer list argument to
      'std::tuple<std::__cxx11::basic_string<char>, unsigned long>'
    GeneratorWrapper<std::tuple<Ts...>> table( std::initializer_list<std...
                                        ^
/home/travis/build/catchorg/Catch2/examples/302-Gen-Table.cpp:31:49: error: no
      matching function for call to 'generate'
  ...expected_output) = GENERATE( table<std::string, size_t>({
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/build/catchorg/Catch2/single_include/catch2/catch.hpp:4105:5: note: 
      expanded from macro 'GENERATE'
    Catch::Generators::generate( INTERNAL_CATCH_STRINGIZE(INTERNAL_CATCH...
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/build/catchorg/Catch2/single_include/catch2/catch.hpp:4089:10: note: 
      candidate template ignored: substitution failure [with L = (lambda at
      /home/travis/build/catchorg/Catch2/examples/302-Gen-Table.cpp:31:49)]:
      member reference base type 'typename add_rvalue_reference<void>::type'
      (aka 'void') is not a structure or union
  ...generate( StringRef generatorName, SourceLineInfo const& lineInfo, L const& generatorExpression ) -> decltype(std::declval<decltype(generatorExpression())>()....
     ^                                                                                                                                                            ~
2 errors generated.
```
I'm guessing that "cannot convert initializer list argument to       `'std::tuple<std::__cxx11::basic_string<char>, unsigned long>'`" is the important error here, the rest will go away if that did not occur. However I can't find any other configuration of { which will compile with my current compilers ...

horenmar:
At a glance, this is likely the same problem as described in `Generators.tests.cpp` line 34 and below.

Also please avoid tabs (CMakeLists.txt for examples)

richardash1981:
Thanks for the pointers. Hopefully this version will build on all targets ...

richardash1981:
Thank you for the merge!
~~Was the re-format done by an automated tool or manually? In other words is there something I should have done to the code before pushing, or a coding standard I should have followed?~~ Ah, #1182. I need to learn about using clang-tidy ...

horenmar:
Both. Clang-format was just introduced, and I am still experimenting with the settings, so I sometimes override it manually.

I also updated the contributing docs to mention clang-format.

PR #1971, opened 2020-06-29
Fix for macOS on ARM

grhoten:
These changes resolve compilation issues for the Catch headers when compiling with macOS on ARM, which is not x86/x86_64 based. The changes are similar to the "elif defined(CATCH_PLATFORM_IPHONE)" logic later in these files.  ARM based macOS machines were already announced at Apple's WWDC.

The error being resolved looks like the following:
```
catch.hpp:8237:13: error: invalid token in expression
            CATCH_BREAK_INTO_DEBUGGER();
            ^
catch.hpp:7948:83: note: expanded from macro 'CATCH_BREAK_INTO_DEBUGGER'
        #define CATCH_BREAK_INTO_DEBUGGER() []{ if( Catch::isDebuggerActive() ) { CATCH_TRAP(); } }()
                                                                                  ^
catch.hpp:7913:34: note: expanded from macro 'CATCH_TRAP'
    #define CATCH_TRAP() __asm__("int $3\n" : : ) /* NOLINT */
                                 ^
<inline asm>:1:6: note: instantiated into assembly here
        int $3
            ^
```

### DIFF
--- a/include/internal/catch_debugger.h
+++ b/include/internal/catch_debugger.h
@@ -17,7 +17,11 @@ namespace Catch {
 
 #ifdef CATCH_PLATFORM_MAC
 
-    #define CATCH_TRAP() __asm__("int $3\n" : : ) /* NOLINT */
+    #if defined(__i386__) || defined(__x86_64__)
+        #define CATCH_TRAP() __asm__("int $3\n" : : ) /* NOLINT */
+    #elif defined(__aarch64__)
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks for the patch. Do note, that as per the [contributing docs](https://github.com/catchorg/Catch2/blob/master/docs/contributing.md) files in the `single_include` should not be changed as part of a PR.

grhoten:
I have reverted the change to single_include.

PR #1973, opened 2020-07-05
Escaping "*" ("times") to fix Markdown presentation

echuber2:
## Description

If you want to write "2\*2" you need to escape the asterisk, or else it becomes formatting in Markdown.

## GitHub Issues

None that I know of.

### DIFF
--- a/docs/generators.md
+++ b/docs/generators.md
@@ -24,9 +24,9 @@ TEST_CASE("Generators") {
 }
 ```
 
-The `SECTION` "one" will be run 4 (2*2) times, because the outer
+The `SECTION` "one" will be run 4 (2\*2) times, because the outer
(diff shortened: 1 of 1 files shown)


horenmar:
AFAIK it currently renders fine, but sure, why not.

echuber2:
Thanks! You can see what it looked like previously here (on the GitHub web UI). The two asterisks within the same paragraph rendered that section as italic instead of actually showing the asterisks.

https://github.com/catchorg/Catch2/blob/c299133a31fb8594a5bbacc934cf30b7aabda87e/docs/generators.md

horenmar:
Sorry. I checked the rendered page before making the comment, but thinking about it I checked it with your diff, rather than before. 🤦‍♂️

PR #1976, opened 2020-07-06
Better support for stringifying pointer types

stefanhaller:
When stringifying a pointer to a type, and there's an ostream operator<< for
that type, use it instead of printing a hex address.

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -279,7 +279,26 @@ namespace Catch {
     template <typename T>
     struct StringMaker<T*> {
         template <typename U>
-        static std::string convert(U* p) {
+        static typename std::enable_if<
+            ::Catch::Detail::IsStreamInsertable<U>::value,
+            std::string>::type
+        convert(U* p) {
+            if (p) {
+                ReusableStringStream rss;
+                // NB: call using the function-like syntax to avoid ambiguity with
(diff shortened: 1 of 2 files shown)


horenmar:
No, absolutely not.

`operator==` of pointer types compares pointer values and not underlying types.

stefanhaller:
OK, thanks for the friendly reaction. 😄 

I should have said more about the use case that I need this for. It's very similar to how you would use a `std::optional`: you can compare a `std::optional` to the contained type, and it will compare equal if it is non-null and the contained instance matches. This is very convenient because it allows you to express your expectations in one line instead of two:
```
std::optional<int> o;
// ...
CHECK(o == 5);
```
instead of
```
REQUIRE(o.has_value());
CHECK(*o == 5);
```
And catch2 very helpfully supports this case by printing the contained instance of the optional if the comparison fails.

I want to do the same thing with pointers, which have very similar semantics:
```
int *p = 7;
// ...
CHECK(p == 5);
```
Now, for this to work I have to define my own `operator==(int*, int)`, which you might consider a hack, and I sort of agree, but it works well and is convenient. (In my actual use case the type is not int, but a class.)

I realize that even after this explanation it's unlikely that you will change your mind, but I'd very much appreciate some help as to how else I can make this work. I guess I'll have to add my own specialization of `StringMaker` for `MyType*`?

Issue #1984, opened 2020-07-16
Unable to find executable under CMake 3.18


NeroBurner:
**Describe the bug**
When upgrading CMake to 3.18.0 Catch2 tests are not found anymore. They get compiled, but cannot be executed. With CMake 3.17.3 everything works as expected

**Expected behavior**
test executables should be found and executed, also on CMake 3.18.0

**Reproduction steps**
Steps to reproduce the bug.

install CMake 3.18.0
create a minimal example using Catch2, for example https://github.com/NeroBurner/Catch2_minimial_example_cmake_3_18

```cpp
#define CATCH_CONFIG_MAIN  // This tells Catch to provide a main() - only do this in one cpp file

// external libraries in this repository
#include <catch2/catch.hpp>

TEST_CASE("all is good")
{
    REQUIRE(true);
}
```

```sh
cmake -H. -B_build
cmkae --build _build
cmake --build _build --target test
```

Log with CMake 3.18.0
```sh
-- The CMake version used is 3.18.0
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- folder: Catch2_CMake_3_18_test
-- Set default build type to Release
-- Configuring done
-- Generating done
-- Build files have been written to: /home/nero/repos/external/Catch2_minimial_example_cmake_3_18/_build
Scanning dependencies of target test_yes
[ 50%] Building CXX object CMakeFiles/test_yes.dir/test_yes.cpp.o
[100%] Linking CXX executable test_yes
[100%] Built target test_yes
Running tests...
Test project /home/nero/repos/external/Catch2_minimial_example_cmake_3_18/_build
CMake Warning (dev) at CTestTestfile.cmake:7:
  Syntax Warning in cmake code at column 12

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CTestTestfile.cmake:8:
  Syntax Warning in cmake code at column 24

  Argument not separated from preceding token by whitespace.
[...]

PR #1985, opened 2020-07-16
Fix CMake add test helper for CMake 3.18.0

NeroBurner:
## Description
With CMake 3.18.0 the `add_test(NAME)` handling changed. The escaped
double quotes confuse the new call. Work around this upstream change.

## GitHub Issues
Closes: https://github.com/catchorg/Catch2/issues/1984


horenmar:
Thanks for the investigation and fix.

PR #1986, opened 2020-07-19
Don't apply global settings when configuring as a subproject.

Bekenn:
## Description
When adding Catch2 as a subdirectory, a couple of choices in the main CMakeLists.txt file end up having a global effect:
* The `USE_FOLDERS` property is global; setting it affects the entire workspace.
* Including the `CTest` module will cause the generation of several new targets.

These global choices should be decided by the consuming project, not by one of its dependencies.  This change moves those settings into regions  guarded by `if(NOT_SUBPROJECT)`.  This should have no effect on Catch2 standalone builds, while being friendlier to consuming projects.

## GitHub Issues
I'm not aware of any existing issues relevant to this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.5)
 # detect if Catch is being bundled,
 # disable testsuite in that case
 if(NOT DEFINED PROJECT_NAME)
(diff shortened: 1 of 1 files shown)


horenmar:
That seems reasonable, thanks.

PR #1993, opened 2020-07-31
Update ParseAndAddCatchTests.cmake for cmake 3.18.1

knzivid:
## Description

https://cmake.org/cmake/help/v3.18/release/3.18.html#id1

In CMake 3.18.0 the add_test() command learned to support special characters in test names. This was accidentally left out of its release notes. Unfortunately the fix breaks existing projects that were using manual quoting or escaping to work around the prior limitation. This fix has been reverted in 3.18.1, but may be re-introduced in future versions of CMake with a policy for compatibility.


horenmar:
Thanks, that change was a giant mess.

PR #1996, opened 2020-08-03
Issue 1992: Better c++17 std::byte detection

### LINKED ISSUE #1992: C++17 std::byte detection fails on ubuntu+clang-6.0
**Describe the bug**
Clang6 and clang8 in c++17 mode cannot compile catch unit tests 

**Expected behavior**
C++17 should be supported

**Reproduction steps**
Ubuntu 16, clang++6.0 (and clang++-8) installed with "apt install clang6.0"
Simple null unit test:
```
#include "catch.hpp"
TEST_CASE("simple", "[simple]") {
}
```
```
$ clang++-6.0 -std=c++17 -c cb.cpp 
In file included from cb.cpp:1:
./catch.hpp:1735:29: error: no member named 'byte' in namespace 'std'
    struct StringMaker<std::byte> {
                       ~~~~~^
1 error generated.
```

**Platform information:**

 - OS: ubuntu 16
 - Compiler+version: 
clang version 6.0.0-1ubuntu2~16.04.1 (tags/RELEASE_600/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

 - Catch version: v2.13.0 downloaded in the single file version

**Additional context**
Also happens with clang-8 package.
Compiles fine in clang-6 or clang-8 in c+11 or c++14 mode.

This may be a packaging error in ubuntu, or it might be I have not installed some vital piece, or need to configure clang to use a better standard library or something.

gnbond:
This fixed buld errors for ubuntu-16 + clang and similar situations
where C++17 is supported in the compiler but not the default
C++ standard library.

## Description
Clang packages on ubuntu-16 use the clang compiler which supports C++17, but use the system-default C++ library which does not fully support C++17.  This results in compile errors where Catch2 mis-identifies support for std::bytes.

## GitHub Issues
Closes #1992

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -240,7 +240,10 @@
 
   // Check if byte is available and usable
   #  if __has_include(<cstddef>) && defined(CATCH_CPP17_OR_GREATER)
-  #    define CATCH_INTERNAL_CONFIG_CPP17_BYTE
+  #    include <cstddef>
+  #    if __cpp_lib_byte > 0
+  #      define CATCH_INTERNAL_CONFIG_CPP17_BYTE
+  #    endif
   #  endif // __has_include(<cstddef>) && defined(CATCH_CPP17_OR_GREATER)
 
   // Check if variant is available and usable


horenmar:
Thanks

Issue #2003, opened 2020-08-13
Detail::rangeToString doesn't support c++20 ranges with sentinels


seanmiddleditch:
**Describe the bug**
The function `Detail::rangeToString` (https://github.com/catchorg/Catch2/blob/c190061001739e8bf240a78059f1a3e927a35efc/include/internal/catch_tostring.h#L313) is defined as:

```c++
        template<typename InputIterator>
        std::string rangeToString(InputIterator first, InputIterator last) {
```

This breaks for C++20 ranges or user-defined C++17 types that make use of a sentinel type for a container's `.end()` return, e.g. `counted_iterator` and `default_sentinel_t` in the latest ranges code on the MSVC STL.

The fix in this case is super easy: use two template types.

```c++
        template<typename InputIterator, typename Sentinel = InputIterator>
        std::string rangeToString(InputIterator first, Sentinel last) {
```

Actual error received in user code (for a custom `zstring_view` type that uses a sentinel type for `end()`):

```c++
C:\Projects\potato\deps\catch2-src\single_include\catch2\catch.hpp(2023): error C2782: 'std::string Catch::Detail::rangeToString(InputIterator,InputIterator)': template parameter 'InputIterator' is ambiguous
  C:\Projects\potato\deps\catch2-src\single_include\catch2\catch.hpp(1833): note: see declaration of 'Catch::Detail::rangeToString'
  C:\Projects\potato\deps\catch2-src\single_include\catch2\catch.hpp(2023): note: could be 'up::zstring_view::const_sentinel'
  C:\Projects\potato\deps\catch2-src\single_include\catch2\catch.hpp(2023): note: or       'up::zstring_view::const_iterator'
C:\Projects\potato\deps\catch2-src\single_include\catch2\catch.hpp(2023): error C2784: 'std::string Catch::Detail::rangeToString(InputIterator,InputIterator)': could not deduce template argument for 'InputIterator' from 'up::zstring_view::const_sentinel'
```

**Expected behavior**

Code using sentinels to compile.

**Reproduction steps**

Steps to reproduce the bug.

```c++
#include <catch2/catch.hpp>

// bogus range, but illustrates problem
struct test {
    using const_iterator = int*;
    using const_sentinel = std::nullptr_t;

    const_iterator begin() const { return nullptr; }
    const_sentinel end() const { return nullptr; }
    bool operator==(int) const { return false; }
};

TEST_CASE("sentinel") {
    CHECK(test{} == 0);
}
```

[...]

PR #2004, opened 2020-08-13
Support sentinel-based ranges in default stringify

seanmiddleditch:
Fix for issue #2003 

- Enhances `rangeToString` to work with a separate iterator and sentinel type
- Adds a test to ensure this compiles

### DIFF
--- a/include/internal/catch_tostring.h
+++ b/include/internal/catch_tostring.h
@@ -309,8 +309,8 @@ namespace Catch {
 #endif
 
     namespace Detail {
-        template<typename InputIterator>
-        std::string rangeToString(InputIterator first, InputIterator last) {
+        template<typename InputIterator, typename Sentinel = InputIterator>
+        std::string rangeToString(InputIterator first, Sentinel last) {
             ReusableStringStream rss;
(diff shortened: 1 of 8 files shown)


horenmar:
Tests in default run cannot fail, or it will fail CI. Instead, you should use `::Catch::Detail::stringify` in the test and check that the return value matches expected stringification.

seanmiddleditch:
Ah, didn't expect the tests to fail, actually. Looking it over though it makes sense, though I'm not sure why I didn't experience that locally.

seanmiddleditch:
@horenmar - I'm not familiar with the approval test stuff here; should I be checking in the files that are modified locally on build?

horenmar:
On a second look, this part seems to be missing [from the documentation](https://github.com/catchorg/Catch2/blob/master/docs/contributing.md#testing-your-changes) after the rewrite. 🤔 

Approval tests diff known good output with current output. If there are changes, check that the diff makes sense, run `scripts/approve.py` and commit the updated baselines.

horenmar:
Looks good now, thanks.

PR #2007, opened 2020-08-24
Remove superfluous values

dawikur:
## Description

The `0`s were needed for the expansion of parameter pack for the sake of
`expander`. Now when we have `index++` it is no more needed.


horenmar:
lgtm, thanks.

Issue #1821, opened 2019-12-20
Conditional SECTION


Resurr3ction:
**Description**
To have a `SECTION` variant `SECTION_IF` that would be only executed if condition was true:

```
#define SECTION_IF(cond, ...)
```
The motivational example:
```
TEST_CASE("MyVector")
{
    auto vector = GENERATE(std::vector<int>{}, std::vector<int>{1,2,3,4,5});

    SECTION_IF(!vector.isEmpty(), "operator[]")
    {
        REQUIRE(vector[0] == 1);
    }
}
```
Currently I can skip code inside a section with `if` but the section will still cause the whole test to be needlessly restarted even if it is not relevant for given generated input (e.g. empty vector). Another option would be to have two separate test cases but since the tests share a lots of code it that is less than ideal. Same as putting the generators or initialization code in only the relevant sections. Too much code duplication.

 Instead I resorted to the following workaround:
```
if(condition) 
{
    SECTION("MySection")
    {
    }
}
```
That seems to work but there is a bug report #714 that seems to have issues with this approach (might be outdated though) so this might be buggy way of doing things.

I hope this can be added seemelssly. Not sure if this would greatly upset how SECTIONs are handled by Catch2.

free-easy:
I've tested with the newest version of v2.11.1 (d10b9bd02e098476670f5eb0527d2c7281476e8a) of Catch2 and the bug #714 is still there.

PR #2018, opened 2020-09-05
Introduced SECTION_IF to run sections conditionally.

dadocolussi:
## Description

This feature introduces ```SECTION_IF``` to run sections conditionally. This is useful with test cases that use generators.

Consider a test case with generated values, a long setup, and a number of small sections. ```SECTION_IF``` reduces code duplication and makes test cases more legible.

```
TEST_CASE("Long test with generated values and a section depending on the generated value")
{
    auto i = GENERATE(2, 3, 5, 7, 11);
    // long setup here
    SECTION("test something for all i") {}
    SECTION("test something else for all i") {}
    SECTION_IF(i == 11, "test something when i is 11 only") {}
}
```

## GitHub Issues

This feature addresses issue #1821.


benldrmn:
@horenmar I noticed you deleted `catchorg:master` (the branch this feature was merged into), and I couldn't find `SECTION_IF` in the current master branch. Is this feature planned to be a part of an upcoming Catch2 release?

PR #2023, opened 2020-09-12
Add REPORTER and OUTPUT_DIR args to catch_discover_tests

### LINKED ISSUE #1946: Generate JUnit report using catch_discover_tests
**Description**
When using the catch2_discover_tests CMake function, it is currently not possible to generate a JUnit report for each test (i think). 

My project CMakeLists.txt file contains the following

```
catch_discover_tests(
${PROJECT_NAME}_tests
EXTRA_ARGS -r junit -o report.xml
)
```

This will generate "report.xml" for each test scenario and overwrite the previous "report.xml"

Is there a way to generate a JUnit report for all tests when using catch_discover_tests?

globberwops:
## Description

When using `catch_discover_tests` with `EXTRA_ARGS -o foo`, the test reports will be overwritten with each test case.

The args `REPORTER`, `OUTPUT_DIR`, `OUTPUT_PREFIX`, and `OUTPUT_SUFFIX` will allow for further customization of the test case runs.
* `REPORTER` can be any valid reporter of the Catch executable. 
  The specified reporter will be validated against the available reporters of the Catch executable. 
* `OUTPUT_DIR` can be a relative or an absolute path.
  If relative, it will be relative to the `WORKING_DIRECTORY`, which in turn defaults to `CMAKE_CURRENT_BINARY_DIR`.
* `OUTPUT_PREFIX` is a string that will be added before each test report name.
* `OUTPUT_SUFFIX` is a string that will be added after each test report name, e.g. a file extension.

## GitHub Issues

Resolves  #1946

### DIFF
--- a/contrib/CatchAddTests.cmake
+++ b/contrib/CatchAddTests.cmake
@@ -6,6 +6,10 @@ set(suffix "${TEST_SUFFIX}")
 set(spec ${TEST_SPEC})
 set(extra_args ${TEST_EXTRA_ARGS})
 set(properties ${TEST_PROPERTIES})
+set(reporter ${TEST_REPORTER})
+set(output_dir ${TEST_OUTPUT_DIR})
+set(output_prefix ${TEST_OUTPUT_PREFIX})
+set(output_suffix ${TEST_OUTPUT_SUFFIX})
 set(script)
 set(suite)
 set(tests)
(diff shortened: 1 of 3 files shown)


horenmar:
A quick note: don't try and enumerate valid reporters up-front, because users might have their own reporters they want to use. It also currently only lists the built-in reporters, but not the ones that Catch2 also provides as part of release (e.g. TAP and TeamCity reporters).

Since you are modifying `catch_discover_tests`, which already expects to be able to run the test executable, you can instead use `--list-reporters` and handle its output while validating the argument.

globberwops:
@horenmar you are totally right, I have added using `--list-reporters` now. Thanks for the feedback!

contrib/CatchAddTests.cmake
```diff
@@ -48,6 +50,47 @@ endif()
...
+  )
+endif()
+
+# Prepare report
+set(report_extension ".txt")
+if(reporter)
+  set(reporter_arg "-r ${reporter}")
+  if(reporter MATCHES "junit|xml")
+    set(report_extension ".xml")
+  endif()
```
horenmar, line 83: I am not quite sold on this -- these two aren't the only Catch2-provided reporters that produce XML output, and in general it seems very limited (if potentially useful).

I think the ideal solution would be to let the user set a filename pattern, like `output-{date}-{testname}.json`, but that brings in _a lot_ more complexity.

I'll have to think about this a bit more.

contrib/CatchAddTests.cmake
```diff
@@ -48,6 +50,47 @@ endif()
...
+if(reporter AND ${reporter_is_valid} EQUAL -1)
+  message(FATAL_ERROR
+    "\"${reporter}\" is not a valid reporter!\n"
+  )
+endif()
+
+# Prepare report
+set(report_extension ".txt")
+if(reporter)
+  set(reporter_arg "-r ${reporter}")
```
horenmar, line 80: Small stylistic nitpick: prefer the long option names in scripts. (Same goes for `-o` below)

globberwops: Done 👌

contrib/CatchAddTests.cmake
globberwops: I could always make `OUTPUT_DIR` a multi value keyword, optional 2nd value would be the file extension, defaulting to none:
```
catch_discover_tests(foo REPORTER junit OUTPUT_DIR test-results xml)
```

Let me know what you think.

horenmar: Agreed with 4) being the best option.

docs/cmake-integration.md
```diff
@@ -83,6 +83,8 @@ catch_discover_tests(target
                      [TEST_SUFFIX suffix]
                      [PROPERTIES name1 value1...]
                      [TEST_LIST var]
+                     [REPORTER reporter]
+                     [OUTPUT_DIR dir]
 )
```
horenmar, line 88: This is missing the prefix and suffix args.

horenmar:
Otherwise LGTM. Can you rebase + squash this into atomic commits?

docs/cmake-integration.md
globberwops: Fixed, thanks!

horenmar:
Thanks

PR #2028, opened 2020-09-18
Docu/Show how to use CMake FetchContent

blackliner:
## Description
Since CMake 3.11 a new feature called FetchContent is available, that makes it easy to integrate external projects.

https://cmake.org/cmake/help/latest/module/FetchContent.html#fetch-content-examples


horenmar:
Thanks, lgtm

PR #2031, opened 2020-09-23
Support template test cases in ParseAndAddCatchTests

### LINKED ISSUE #1848: ParseAndAddCatchTests.cmake silently ignores test that contains right parenthesis
**Describe the bug**
The https://github.com/catchorg/Catch2/blob/master/contrib/ParseAndAddCatchTests.cmake script
does not handle when a test name contains a right parenthesis. For example

> TEST_CASE("Test ) abc")
> {
> }

This is silently ignored and the test is never run. 

**Expected behavior**
Expected that the test is run. An alternative is that the script returns with an error signalling that the test name was invalid.

**Platform information:**
OS Ubuntu 18.04
Cmake version 3.15.5
Tested on newest master 2d172dc688ffab0867a94928159e93384475a870

agserm:
* Change regex to allow parentheses inside the test macro for a type list
* Append a wildcard to the CTestName if the test case is a template
* Also change the regular expression so parentheses are allowed in names
  (fixes #1848)

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -1,9 +1,11 @@
 #==================================================================================================#
 #  supported macros                                                                                #
 #    - TEST_CASE,                                                                                  #
+#    - TEMPLATE_TEST_CASE                                                                          #
 #    - SCENARIO,                                                                                   #
 #    - TEST_CASE_METHOD,                                                                           #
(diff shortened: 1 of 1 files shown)


horenmar:
> Also change the regular expression so parentheses are allowed in names

I am going to take you on your word with that, because that regex is too much.

Anyway, thanks.

PR #2032, opened 2020-09-23
add fuzzer for TestSpecParse

pauldreik:
This adds a fuzzer for the TestSpecParser. I ran it for a few minutes, and it found nothing (unusual!).

This addition should not affect anyone, fuzzing is disabled by default.

I do not want to add more fuzzers, scripts, documentation etc at this point because I want to have early feedback. For instance, the cmake I wrote is probably horrible and there is some better way to do it.

Also, it is libFuzzer only. Making it oss-fuzz compatible can easily be done later (I have done that with simdjson and fmt).


fuzzing/build_fuzzers.sh
```diff
@@ -0,0 +1,25 @@
...
+cmake $CATCHROOT \
+  -DCMAKE_CXX_COMPILER=clang++ \
+  -DCMAKE_CXX_FLAGS="-fsanitize=fuzzer-no-link,address,undefined -O3 -g" \
+  -DCATCH_BUILD_EXAMPLES=Off \
+  -DCATCH_BUILD_EXTRA_TESTS=Off \
+  -DCATCH_BUILD_TESTING=Off \
+  -DBUILD_TESTING=Off \
+  -DCATCH_BUILD_FUZZERS=On
+
+cmake --build . -j 4
```
horenmar, line 24: The argument to `-j` should be the result of `nproc`.

pauldreik: fixed

horenmar:
You should target the `dev-v3` branch. Its cmake is significantly different, it would remove the need to declare the library files again, and it is about one good day of work away from being the main branch (I want to rewrite its contributing.md first).

horenmar:
Also fuzzers should then live under tests/

fuzzing/fuzz_XmlWriter.cpp
```diff
@@ -0,0 +1,33 @@
...
+  NulOStream() : std::ostream(this) {}
+  NulStreambuf *rdbuf() { return this; }
+};
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+
+    std::string buf(Data,Data+Size);
+    NulOStream nul;
+    Catch::XmlEncode encode(buf);
+    encode.encodeTo(nul);
```
horenmar, line 31: I am reasonably sure that this needs an XML validator oracle to catch anything.

pauldreik: To know that the xml it outputs is correct? Yes. But I just wanted to exercise the code a bit, I browsed through the cpp files and that was the only one that looked scary.

pauldreik:
> Also fuzzers should then live under tests/

I am not a fan of this, that usually implies having to build the unit tests, right?

pauldreik:
Made this on top of v3, but I can not change the branch to merge to. Sorry for my bad github fu. Closing this and opening a new PR.

PR #2033, opened 2020-09-23
add fuzzer for TestSpecParser and XmlWriter

pauldreik:
This adds a fuzzer for the TestSpecParser and XmlWriter. I ran them for a few minutes, and it found nothing (unusual!).

This addition should not affect anyone, fuzzing is disabled by default.

I do not want to add more fuzzers, scripts, documentation etc at this point because I want to have early feedback. For instance, the cmake I wrote is probably horrible and there is some better way to do it.

Also, it is libFuzzer only. Making it oss-fuzz compatible can easily be done later (I have done that with simdjson and fmt).

### DIFF
--- a/fuzzing/CMakeLists.txt
+++ b/fuzzing/CMakeLists.txt
@@ -7,11 +7,14 @@
 add_library(fuzzhelper NullOStream.h NullOStream.cpp)
 target_link_libraries(fuzzhelper PUBLIC Catch2::Catch2)
 
+# use C++17 so we can get string_view
+target_compile_features(fuzzhelper PUBLIC cxx_std_17)
+
 # This should be possible to set from the outside to be oss-fuzz compatible,
 # fix later. For now, target libFuzzer only.
 target_link_options(fuzzhelper PUBLIC "-fsanitize=fuzzer")
 
-foreach(fuzzer TestSpecParser XmlWriter)
+foreach(fuzzer TestSpecParser XmlWriter textflow)
 add_executable(fuzz_${fuzzer} fuzz_${fuzzer}.cpp)
 target_link_libraries(fuzz_${fuzzer} PRIVATE fuzzhelper)
 endforeach()
--- a/fuzzing/CMakeLists.txt
+++ b/fuzzing/CMakeLists.txt
@@ -1,0 +1,17 @@
+# License: Boost 1.0
+# By Paul Dreik 2020
+
+# add a library that brings in the main() function from libfuzzer
+# and has all the dependencies, so the individual fuzzers can be
+# added one line each.
+add_library(fuzzhelper NullOStream.h NullOStream.cpp)
+target_link_libraries(fuzzhelper PUBLIC Catch2::Catch2)
+
+# This should be possible to set from the outside to be oss-fuzz compatible,
+# fix later. For now, target libFuzzer only.
+target_link_options(fuzzhelper PUBLIC "-fsanitize=fuzzer")
+
+foreach(fuzzer TestSpecParser XmlWriter)
+add_executable(fuzz_${fuzzer} fuzz_${fuzzer}.cpp)
+target_link_libraries(fuzz_${fuzzer} PRIVATE fuzzhelper)
+endforeach()
--- a/fuzzing/NullOStream.h
+++ b/fuzzing/NullOStream.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <iostream>
+
+// from https://stackoverflow.com/a/8244052
+class NullStreambuf : public std::streambuf {
+  char dummyBuffer[64];
+
+protected:
+  virtual int overflow(int c) override final;
+};
+
+class NullOStream final : private NullStreambuf, public std::ostream {
+public:
+  NullOStream() : std::ostream(this) {}
+  NullStreambuf *rdbuf() { return this; }
+  virtual void avoidOutOfLineVirtualCompilerWarning();
+};
+
+
--- a/fuzzing/NullOStream.h
+++ b/fuzzing/NullOStream.h
@@ -1,6 +1,7 @@
 #pragma once
 
-#include <iostream>
+#include <ostream>
+#include <streambuf>
 
 // from https://stackoverflow.com/a/8244052
 class NullStreambuf : public std::streambuf {
@@ -16,5 +17,4 @@ public:
   NullStreambuf *rdbuf() { return this; }
   virtual void avoidOutOfLineVirtualCompilerWarning();
 };
-
 
--- a/fuzzing/build_fuzzers.sh
+++ b/fuzzing/build_fuzzers.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+#
+# Builds the fuzzers
+#
+# By Paul Dreik 20200923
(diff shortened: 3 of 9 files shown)


fuzzing/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# License: Boost 1.0
+# By Paul Dreik 2020
+
+set(INC ${SOURCES_DIR})
+message(STATUS "inc is ${INC}")
+add_executable(fuzz_TestSpecParser fuzz_TestSpecParser.cpp)
+target_link_libraries(fuzz_TestSpecParser PRIVATE Catch2::Catch2)
+target_include_directories(fuzz_TestSpecParser PRIVATE ${INC})
```
horenmar, line 8: Is this needed?

horenmar:
On top of my head, there are 2 more areas that are brittle-ish and you could find bugs in there without specific validation of the results:
* Text formatting for stdout (`Column` and friends in TextFlow). There were some bugs here around handling text starting/ending with newlines, and having newlines at the column limit.
* Tag parsing in `TestCaseInfo`. As an optimization, tag parsing works by allocating large backing array, and then creating references into that array for the individual tags. However, the bugs here are unlikely to manifest just by constructing the type.

The reason why I don't expect fuzzing `XmlWriter` to find anything is that in general, Catch2 internals prioritize being foolproof to being fast. There are definitely areas where I intend to change it at some point, they are generally safe 😃

fuzzing/CMakeLists.txt
pauldreik: Which part did you think of? The status message, the include, the link, fuzzing testspecparser at all?

horenmar: Fair.

The manual include setting.

pauldreik: You are right, fixed it.

pauldreik:
>     * Text formatting for stdout (`Column` and friends in TextFlow). There were some bugs here around handling text starting/ending with newlines, and having newlines at the column limit.

I added a fuzzer in https://github.com/catchorg/Catch2/pull/2033/commits/4b063665d61dcad8dbc0ca2b16306c97ee8470f4, do you think it is sufficient to cover interesting cases of the Column/Columns classes?

fuzzing/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# By Paul Dreik 2020
+
+# add a library that brings in the main() function from libfuzzer
+# and has all the dependencies, so the individual fuzzers can be
+# added one line each.
+add_library(fuzzhelper NullOStream.h NullOStream.cpp)
+target_link_libraries(fuzzhelper PUBLIC Catch2::Catch2)
+
+# use C++17 so we can get string_view
+target_compile_features(fuzzhelper PUBLIC  cxx_std_17)
```
horenmar, line 11:
```suggestion
target_compile_features(fuzzhelper PUBLIC cxx_std_17)
```

That double space out of nowhere looked weird.

fuzzing/build_fuzzers.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+cmake $CATCHROOT \
+  -DCMAKE_CXX_COMPILER=$CXX \
+  -DCMAKE_CXX_FLAGS="-fsanitize=fuzzer-no-link,address,undefined -O3 -g" \
+  -DCATCH_DEVELOPMENT_BUILD=On \
+  -DCATCH_BUILD_EXAMPLES=Off \
+  -DCATCH_BUILD_EXTRA_TESTS=Off \
+  -DCATCH_BUILD_TESTING=Off \
+  -DBUILD_TESTING=Off \
```
horenmar, line 28: `-DBUILD_TESTING=OFF` shouldn't be needed, but there is no harm in keeping it I guess.

fuzzing/fuzz_TestSpecParser.cpp
```diff
@@ -0,0 +1,16 @@
+//License: Boost 1.0
+//By Paul Dreik 2020
+
+#include "catch2/internal/catch_test_spec_parser.hpp"
+#include "catch2/internal/catch_tag_alias_registry.hpp"
```
horenmar, line 5:
```suggestion
#include "catch2/internal/catch_tag_alias_registry.hpp"
```
```suggestion
#include <catch2/internal/catch_test_spec_parser.hpp>
#include <catch2/internal/catch_tag_alias_registry.hpp>
```

Just a style nitpick

fuzzing/fuzz_XmlWriter.cpp
```diff
@@ -0,0 +1,17 @@
+//License: Boost 1.0
+//By Paul Dreik 2020
+
+#include "catch2/internal/catch_xmlwriter.hpp"
```
horenmar, line 4:
```suggestion
#include <catch2/internal/catch_xmlwriter.hpp>
```

As above.

fuzzing/fuzz_textflow.cpp
```diff
@@ -0,0 +1,48 @@
+//License: Boost 1.0
+//By Paul Dreik 2020
+
+#include "catch2/internal/catch_textflow.hpp"
```
horenmar, line 4:
```suggestion
#include <catch2/internal/catch_textflow.hpp>
```

fuzzing/NullOStream.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <iostream>
```
horenmar, line 3:
```suggestion
#include <ostream>
#include <streambuf>
```

horenmar:
I still think that fuzzing belongs under tests, but I am willing to make that change myself later, so that it can be there and be included without having to enable tests.

Otherwise the fuzzers look like a reasonable start. 👍

fuzzing/build_fuzzers.sh
pauldreik: Anything that makes building faster is very welcome when developing fuzzers, since it make take a few tries to get the build scripts/compiler flags correct.

pauldreik:
It is good to get these fuzzers used, more fuzzers can always be added later. May I suggest adding a CI job for running them? That way, easy to find bugs never even get merged. I have used github actions for simdjson and storing corpus between runs on bintray. That works fine, it runs both on pull requests and on crontab. Recently, I added a similar setup to boost.json (which found several issues before they were merged!). There, I use github action but the storage between runs is made in github action cache. You can see the config here: https://github.com/CPPAlliance/json/blob/develop/.github/workflows/run_fuzzer.yml

I recommend the latter approach, I can more or less copy paste the structure from there (it took quite a while to get the caching to work as I wanted). But it requires that you enable github actions on this repository. Are you ok with this approach? If so, I will complete the PR with the github action script. 
I just got started on a new assignment, so it would be nice to get something working with the least effort possible for now.

horenmar:
GitHub Actions should already be enabled. I'll respond to the rest tomorrow.

horenmar:
I squashed and cleaned up the commits a bit. Looking forward to future improvements.

pauldreik:
Nice! Will have a go at adding an action later, probably around oct 17:th

PR #2039, opened 2020-10-01
Add WORKING_DIRECTORY to CatchAddTests.cmake commands

RemySphere:
## Description
In a CMake project, when using `catch_discover_tests(WORKING_DIRECTORY [...])` it usually means that the test binary needs resources that are relative to said working directory. This could lead to UB or a crash when running `--list-test-names-only` or `--list-reporters` on the produced test binary i.e when a .dll is not lazy loaded and not available in the current PATH.
This commit simply forwards the `WORKING_DIRECTORY` argument to the `execute_process()` calls.

This also remove the need of using CMake's own magic with `CMAKE_MSVCIDE_RUN_PATH` and the like...


horenmar:
That looks reasonable enough, thanks.

PR #2042, opened 2020-10-07
Fixed wrong script path

leozz37:
The path to the `generateAmalgamatedFiles.py` script is wrong on the documentation. There is no "tools" directory.


horenmar:
Have you looked in the branch the documentation is for?

https://github.com/catchorg/Catch2/tree/dev-v3/tools/scripts

Issue #1410, opened 2018-10-16
Exit code goes against Unix conventions - breaks basic assumptions


jamesfrullo:
## Description
When your run your unit test with "-l" or "--list-test-names-only", the exit code of the process is the number of tests. This is quite incorrect behavior on any Unix operating system. A non-zero exit code indicates the program could not perform its intended function. It is reasonable for a tool that runs the program to assume that something went wrong with listing tests, and indeed, I'm using Catch in a framework that assumes exactly that. Furthermore, exit codes above 128 are supposed to have special meaning, and the exit code may not exceed 255. See here, for example:
https://www.tldp.org/LDP/abs/html/exit-status.html

Please fix this to have the test runner exit 0 if there is, in fact, no error.

### Steps to reproduce
Simply build and run any catch unit test with "-l" or "--list-test-names-only".

### Extra information
* Catch version: v2.4.1
* Operating System: Linux
* Compiler+version: GCC 5.4

horenmar:
#1146 

The TL;DR version is that it won't be changed until next major version.

hpesoj:
This is not only a problem when using `-l`, but also when tests fail. I have left a comment on #1146.

Issue #1146, opened 2018-01-12
Bug: Abuse of the exit code


rakhimov:
## Description

It is *nasty* to abuse exit codes in environment where exit codes are specifically for **errors**.
It doesn't compose or play well with every other command line tools.
If you need to count things, have '--count' flag (e.g, git) that reports to stdout.

It doesn't seem to be documented currently,
but exit codes are limited in range to MaxExitCode (=255).
This is just another smell.

### Steps to reproduce

```bash
set -e
set -o pipefail

catch_main --list-test-names-only | wc -l
```

or

```
# No way to specify memory leak return code. Catch can return 127 as well!
valgrind --tool=memcheck --error-exitcode=127 catch_main
```
### Extra information

* Catch version: **v2.1**
* Operating System: **GNU/Linux**
* Compiler+version: **GCC, Clang**

horenmar:
There are basically three points in this

1) Returning the number of tests/reporters etc. in return code is weird and unexpected

This one I agree with, but because of backward compatibility is very likely here to stay.

2) Catch shouldn't use the number of failed tests as return code.

This one I don't agree with. Tests failing is an error. If you care that you cannot tell whether you have XYZ failing tests or are leaking memory, you care about the wrong problem.

3) Clamping the return code is weird.

Blame POSIX. Both `waitid` and `waitpid` only keep lowest 8 bits of process return value, so a return value in the multiple of 256 would lead to a false positive otherwise.


Anyway, 1) and 2) can [be easily solved by defining your own main function](https://github.com/catchorg/Catch2/blob/master/docs/own-main.md#let-catch-take-full-control-of-args-and-config), that returns different exit codes.

philsquared:
FTR I agree with everything @horenmar says. WRT (1) it was a case of "seemed like a good idea at the time".
I don't know if anyone really relies on it - but if they do it's not many. We should probably have changed it in the move to Catch2, but we could do it on the next major version?

And finally I'll add that (3) can also be addressed by providing your own `main()` - as long as your platform supports the unclamped values (e.g. Windows). I'm not sure how useful it really is in practice, though (if you have more than 255 tests failing do you really care about the exact number?)

Bu11etmagnet:
Returning nonzero when tests fail is very useful (I'd say required) when running tests from a build system.
Let's hope nobody will have 256 (or multiples of it) test failures, because it would look like success.

philsquared:
[...]

PR #2044, opened 2020-10-08
fix warning in extras/CatchAddTests.cmake

lkeegan:
## Description
  - don't warn on zero return code of `--list-reporters`
  - previously return code was the number of reporters (#1410, #1146)
  - as of 2c06ee9 return code is zero on success


horenmar:
Thanks

PR #2045, opened 2020-10-10
Add AppMesh to Open Source projects using Catch

laoshanxi:
## Description
Add opensource users [laoshanxi/app-mesh] which is using Catch2 for Unit Test.

## GitHub Issues
N/A


horenmar:
Thanks.

PR #2049, opened 2020-10-10
Prevent macro redefinition when using Clang with MSVC on Windows #2047

### LINKED ISSUE #2047: Using Clang on Windows fail due to redefined macro in catch_compiler_capabilities.hpp (v3.0.0-preview3)
**Describe the bug**
As per the title

**Expected behavior**
There are no redefined symbols.

**Reproduction steps**
1. Use Clang on Windows
2. Compile when including any of the Catch2 v3.0.0-preview3 headers

**Platform information:**
 - OS: **Windows NT**
 - Compiler+version: **Clang 10**
 - Catch version: **v3.0.0-preview3**

**Additional context**
The issue is line 62:
```
#if defined(__clang__)
```
which should probably be:
```
#if defined(__clang__) && !defined(_MSC_VER)
```

michaelvlach:
## Description

When using Clang on Windows the macro `CATCH_INTERNAL_START_WARNINGS_SUPPRESSION` gets redefined because both `__clang__` and `_MSC_VER` are set and both lead to a definition of this macro. This change is supposed to allow using Clang on Windows without such issue by adding additional condition to the clang's variant to prevent it taking effect when used with MSVC headers (clang-cl).

## GitHub Issues

Resolves #2047

### DIFF
--- a/src/catch2/internal/catch_compiler_capabilities.hpp
+++ b/src/catch2/internal/catch_compiler_capabilities.hpp
@@ -57,7 +57,7 @@
 
 #endif
 
-#if defined(__clang__)
+#if defined(__clang__) && !defined(_MSC_VER)
 
 #    define CATCH_INTERNAL_START_WARNINGS_SUPPRESSION _Pragma( "clang diagnostic push" )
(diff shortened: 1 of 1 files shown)


horenmar:
This is not a great fix, but it is a good _now_ fix. Thanks

Issue #2055, opened 2020-10-19
TEMPLATE_TEST_CASE from 2.13.2 breaks ParseAndAddCatchTest detection


NeroBurner:
**Describe the bug**
TEMPLATE_TEST_CASE from 2.13.2 breaks ParseAndAddCatchTest detection and no tests are added to `ctest`

**Expected behavior**
All previously working test cases should be found

**Reproduction steps**

create a minimal example using Catch2, for example https://github.com/NeroBurner/Catch2_minimial_example_cmake_3_18/tree/catch_2_13_2_template_fail

```cpp
#define CATCH_CONFIG_MAIN  // This tells Catch to provide a main() - only do this in one cpp file

// external libraries in this repository
#include <catch2/catch.hpp>

TEST_CASE("all is good")
{
    REQUIRE(true);
}
```

```cmake
cmake_minimum_required(VERSION 3.10...${CMAKE_VERSION})

project(Catch2_CMake_test)
enable_testing()
add_subdirectory(external/catch2 EXCLUDE_FROM_ALL)
add_executable(test_yes test_yes.cpp)
target_link_libraries(test_yes PRIVATE Catch2::Catch2)
ParseAndAddCatchTests(test_yes) # add_test with some magic
```

```sh
cmake -H. -B_build -DPARSE_CATCH_TESTS_VERBOSE=ON
cmkae --build _build
cmake --build _build --target test
```

In the configure step the `test_yes.cpp` file is parsed, but no test cases are found
```sh
-- ParseAndAddCatchTests: Started parsing test_yes
-- ParseAndAddCatchTests: Found the following sources: test_yes.cpp
-- ParseAndAddCatchTests: parsing /home/nero/repos/external/Catch2_minimial_example_cmake_3_18/test_yes.cpp
-- ParseAndAddCatchTests: Finished parsing test_yes
```

The last command with `--target test` (or `ctest`) returns
```
Test project /home/nero/repos/external/Catch2_minimial_example_cmake_3_18/_build_cmake_3.16.3
No tests were found!!!
```

**Platform information:**

 - OS: tested on **Ubuntu 18.04** and 20.04
 - Catch version: **v2.13.2**
 - CMake version: tested on 3.16.3, 3.18.4, 3.19-RC1

**Additional context**

In ParseAndAddCatchTest I tested the two regexes on the project, and only the old one extracted the `TEST_CASE()` successfully
```cmake
[...]

PR #2056, opened 2020-10-19
Fix CMake regex in ParseAndAddCatchTest helper

NeroBurner:
## Description

Fix regex that requires two string arguments in the form of
TEST_CASE("a", "b") resulting in not finding TEST_CASE("a") entries.

See regex: https://regex101.com/r/rQe6Jd/1

## GitHub Issues

Fixes: https://github.com/catchorg/Catch2/issues/2055

### DIFF
--- a/contrib/ParseAndAddCatchTests.cmake
+++ b/contrib/ParseAndAddCatchTests.cmake
@@ -108,7 +108,8 @@ function(ParseAndAddCatchTests_ParseFile SourceFile TestTarget)
     ParseAndAddCatchTests_RemoveComments(Contents)
 
     # Find definition of test names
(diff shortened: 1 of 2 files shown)


horenmar:
Honestly this calls for a proper parser, because the new regex breaks `TEMPLATE_TEST_CASE` instead:

https://regex101.com/r/rQe6Jd/2

NeroBurner:
Do you maybe also have an idea on how we could add tests for the parser, that such oversights are covered by the test cases?

horenmar:
I'd go with having a separate CMake project that uses the parser for a known source file, and a Python script that first configures the project, and then runs `ctest -N` and checks that the output lists all tests.

NeroBurner:
updated the regex to cover the template cases https://regex101.com/r/JygOND/1

horenmar:
Okay, looks fine.

However, I can't help but think of the regex, two problems saying. 😃

horenmar:
I just noticed that you target `master` branch. Don't, it is dead, I just kept it around for a while in case something broke too terribly after I changed the default branch.

For v2 branch target `v2.x`, for current development target `devel`.

Don't worry about the already merged PR, I'll cherry-pick it into the right branch later.

NeroBurner:
changed it, thanks for the heads up

PR #2057, opened 2020-10-20
Consider CMP0110 add_test() policy

NeroBurner:
## Description
CMake 3.19 introduces new add_test() behavior guarded with the policy
CMP0110.

See: https://cmake.org/cmake/help/latest/policy/CMP0110.html

Update the helper script ParseAndAddCatchTests to consider the policy and
handle the test case name accordingly.

## GitHub Issues
Related issue (CMake 3.18.0): https://github.com/catchorg/Catch2/issues/1984


horenmar:
That looks good, thanks.

Issue #421, opened 2015-05-23
Document How to make Catch tests in a library be found


martinmoene:
See Catch forum messages
- [Getting catch to work across multiple compilation units](https://groups.google.com/forum/?fromgroups=#!topic/catch-forum/pQchmF6Pr0E) by John Byrd
- [CATCH_CONFIG_RUNNER usage](https://groups.google.com/forum/#!msg/catch-forum/FV0Qo62DvgY/jxEO6c9_q3kJ) by Maxim

johnwbyrd:
On this topic... it would be useful to have a clearly documented method for adding test cases manually rather than depending on AutoReg to do so for you by instancing objects before main() is called.  Test fixtures as written depend on the AutoReg behavior, which I expect would not work for the reasons already given in this issue.

snortadmin:
Catch v1.2.1
 *  Generated: 2015-06-30 18:23:27.961086

Seems I'm running into the same or similar issue on both Linux and OSX building with g++ (not clang++) and autotools.  Here is the layout:

program
    lib1
        mod1a.cc
        mod1a_test.cc
        mod1b.cc
        ...
    lib2
        mod2a.cc
        ...

The lib# are convenience libs (built directly into the executable).

The above doesn't work because auto registration is apparently failing.  The work around is to add

    #include "mod1a_test.cc"

to the bottom of mod1a.cc.  Then the tests run just fine.

mod1a_test.cc is being compiled into lib1 but there is a difference in the symbols therein.  With the include there are 60 Catch symbols, w/o there are only 30.  The missing symbols are given below.

I'm not sure how else to convince the linker this stuff is needed.  2nd the suggestion that explicit registration be supported as a fallback.

1,30d0
< 0000000000002fae S __ZN5Catch11NameAndDescC1EPKcS2_
< 0000000000002fe4 S __ZN5Catch13AssertionInfoD1Ev
< 000000000000316c S __ZN5Catch13ExpressionLhsIRKiE17captureExpressionILNS_8Internal8OperatorE0EiEERNS_13ResultBuilderERKT0_
< 0000000000003142 S __ZN5Catch13ExpressionLhsIRKiEC1ERNS_13ResultBuilderES2_
< 000000000000311c S __ZN5Catch13ExpressionLhsIRKiEeqIiEERNS_13ResultBuilderERKT_
<                  U __ZN5Catch13ResultBuilder13endExpressionEv
<                  U __ZN5Catch13ResultBuilder13setResultTypeEb
[...]

PR #2062, opened 2020-10-21
V.2x bundle static lib

blackliner:
## Description
Starting of from (the somehow inactive PR) #1423:
1. add a new bazel target catch2_main
2. add documentation how to use bazel target

This targets the version 2 branch, because our project will need to stick with it for now.

oh, and first :D

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -10,7 +10,7 @@ cc_library(
 )
 
 cc_library(
-    name = "catch2_main",
+    name = "catch2_with_main",
     srcs = ["src/catch_with_main.cpp"],
     visibility = ["//visibility:public"],
     deps = ["//:catch2"],
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -3,8 +3,15 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
 
 # Header-only rule to export catch2/catch.hpp.
 cc_library(
-  name = "catch2",
-  hdrs = ["single_include/catch2/catch.hpp"],
-  visibility = ["//visibility:public"],
-  includes = ["single_include/"],
+    name = "catch2",
+    hdrs = ["single_include/catch2/catch.hpp"],
+    includes = ["single_include/"],
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "catch2_main",
+    srcs = ["src/catch_with_main.cpp"],
+    visibility = ["//visibility:public"],
+    deps = ["//:catch2"],
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,14 @@ endif()
 # provide a namespaced alias for clients to 'link' against if catch is included as a sub-project
 add_library(Catch2::Catch2 ALIAS Catch2)
 
+if (NOT NOT_SUBPROJECT)
+    add_library(Catch2Main src/catch_with_main.cpp)
+
+    target_link_libraries(Catch2Main Catch2)
+
+    add_library(Catch2::Main ALIAS Catch2Main)
+endif ()
+
 # Only perform the installation steps when Catch is not being used as
 # a subproject via `add_subdirectory`, or the destinations will break,
 # see https://github.com/catchorg/Catch2/issues/1373
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,11 +104,11 @@ endif()
 add_library(Catch2::Catch2 ALIAS Catch2)
 
 if (NOT NOT_SUBPROJECT)
-    add_library(Catch2Main src/catch_with_main.cpp)
+    add_library(Catch2WithMain src/catch_with_main.cpp)
 
-    target_link_libraries(Catch2Main Catch2)
+    target_link_libraries(Catch2WithMain PUBLIC Catch2)
 
-    add_library(Catch2::Main ALIAS Catch2Main)
+    add_library(Catch2::Catch2WithMain ALIAS Catch2WithMain)
 endif ()
 
 # Only perform the installation steps when Catch is not being used as
--- a/docs/slow-compiles.md
+++ b/docs/slow-compiles.md
@@ -64,6 +64,18 @@ tests-factorial.cpp:11: failed: Factorial(0) == 1 for: 0 == 1
 Failed 1 test case, failed 1 assertion.
 ```
 
+## Using the static library Catch2Main
+
+Catch also provides a static library that implements the runner. When
+Catch2 is included with `add_subdirectory` using CMake, it is possible to simply
+link against this library which also provides the include path for the header.
+
+```cmake
+add_executable(tests-factorial tests-factorial.cpp)
+
+target_link_libraries(tests-factorial Catch2::Main)
+```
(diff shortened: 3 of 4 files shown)


horenmar:
Okay, so this is a long standing thing with v2, and I am not sure if I want to merge it.

The reason why static library for v2 is problematic is that in v2, built-in reporters rely on global constructors to register themselves. Relying on global constructors in static libraries has historically been problematic, as the linker is free to, and often does, discard them. See e.g. #421 for example on how this affects `TEST_CASE` macros in static library. This means that I historically didn't want to provide static library target for Catch2, because it would be likely to generate various bugs and problems, that I would then have to deal with in our issue tracker.

However, when I tested the change locally, reporters registered themselves properly. This has led me to do some investigating, and the situation with global constructors in static libraries is actually quite complex, or maybe quite simple :smiley:.

As far as I can tell, global constructors are brought in as long as any _other_ symbol from a TU is brought in. Because the static library is compiled from the single-header implementation TU, rather than the individual pre-stitch TUs, relying on autoregistering reporters should be safe enough.

blackliner:
Sounds awesome! And honestly that's what the world out there does usually, compiling one main and linking it hundred of times. So they are already in that "gray" area, if it is any.

horenmar:
The rules for `.o` files are different: all global constructors in them are always included. Otherwise `TEST_CASE` macro wouldn't work.


Basically linking is terrible. 😃

blackliner:
So, anything left todo for you to consider a merge?

BUILD.bazel
```diff
@@ -3,8 +3,15 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
...
-  visibility = ["//visibility:public"],
-  includes = ["single_include/"],
+    name = "catch2",
+    hdrs = ["single_include/catch2/catch.hpp"],
+    includes = ["single_include/"],
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "catch2_main",
```
horenmar: Is the expectation here that user links against both `catch2` and `catch2_main`, or will linking against `catch2_main` also bring in `includes` from `catch2`, since it is a dependency?

If the second, I would prefer if the name was `catch2_with_main` (I assume bazel uses snake_case), the way v3 does.

CMakeLists.txt
```diff
@@ -103,6 +103,14 @@ endif()
 # provide a namespaced alias for clients to 'link' against if catch is included as a sub-project
 add_library(Catch2::Catch2 ALIAS Catch2)
 
+if (NOT NOT_SUBPROJECT)
+    add_library(Catch2Main src/catch_with_main.cpp)
+
+    target_link_libraries(Catch2Main Catch2)
```
horenmar: This should probably be `PUBLIC` to get dependency propagation.

CMakeLists.txt
```diff
@@ -103,6 +103,14 @@ endif()
...
+
+    add_library(Catch2::Main ALIAS Catch2Main)
+endif ()
```
horenmar, line 112: Couple of things here:

* I am unclear on why this is enabled only if Catch2 is a subproject...
* The alias needs to be `Catch2::Catch2Main`*, because the alias library exists to make usage of Catch2-as-a-subproject and Catch2-as-an-installed-library uniform. During installation, the targets are prefixed with "namespace" `Catch2::`, so the alias prepends that "namespace" to real targets.
* Again, I would prefer `Catch2WithMain` as the target name.



\* Actually, since this is available only if Catch2 is a subproject, the alias doesn't have to be there at all.

BUILD.bazel
blackliner: only catch2_main would be needed, so went with your proposal 👍

blackliner:
```suggestion
    name = "catch2_with_main",
```

CMakeLists.txt
blackliner:
```suggestion
    target_link_libraries(Catch2WithMain PUBLIC Catch2)
```

docs/slow-compiles.md
```diff
@@ -64,6 +64,28 @@ tests-factorial.cpp:11: failed: Factorial(0) == 1 for: 0 == 1
...
+### bazel
+```python
+cc_test(
+    name = "hello_world_test",
+    srcs = [
+        "test/hello_world_test.cpp",
+    ],
+    deps = [
+        "lib_hello_world",
+        "@catch2//:catch2_main",
```
blackliner:
```suggestion
        "@catch2//:catch2_with_main",
```

CMakeLists.txt
```diff
@@ -103,6 +103,14 @@ endif()
 # provide a namespaced alias for clients to 'link' against if catch is included as a sub-project
 add_library(Catch2::Catch2 ALIAS Catch2)
 
+if (NOT NOT_SUBPROJECT)
+    add_library(Catch2Main src/catch_with_main.cpp)
+
+    target_link_libraries(Catch2Main Catch2)
+
+    add_library(Catch2::Main ALIAS Catch2Main)
```
blackliner:
```suggestion
    add_library(Catch2::Catch2WithMain ALIAS Catch2WithMain)
```

CMakeLists.txt
```diff
@@ -103,6 +103,14 @@ endif()
 # provide a namespaced alias for clients to 'link' against if catch is included as a sub-project
 add_library(Catch2::Catch2 ALIAS Catch2)
 
+if (NOT NOT_SUBPROJECT)
+    add_library(Catch2Main src/catch_with_main.cpp)
```
blackliner:
```suggestion
    add_library(Catch2WithMain src/catch_with_main.cpp)
```

CMakeLists.txt
blackliner: Fixed the naming, the subproject thing was there before. No idea why or how this is required.

docs/slow-compiles.md
```diff
@@ -64,6 +64,28 @@ tests-factorial.cpp:11: failed: Factorial(0) == 1 for: 0 == 1
 Failed 1 test case, failed 1 assertion.
 ```
 
+## Using the static library Catch2Main
+Catch also provides a static library that implements the runner.
+### CMake
+```cmake
+add_executable(tests-factorial tests-factorial.cpp)
+
+target_link_libraries(tests-factorial Catch2::Main)
```
blackliner:
```suggestion
target_link_libraries(tests-factorial Catch2::Catch2WithMain)
```

CMakeLists.txt
blackliner: done

CMakeLists.txt
horenmar: The reason is to avoid intruding upon the parent project's build with our tests, and its install step... If you don't have an explicit reason to do this, I'll clean it up and merge it afterwards.

blackliner: No reason from my side 👍

horenmar:
I found out that I cannot push cleanups into your branch, so I pushed the branch locally.

What this means is that the changes are in `v2.x`, but I have to close this manually.

blackliner:
Awesome, thanks!

PR #2067, opened 2020-10-23
replacing the rest of the instances of the term "master" from the repo…

ryancdavison:
Replacing the rest of the instances of the term "master" from the repo after "master" branch was replaced with "devel"


horenmar:
The changes to codecov and conan make sense, the others do not.

PR #2071, opened 2020-10-23
Add infrastructure for embedding code samples in docs, and fix documentation

claremacrae:
## Description

This adds the scripts to use mdsnippets to embed code samples in the docs.

* tools/scripts/updateDocumentSnippets.py regenerates docs. See the comments for how to install the tool
* tools/scripts/buildAndTest.sh does a developer build and runs the tests. Much of its contents is now included in docs/contributing.md, which also lets users know about the script, so they don't need to run all the commands by hand.

Having created the above, I also fixed the build script to add the missing CMake option that is now needed.

If the PR is accepted, I'm happy to write a description in the contributing docs on how to set up mdsnippets.

### DIFF
--- a/tools/scripts/updateDocumentSnippets.py
+++ b/tools/scripts/updateDocumentSnippets.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python3
+
+from scriptCommon import catchPath
+import os
+import subprocess
+
+# ---------------------------------------------------
+# Update code examples
+# ---------------------------------------------------
+# For info on mdsnippets, see https://github.com/SimonCropp/MarkdownSnippets
+
+# install dotnet SDK from http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
+# Then install MarkdownSnippets.Tool with
+#   dotnet tool install -g MarkdownSnippets.Tool
+# To update:
+#   dotnet tool update  -g MarkdownSnippets.Tool
+# To uninstall (e.g. to downgrade to a lower version)
+# dotnet tool uninstall -g MarkdownSnippets.Tool
+
+os.chdir(catchPath)
+
+subprocess.run('dotnet tool update  -g MarkdownSnippets.Tool --version 21.2.0', shell=True, check=True)
+subprocess.run('mdsnippets', shell=True, check=True)
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ benchmark-dir
 .conan/test_package/build
 bazel-*
 build-fuzzers
+debug-build
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -56,7 +56,7 @@ By default a console width of 80 is assumed but this can be controlled by defini
     CATCH_CONFIG_NOSTDOUT
 
 To support platforms that do not provide `std::cout`, `std::cerr` and
-`std::clog`, Catch does not usem the directly, but rather calls
+`std::clog`, Catch does not use them directly, but rather calls
 `Catch::cout`, `Catch::cerr` and `Catch::clog`. You can replace their
 implementation by defining `CATCH_CONFIG_NOSTDOUT` and implementing
 them yourself, their signatures are:
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -85,7 +85,7 @@ and run all tests in the `Debug` compilation.
 ./tools/scripts/generateAmalgamatedFiles.py
 
 # 2. Configure the full test build
-cmake -Bdebug-build -H. -DCMAKE_BUILD_TYPE=Debug -DCATCH_BUILD_EXAMPLES=ON -DCATCH_BUILD_EXTRA_TESTS=ON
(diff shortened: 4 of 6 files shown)


tools/scripts/updateDocumentSnippets.py
```diff
@@ -0,0 +1,22 @@
...
+
+# install dotnet SDK from http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
+# Then install MarkdownSnippets.Tool with
+#   dotnet tool install -g MarkdownSnippets.Tool
+# To update:
+#   dotnet tool update  -g MarkdownSnippets.Tool
+# To uninstall (e.g. to downgrade to a lower version)
+# dotnet tool uninstall -g MarkdownSnippets.Tool
+
+assert os.path.exists('mdsnippets.json'), 'This script must be run from the root of Catch2 source tree'
```
horenmar: One of the script files already has a way to find out the base path to the repository -- I think this script should use it to ensure its working dir is at the repo's root.

claremacrae: Good point - done.

claremacrae:
I've switched the snippet naming convention to kebab-case now.

tools/scripts/updateDocumentSnippets.py
```diff
@@ -0,0 +1,23 @@
...
+# Then install MarkdownSnippets.Tool with
+#   dotnet tool install -g MarkdownSnippets.Tool
+# To update:
+#   dotnet tool update  -g MarkdownSnippets.Tool
+# To uninstall (e.g. to downgrade to a lower version)
+# dotnet tool uninstall -g MarkdownSnippets.Tool
+
+os.chdir(catchPath)
+
+subprocess.run('dotnet tool update  -g MarkdownSnippets.Tool --version 21.2.0', shell=True, check=True)
```
horenmar, line 22: This will fail if the system already has a newer version of the tool.

claremacrae: True. The reason for specifying a specific version is that there is a tendency for the formatting of output to change between versions... Which can lead to difficult merges, and random changes in doc files being mixed in with code changes...

Do you know of another way to install, that would solve that problem?

If not, I'll experiment with non-global installations...

horenmar:
I looked through the docs and there doesn't seem to be an option to say "look for snippets here, and for docs here". This is kinda annoying, as the scan could likely be much faster.

Otherwise lgtm.

tools/scripts/updateDocumentSnippets.py
horenmar: Not sure what is the best solution, but the global install makes this situation worse. If someone already has the tool for different project, then their version must be the same or it will fail for them.

claremacrae:
> I looked through the docs and there doesn't seem to be an option to say "look for snippets here, and for docs here". This is kinda annoying, as the scan could likely be much faster.
> 
> Otherwise lgtm.

The main thing for speed is to remember to exclude any generated directories from the search, by editing `mdsnippets.json`

How long does it take for you? It's under 2 seconds for me, which I'm comfortable with, despite being quite impatient!

tools/scripts/updateDocumentSnippets.py
claremacrae: > Not sure what is the best solution, but the global install makes this situation worse. If someone already has the tool for different project, then their version must be the same or it will fail for them.

Off the top of my head, there are at least two options - both of which need experimentation and/or more discussion:

- Enable the CI to update docs, so that users do not have to do so
- Experiment with non-global install (which I have just tried, and looks promising)

Rather than rushing to change the code now, would you consider accepting the PR for now, as a step on the way... It is at least possible to now auto-generate docs, and for anyone who has a newer install of mdsnippets, they always have the option of not updating the docs for now?

horenmar:
> How long does it take for you? It's under 2 seconds for me, which I'm comfortable with, despite being quite impatient!

5-ish seconds to run the Python script.

I have _a lot_ of build dirs. I recently changed machines, and I am already up to 4 build dirs for MSVC (and 7 build dirs for Clang+GCC, but luckily those live on a completely separate filesystem).

claremacrae:
> > How long does it take for you? It's under 2 seconds for me, which I'm comfortable with, despite being quite impatient!
> 
> 5-ish seconds to run the Python script.
> 
> I have _a lot_ of build dirs. I recently changed machines, and I am already up to 4 build dirs for MSVC (and 7 build dirs for Clang+GCC, but luckily those live on a completely separate filesystem).

1. What happens if you comment out this line in the script? (For me, most of the time is spent on the updating - but then all my builds are in directories with names starting `cmake-build-` so are excluded)

`subprocess.run('dotnet tool update  -g MarkdownSnippets.Tool --version 21.2.0', shell=True, check=True)`

I could always make it so that it checks the version number of the tool, and doesn't update if the version matches?

2. Do any in-source-tree build spaces not match the cmake-build pattern already excluded? What if you add them to the `Exclude` list in `mdsnippets.json`?

PR #2072, opened 2020-10-25
Bump OpenCppCoverage to 0.9.9.0

hanneskaeufler:


horenmar:
That's an... interesting failure. I'll have to investigate it properly later, but I guess that AppVeyor updated its images recently.

horenmar:
Found the issue, I'll have to fix up the history later.

PR #2074, opened 2020-10-28
Fix sign conversion error in template product test cases

dkavolis:
## Description

Any `TEMPLATE_PRODUCT_TEST_CASE_SIG` tests compiled with GCC `-Werror=sign-conversion` yield compilation errors. Only template product test cases are affected due to `int` and `unsigned long` arithmetics.


horenmar:
The changes only trigger more warnings.

PR #2078, opened 2020-10-29
Silence -Wuseless-cast warning with GCC10

Morwenn:
I open this pull request as a way to open the discussion with a very simple fix. I couldn't find any issue to link it to.

## Description

With GCC 10, the `static_cast<bool>` triggers the -Wuseless-cast warning. This commit changes the cast into `static_cast<const bool&>`: it achieves the same thing but doesn't trigger the warning thanks to the "gratuitous" type conversion to `const bool&`. As per references rules, `const bool&` should bind to anything, be it `const` or not, an rvalue or an lvalue, so I doubt that this change is breaking anything.


horenmar:
Ok, I am gonna try this.

PR #2086, opened 2020-11-01
Enable IPO for CMake versions that support it

AMS21:
## Description

I've began porting a project of mine to Catch3 where I have IPO/LTO enabled and since Catch is now build as a static library I've noticed the following warning coming up:
```
CMake Warning (dev) at external/Catch2/src/CMakeLists.txt:230 (add_library):
Policy CMP0069 is not set: INTERPROCEDURAL_OPTIMIZATION is enforced when
enabled.  Run "cmake --help-policy CMP0069" for policy details.  Use the
cmake_policy command to set the policy and suppress this warning.

INTERPROCEDURAL_OPTIMIZATION property will be ignored for target 'Catch2'.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
This patch should fix that warning and allow Catch to be build with link time optimizations with non Intel compilers like Clang or GCC, when enabled.
Note that this patch does not enable link time optimizations or directly modifies compile/link flags in any way. I've you want to build with these flags you still need to set 'CMAKE_INTERPROCEDURAL_OPTIMIZATION' for example or enable them per target.

[CMP0069 Refernce](https://cmake.org/cmake/help/git-stage/policy/CMP0069.html)
## GitHub Issues

N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,11 @@ if (CMAKE_BINARY_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
     message(FATAL_ERROR "Building in-source is not supported! Create a build dir and remove ${CMAKE_SOURCE_DIR}/CMakeCache.txt")
 endif()
 
+if(CMAKE_VERSION VERSION_GREATER 3.8)
+  # Enable IPO for CMake versions that support it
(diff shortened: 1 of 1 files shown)


horenmar:
Okay, makes sense.

PR #2091, opened 2020-11-02
include_guards obey pattern

innerout:
## Description

This PR introduces a new job in the CI and checks if all include_guards obey the simple pattern: `${UPPER_CASE_FILENAME}_INCLUDED`.


innerout:
For testing purposes, I will modify two header files. In the first file, I will add whitespaces to emulate a valid header file that should pass the checkguard test. In the second file, I will change the include guard to an invalid form to emulate a fail test.  After verifying that checkguard works as expected I will revert the changes in the two files.

PR #2098, opened 2020-11-06
Adds duplication check for source files in CI

### LINKED ISSUE #2076: Our CI should check that there are no duplicate filenames
If we have duplicated filenames, our include guard pattern (see #2075) could cause breakages.

innerout:
## Description

This PR checks for duplicate files in `src/catch2` to avoid include guard conflicts.
It adds a python script in the CI that recursively checks that all files are unique.  
## GitHub Issues
Closes #2076

### DIFF
--- a/.github/workflows/validate-header-guards.yml
+++ b/.github/workflows/validate-header-guards.yml
@@ -5,16 +5,13 @@ name: Check header guards
 on: [push, pull_request]
 
 jobs:
-  # Set the job key. The key is displayed as the job name
-  # when a job name is not provided
-  checkguard:
-    # Name the Job
-    name: Check include guard naming convention
+  build:
     # Set the type of machine to run on
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@v2
+
+      - name: Checkout source code
+        uses: actions/checkout@v2
 
       - name: Setup Dependencies
         uses: actions/setup-python@v2
@@ -23,7 +20,8 @@ jobs:
       - name: Install checkguard
         run: pip install guardonce
 
-      - name: Run checkguard
+        # Check include guard naming convention
+      - name: checkguard
         run: |
           wrong_files=$(checkguard -r src/catch2/ -p "name | append _INCLUDED | upper")
           if [[ $wrong_files ]]; then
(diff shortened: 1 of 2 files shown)


check_duplicates.py
```diff
@@ -0,0 +1,14 @@
...
+
+import os
+import sys
+
+files_map = {}
+
+for root, dir, files in os.walk("src/catch2"):
+    for file in files:
+        if file not in files_map:
+            files_map[file] = 1
```
horenmar: Why not use a `set` and not have to make up a dummy key?

Alternatively, keep around the full path to the file, so the error can also report paths to the file... (But using plain set is simpler).

.github/workflows/check-duplicates.yml
```diff
@@ -0,0 +1,26 @@
+name: Check duplicates
+
+# Run this workflow every time a new commit pushed to your repository
+
+on: [push, pull_request]
+
+jobs:
+  # Set the job key. The key is displayed as the job name
+  # when a job name is not provided
+  checkguard:
```
horenmar: This should have a different key (`checknames`?), and be part of `validate-header-guards.yml`. Both of the jobs are part of validating the header guards, so let's keep them together. (It also removes some shared boilerplate.)

check_duplicates.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
```
horenmar: This file should be in `tools/scripts`, and probably have a more obvious filename (`checkDuplicateFilenames`?).

horenmar (changes requested):

check_duplicates.py
innerout: I did not know python had sets thanks for telling me.

.github/workflows/check-duplicates.yml
innerout: Refactored `validate-header-guards.yml` to avoid duplicated code.

horenmar (approved):

horenmar:
Looks good. Can you rebase + squash the changes?

innerout:
Yes, I can rebase. Should I squash all the commits into the first?

horenmar:
Yes

PR #2099, opened 2020-11-07
Adds license check in CI

### LINKED ISSUE #2077: Our CI should check that all library files include the proper licence header
As part of the cleanups in `devel` branch, I standardized the licence headers in all of the library (as opposed to test) files.

The new header is, as contributing.md says,
```cpp
    
    //              Copyright Catch2 Authors
    // Distributed under the Boost Software License, Version 1.0.
    //   (See accompanying file LICENSE_1_0.txt or copy at
    //        https://www.boost.org/LICENSE_1_0.txt)

```

We should check that all files use this licence header as part of automatic per-commit CI.

innerout:
## Description
Adds checkLicense.py that runs on a per-commit basis in CI checking the license for all source files.  

## GitHub Issues
Closes #2077

### DIFF
--- a/.github/workflows/validate-header-guards.yml
+++ b/.github/workflows/validate-header-guards.yml
@@ -34,3 +34,6 @@ jobs:
       - name: checknames
         run: |
           python tools/scripts/checkDuplicateFilenames.py
+      - name: checklicense
+        run: |
+          python tools/scripts/checkLicense.py
--- a/tools/scripts/checkLicense.py
+++ b/tools/scripts/checkLicense.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python3
+import os
+import sys
+
+
+def get_license():
+    with open("src/catch2/catch_all.hpp", "r") as f:
+        license = f.readlines()[0:7]
+
+    return license
+
+
+def check_license(license):
+    failed = 0
+    base_dir = "src/catch2/"
+
+    # The _ represents the list of directories in base_dir
+    for root, _, files in os.walk(base_dir):
+        for file in files:
+            with open(root + "/" + file, "r") as f:
+                file_license = f.readlines()[0:7]
+
+            if file_license != license:
+                print("File %s does not have license" % file)
+                failed = 1
+
+    return failed
+
+
+license = get_license()
+status = check_license(license)
+sys.exit(status)


tools/scripts/checkLicense.py
```diff
@@ -0,0 +1,49 @@
...
+
+def extract_license(split_license):
+    assert split_license != None
+
+    license = ""
+
+    for line in split_license:
+        license += line
+
+    return license
```
horenmar: You can simplify this into single line, but at that point you might just get rid of the function altogether.

```suggestion
    return ''.join(split_license)
```

tools/scripts/checkLicense.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+import os
+import sys
+
+
+def get_license():
+    f = open("src/catch2/catch_all.hpp", "r")
+    license = "".join(f.readlines()[0:7])
+    f.close()
```
horenmar: You should use context managers for handling closing files automatically.

tools/scripts/checkLicense.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def check_license(license):
+    failed = 0
+    base_dir = "src/catch2/"
+
+    # The _ represents the list of directories in base_dir
+    for root, _, files in os.walk(base_dir):
+        for file in files:
+            f = open(root + "/" + file, "r")
```
horenmar: Same as above.

tools/scripts/checkLicense.py
innerout: Fixed, thanks again for teaching me something new!

tools/scripts/checkLicense.py
```diff
@@ -0,0 +1,34 @@
...
+            file_license = "".join(f.readlines()[0:7])
```
horenmar: Thinking about this some more:
Why turn this into a string? Python lists are comparable, so you can just compare the slices directly.

innerout: Well managing one big string is easier than managing many small ones for me. But you are right the code is probably better without the `join`.

.github/workflows/validate-header-guards.yml
```diff
@@ -34,3 +34,6 @@ jobs:
       - name: checknames
         run: |
           python tools/scripts/checkDuplicateFilenames.py
+      - name: checklicense
```
horenmar, line 37: As a side note, this doesn't have to be so simple. The value for `name` key is a string, so you can write more descriptive name, e.g. "Check that files use proper license header", like so

```yaml
name: Check that files use proper license header
```

PR #2112, opened 2020-11-30
Update cmake-integration.md

Emilia-Jarosz:
## Description
The "contrib" folder was renamed to "extras", but the name wasn't updated in the documentation.


horenmar:
I would be highly surprised if that was the only thing out of the date there, but thanks 😃

Issue #2109, opened 2020-11-25
Add AllMatch, AnyMatch, NoneMatch generic range matchers


horenmar:
They should take single matcher during construction, and their `match` method should take a generic range, using the new generic matcher support.

Expected usage:
```cpp
REQUIRE_THAT(get_keys(), AllMatch(KeyPattern({1, 2, 3})));
REQUIRE_THAT(get_numbers(), NoneMatch(IsOdd{}));
REQUIRE_THAT(get_strings(), AnyMatch(Contains("webscale") && !Contains("MongoDB")));
```

the matching semantics are hopefully obvious from the names 😃

horenmar:
There might also be interest in simple `All`, `Any`, `None` (or maybe `AllTrue`, etc.) range matchers that are explicitly specialized for ranges of `bool`s, or types convertible to `bool`. While they can be simply emulated by composing `AllMatch` with `IsTrue` generic matcher, having them out of box would likely be convenient for bunch of users.

urielgarciarivas:
@horenmar Hello! I would love to contribute here. Where should I implement those methods? And could you tell me more about the new generic matcher support?

horenmar:
The [new matchers docs are here](https://github.com/catchorg/Catch2/blob/devel/docs/matchers.md#writing-custom-matchers-new-style). There is a very similar functionality in `ContainsElementMatcher`, in `src/catch2/matchers/catch_matchers_contains.hpp`, which takes a Matcher at construction, a generic range* at match time, and returns true if at least one element in the range matches.

This is essentially the `AnyMatch` implementation done already :smiley: The other two should be analoguous.


* Looking at it, I will need to check whether it accepts C++20 ranges, which can have different type returned from `begin` and from `end`. I think it currently doesn't.

urielgarciarivas:
Thank you! I will be working on this. :sunflower:

PR #2114, opened 2020-12-07
Added AllMatch, AnyMatch, NoneMatch generic range matchers

urielgarciarivas:
## Description
Added AnyMatch, AllMatch and NoneMatch generic range matchers which take single matcher during construction, and their `match` method take a generic range, using the new generic matcher support.

## GitHub Issues
Addressing issue #2109 .


src/catch2/matchers/catch_matchers_all.hpp
```diff
@@ -25,6 +25,7 @@
 #include <catch2/matchers/catch_matchers_contains.hpp>
 #include <catch2/matchers/catch_matchers_exception.hpp>
 #include <catch2/matchers/catch_matchers_floating.hpp>
+#include <catch2/matchers/catch_matchers_generic.hpp>
```
horenmar: This is not a good filename given the contents. "Logical" would be better, but it still doesn't feel ideal.

src/catch2/matchers/catch_matchers_generic.hpp
```diff
@@ -0,0 +1,117 @@
...
+        // Creates a matcher that checks whether no element in a range matches a matcher.
+        template <typename Matcher>
+        std::enable_if_t<Detail::is_matcher<Matcher>::value,
+        NoneMatchMatcher<Matcher>> NoneMatch(Matcher&& matcher) {
+            return { std::forward<Matcher>(matcher) };
+        }
+
+        // Creates a matcher that checks whether any element in a range is equal to a specific element.
+        template <typename Matcher>
+        std::enable_if_t<Detail::is_matcher<Matcher>::value,
```
horenmar: There is no overload set, so this does not need to be conditionally enabled.

I don't think there is even a need to static assert on this -- if it quacks like a matcher, there is no reason not to use it.

horenmar: (This obviously goes for the other functions as well)

src/catch2/matchers/catch_matchers_generic.hpp
```diff
@@ -0,0 +1,117 @@
...
+//        https://www.boost.org/LICENSE_1_0.txt)
+
+// SPDX-License-Identifier: BSL-1.0
+#ifndef CATCH_MATCHERS_GENERIC_HPP_INCLUDED
+#define CATCH_MATCHERS_GENERIC_HPP_INCLUDED
+
+#include <catch2/matchers/catch_matchers_templated.hpp>
+
+#include <algorithm>
+#include <functional>
```
horenmar: I don't think this header uses either of these two.

src/catch2/matchers/catch_matchers_generic.hpp
```diff
@@ -0,0 +1,117 @@
...
+                m_matcher(std::move(matcher))
+            {}
+
+            std::string describe() const override {
+                return "all match " + m_matcher.describe();
+            }
+
+            template <typename RangeLike>
+            bool match(RangeLike&& rng) const {
+                using std::begin; using std::endl;
```
horenmar: You shouldn't need these, range-for already does the requisite ADL lookup

horenmar (changes requested):
Thanks.

The PR is off to a good start, but some changes are required and there need to be tests.

src/catch2/matchers/catch_matchers_all.hpp
urielgarciarivas: Do you mean I need to create a file for each matcher? e. g. `catch_matchers_addall.hpp`

src/catch2/matchers/catch_matchers_generic.hpp
urielgarciarivas: You are right!! Haha thanks

src/catch2/matchers/catch_matchers_all.hpp
horenmar: Not necessarily. But I don't think people would expect to find `AllMatch`, `AnyMatch`, `NoneMatch` in file named "matchers generic", when the documentation already talks about "Generic Matchers" as a way of defining new matchers. Ideally there would be a good name for them as a group, but I can't think of any.

Placing them each into a separate file works too I guess, e.g. `catch_matchers_allmatch.hpp`

urielgarciarivas: What about quantifiers? `catch_matchers_quantifiers.hpp` which includes `AllMatch`, `AnyMatch` and `NoneMatch`.

horenmar: I won't say that I think it is great, but since I've already gotten "quantifiers" as a suggestion from someone else, it is apparently not as unintuitive as I though.

So, sure, go for it. 👍

urielgarciarivas: Great, I'll update it!

urielgarciarivas:
I need to create tests in `test/SelfTest/UsageTests/MatchersRanges.tests.cpp` or should I create them somewhere else?

horenmar (approved):

horenmar:
Put them into `MatchersRanges.tests.cpp` . I intend to refactor (mostly split them apart) the matchers tests, but there is no reason for you to worry about that.

horenmar:
You need to update [Approval tests](https://github.com/catchorg/Catch2/blob/devel/docs/contributing.md#testing-your-changes)


Also you should test that the matchers support ranges that require ADL, and ranges that use different type for `begin` and `end`.

horenmar:
Okay, so what's missing now:
* Document the changes
* Tests
  1) Can the matcher handle a type that requires its specific ADL begin found?
  2) Can the matcher handle a type that returns different types for `begin` and `end`?
  3) Does the matcher shortcircuit?

I can take it from here if you want.

urielgarciarivas:
What would happen to my PR? Would you merge this PR or should this PR wait until you add what's missing?

horenmar:
I would add the tests on top of your branch and then merge it when it is done.

urielgarciarivas:
Please do! Thank you very much!

horenmar:
Okay, tests should be there.

I'll wait for the CI to finish, and then rebase + cleanup the MR once again before merging.

urielgarciarivas:
Closes #2109

PR #2123, opened 2020-12-18
Add the HIP CPU Runtime as a Catch2 user

AlexVlx:
## Description

The HIP CPU Runtime uses Catch2 for its unit tests and micro-benchmarks, and it only feels proper to admit to it. Thank you for an excellent tool gents.


horenmar:
Thanks.

PR #2141, opened 2020-12-30
Make the build reproducible

maaaat:
Make the build reproducible by removing references to the source
directory from the Catch2WithMain static library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,11 @@ add_library(Catch2WithMain ${CMAKE_CURRENT_LIST_DIR}/src/catch_with_main.cpp)
 target_link_libraries(Catch2WithMain PUBLIC Catch2)
 add_library(Catch2::Catch2WithMain ALIAS Catch2WithMain)
 
+# Make the build reproducible on versions of g++ and clang that supports -ffile-prefix-map
+if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND NOT ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 8) OR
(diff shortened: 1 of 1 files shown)


horenmar:
So I looked at https://reproducible-builds.org/docs/build-path/ and supposedly Clang 10+ also supports this. I will merge the PR if you add Clang as well.

CMakeLists.txt
```diff
@@ -108,6 +108,11 @@ add_library(Catch2WithMain ${CMAKE_CURRENT_LIST_DIR}/src/catch_with_main.cpp)
 target_link_libraries(Catch2WithMain PUBLIC Catch2)
 add_library(Catch2::Catch2WithMain ALIAS Catch2WithMain)
 
+# Make the build reproducible on versions of g++ and clang that supports -ffile-prefix-map
+if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND NOT ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 8) OR
+   ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND NOT ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 10))
+  target_compile_options(Catch2WithMain PRIVATE "-ffile-prefix-map=${CMAKE_SOURCE_DIR}=.")
```
horenmar, line 114: Shouldn't this also be added to `Catch2` target? This way only the small static library that defines main is affected, instead of the whole library.

horenmar: Nevermind, for some reason I thought you were targeting the `devel` branch.

maaaat: As far as I could check this is sufficient for the catch2 package to pass the reproducible build test in Debian. I didn't have the time to look much further.

PR #2143, opened 2020-12-31
Do not match exact amount of spaces for errno macro expansion in approvalTests.py

Johnnynator:
## Description
The amount of spaces between the function name and the () is irrelevant, and breaks test on musl libc,
since it expands to `__errno_location()` without the space in between.


horenmar:
Sure. I intend to remove those tests from approvals completely, but there is no reason not to fix it in the meantime.

PR #2152, opened 2021-01-13
Fix ULP matcher

yyny:
In [the ULP matcher implemented](https://github.com/catchorg/Catch2/blob/a091853f4ae62b428983c63224fec9f1af377295/src/catch2/matchers/catch_matchers_floating_point.cpp#L52-L55), there is a check `(lc < 0) != (rc < 0)` seemingly there to allow `+0` and `-0` to compare equal even if `maxUlpDiff` is zero. However, this check passes for _any_ two numbers that do not have the same sign. This is incorrect, for example `almostEqualUlps(-EPSILON, EPSILON, INT64_MAX)` returns false, even though it is obviously true (`2` ULPS < `INT_MAX` ULPS). I've reimplemented the matcher, using `boost::math` as a reference.


horenmar:
Thanks for bringing it to my attention, I redid the fix to simplify the implementation and get rid of superfluous changes in the PR.

PR #2158, opened 2021-01-21
Wrap all std::min and std::max calls in parentheses

ScottHutchinson:
Fixes x86 MSVC v19.28 compiler error C2589 on Windows.

Many of the std::min and std::max calls were already wrapped in parentheses.

Some of the errors were for benchmark builds only.

The template says "do not submit pull requests changing the `version.hpp`
or the single-include `catch.hpp` file", but it's not clear where else to make these changes, so I'm submitting this PR as a WIP until I figure out where to make the changes.


ScottHutchinson:
Replaced with PR #2159

PR #2159, opened 2021-01-23
Wrap all std::min and std::max calls in parentheses

ScottHutchinson:
Fixes x86 MSVC v19.28 compiler error C2589 on Windows.

Many of the std::min and std::max calls were already wrapped in parentheses.

Some of the errors were for benchmark builds only.

### DIFF
--- a/include/internal/benchmark/detail/catch_estimate_clock.hpp
+++ b/include/internal/benchmark/detail/catch_estimate_clock.hpp
@@ -68,7 +68,9 @@ namespace Catch {
             }
             template <typename Clock>
             EnvironmentEstimate<FloatDuration<Clock>> estimate_clock_cost(FloatDuration<Clock> resolution) {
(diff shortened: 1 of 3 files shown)


horenmar:
Ok, thanks.

ferdnyc:
@horenmar @ScottHutchinson 
Just FYI, a possibly-more-robust solution to this — so that everyone doesn't have to live in continual fear that someone will add an unparenthetized `std::min()` or `std::max()`call to the code at some point in the future and re-break things — is to [`#define NOMINMAX`](https://en.sfml-dev.org/forums/index.php?topic=26401.0) wherever this issue comes up. (Or just _everywhere_, when building on Windows / with MSVC.)

That will deactivate the `min()` and `max()` preprocessor macros stupidly defined in `WinDefs.h` (which is presumably being included directly or indirectly); it seems those bad macros are the root cause of these issues. ([More on the issue](http://www.suodenjoki.dk/us/archive/2010/min-max.htm).)

PR #2166, opened 2021-01-31
Add surrogate TUs for header files

### LINKED ISSUE #2106: Add "surrogate TUs" for all library headers
Surrogate TUs are TUs that only include one single header, and do nothing else. These TU must still include: if they do not, then a header doesn't stand on its own, and there is likely some include or declaration missing from it.

There are three parts to this
- [x] Add surrogate TUs for all current headers
- [x] Ensure that they are built during testing (probably under "Extra Tests" CMake toggle, to save compilation times for common builds)
- [x] Add CI script that checks whether all headers have their surrogate TU

innerout:
## Description

This PR adds surrogate TUs to test that each header file in the project can be included without depending on other files.

In order for the PR to be complete I have to clean the commit history.
Also 5 files are commented in the CMakeLists.txt of SurrogateCpps since they need to be fixed.

## GitHub Issues

Closes #2106

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -263,4 +263,6 @@ endif()
 list(APPEND CATCH_WARNING_TARGETS SelfTest)
 set(CATCH_WARNING_TARGETS ${CATCH_WARNING_TARGETS} PARENT_SCOPE)
 
-add_subdirectory(SurrogateCpps)
+if(CATCH_BUILD_SURROGATES)
+    add_subdirectory(SurrogateCpps)
+endif()
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -262,3 +262,5 @@ endif()
 
 list(APPEND CATCH_WARNING_TARGETS SelfTest)
 set(CATCH_WARNING_TARGETS ${CATCH_WARNING_TARGETS} PARENT_SCOPE)
+
+add_subdirectory(SurrogateCpps)
--- a/tests/ExtraTests/CMakeLists.txt
+++ b/tests/ExtraTests/CMakeLists.txt
@@ -224,3 +224,14 @@ set_tests_properties(
   PROPERTIES
     PASS_REGULAR_EXPRESSION "All tests passed \\(14 assertions in 3 test cases\\)"
 )
+
+add_executable(SurrogateCatchBenchmark
+  ${TESTS_DIR}/surr_catch_benchmark.cpp
+)
+
+target_include_directories(SurrogateCatchBenchmark PUBLIC ${CATCH_DIR}/src/)
+set_property( TARGET SurrogateCatchBenchmark PROPERTY CXX_STANDARD 14 )
+set_property( TARGET SurrogateCatchBenchmark PROPERTY CXX_STANDARD_REQUIRED ON )
+set_property( TARGET SurrogateCatchBenchmark PROPERTY CXX_EXTENSIONS OFF )
+
+add_test(NAME SurrogateCatchBenchmarkTest COMMAND SurrogateCatchBenchmark)
--- a/tools/scripts/checkSurrogateExists.py
+++ b/tools/scripts/checkSurrogateExists.py
@@ -1,0 +1,36 @@
+#!/bin/python3
+import os
+import sys
+
+srcdir = "../../src/catch2"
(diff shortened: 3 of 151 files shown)


innerout:
Cleared commit history.

tests/ExtraTests/CMakeLists.txt
```diff
@@ -224,3 +224,14 @@ set_tests_properties(
...
+add_executable(SurrogateCatchBenchmark
+  ${TESTS_DIR}/surr_catch_benchmark.cpp
+)
+
+target_include_directories(SurrogateCatchBenchmark PUBLIC ${CATCH_DIR}/src/)
+set_property( TARGET SurrogateCatchBenchmark PROPERTY CXX_STANDARD 14 )
+set_property( TARGET SurrogateCatchBenchmark PROPERTY CXX_STANDARD_REQUIRED ON )
+set_property( TARGET SurrogateCatchBenchmark PROPERTY CXX_EXTENSIONS OFF )
+
+add_test(NAME SurrogateCatchBenchmarkTest COMMAND SurrogateCatchBenchmark)
```
horenmar, line 237: Why are these separate?

tests/CMakeLists.txt
```diff
@@ -262,3 +262,7 @@ endif()
 
 list(APPEND CATCH_WARNING_TARGETS SelfTest)
 set(CATCH_WARNING_TARGETS ${CATCH_WARNING_TARGETS} PARENT_SCOPE)
+
+if(CATCH_BUILD_SURROGATES)
+    add_subdirectory(SurrogateCpps)
+endif()
```
horenmar, line 268: The option should also have a top level definition, so that various CMake GUIs pick it up better. (I'll take care of this)

tools/scripts/checkSurrogateExists.py
```diff
@@ -0,0 +1,36 @@
+#!/bin/python3
```
horenmar, line 1:
```suggestion
#!/bin/env python3
```

tools/scripts/checkSurrogateExists.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+import sys
+
+srcdir = "../../src/catch2"
```
horenmar, line 5: There is already a centralized way to get the path to the top level folder in, IIRC, `releaseCommon` or `scriptCommon.py`

tools/scripts/checkSurrogateExists.py
```diff
@@ -0,0 +1,36 @@
...
+surrogates_folder = "../../tests/SurrogateCpps"
```
horenmar, line 6: This should be built based on `srcdir`.

tools/scripts/checkSurrogateExists.py
```diff
@@ -0,0 +1,36 @@
...
+
+def get_files(filetype,dir):
+    filenames = []
+
+    for _, _, files in os.walk(dir):
+        for file in files:
+            if filetype in file:
+                filenames.append(file)
+
+    return filenames
```
horenmar, line 19: Is there an advantage to this approach compared to using [`glob`](https://docs.python.org/3/library/glob.html)?

tests/ExtraTests/CMakeLists.txt
innerout: I think I copy-pasted from an existing CMakeLists; there was no specific reason.

tools/scripts/checkSurrogateExists.py
innerout: I will refactor to use one of the two mentioned scripts.

tools/scripts/checkSurrogateExists.py
innerout: I don't know how glob works, but I believe there will be no difference for the number of files in Catch2's case.

horenmar:
Actually don't do a thing now, this might be an entirely wrong approach 😃 

I realized that this approach where the surrogate cpp files actually exist on disk add a lot of mechanical overhead to creating new headers, and that we could replace it with generating the surrogate cpp files on-demand. This would move the mechanical work from the developer to a machine, which is much better.

innerout:
At some point, I got bored copy-pasting headers, and I wrote a script to autogenerate the surrogate files. Should I make it generic to autogenerate them based on what exists in the src directory of catch2?

PR #2172, opened 2021-02-06
use current version of mdsnippets

SimonCropp:


horenmar:
Is the output of mdsnippets stable between versions? Because otherwise I would prefer a pinned version to floating one.

SimonCropp:
very rarely. but yes it can

PR #2177, opened 2021-02-16
Add surrogate TUs option to CMake

### LINKED ISSUE #2106: Add "surrogate TUs" for all library headers

horenmar:
## Description
This PR adds surrogate TUs to Catch2, by having them autogenerated by CMake during build. It also fixes some headers that weren't self-sufficient, and makes surrogate TU builds part of the CI.

To build is dependent on the `CATCH_BUILD_SURROGATES` CMake option (which is in turn dependent on `CATCH_DEVELOPMENT_BUILD`).

## GitHub Issues
#2106

### DIFF
--- a/.github/workflows/linux-builds.yml
+++ b/.github/workflows/linux-builds.yml
@@ -20,6 +20,7 @@ jobs:
           - clang++-10
         build_type: [Debug, Release]
         std: [14]
+        surrogates: [OFF]
         include:
         # cannot be installed on ubuntu-20.04 be default?
 #          - cxx: g++-6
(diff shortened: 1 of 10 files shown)


horenmar:
@innerout I think this covers the surrogates better -- I just need to figure out the right combination of escaping to make the redirect-to-file work on Windows 😃

innerout:
I have been thinking about CMake that must track the autogenerated files. Do you have anything on your mind?

horenmar:
That already happens. The trick is that when the user adds new header, it causes reconfiguration. During reconfiguration, a target is created for each header in the main library... and since there is a different target that depends on their outputs, they will be run and thus generate the surrogate headers.

PR #2179, opened 2021-02-17
Expand documentation for BDD test macros

matthew-limbinar:
## Description
Added documentation and an example on how to use the BDD macros. Perhaps the BDD example is too long to be embedded? I'll let my reviewer(s) decide.

### DIFF
--- a/docs/test-cases-and-sections.md
+++ b/docs/test-cases-and-sections.md
@@ -108,8 +108,32 @@ executed and validated -- there are additional macros starting with `AND_`:
 
 These are used to chain ```GIVEN```s, ```WHEN```s and ```THEN```s together. The `AND_*` clause is placed 
 _inside_ the clause on which it depends. There can be multiple _independent_ clauses that are all _dependent_ 
-on a single outer clause. See [this working example](https://godbolt.org/z/cMjsx4) (with an intentional 
-failure to demonstrate the reporting output).
+on a single outer clause. 
+```cpp
+SCENARIO( "vector can be sized and resized" ) {
+    GIVEN( "An empty vector" ) {
+        auto v = std::vector<std::string>{};
+
+        // Validate assumption of the GIVEN clause
+        THEN( "The size and capacity start at 0" ) {
+            REQUIRE( v.size() == 0 );
+            REQUIRE( v.capacity() == 0 );
+        }
+
+        // Validate one use case for the GIVEN object
+        WHEN( "push_back() is called" ) {
+            v.push_back("hullo");
+
+            THEN( "The size changes" ) {
+                REQUIRE( v.size() == 1 );
+                REQUIRE( v.capacity() >= 1 );
+            }
+        }
+    }
+}
+```
+See [this working example](https://godbolt.org/z/cMjsx4), with an intentional failure to demonstrate the reporting 
+output.
 
 > `AND_GIVEN` was [introduced](https://github.com/catchorg/Catch2/issues/1360) in Catch2 2.4.0.
 
--- a/docs/test-cases-and-sections.md
+++ b/docs/test-cases-and-sections.md
@@ -132,7 +132,7 @@ SCENARIO( "vector can be sized and resized" ) {
     }
 }
 ```
-See [this working example](https://godbolt.org/z/cMjsx4), with an intentional failure to demonstrate the reporting 
+See [this working example](https://godbolt.org/z/e5vPPM), with an intentional failure to demonstrate the reporting 
 output.
 
 > `AND_GIVEN` was [introduced](https://github.com/catchorg/Catch2/issues/1360) in Catch2 2.4.0.
--- a/docs/test-cases-and-sections.md
+++ b/docs/test-cases-and-sections.md
@@ -90,13 +90,77 @@ This macro maps onto ```TEST_CASE``` and works in the same way, except that the 
 * **WHEN(** _something_ **)**
 * **THEN(** _something_ **)**
 
-These macros map onto ```SECTION```s except that the section names are the _something_s prefixed by "given: ", "when: " or "then: " respectively.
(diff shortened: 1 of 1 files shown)


docs/test-cases-and-sections.md
```diff
@@ -90,15 +90,79 @@ This macro maps onto ```TEST_CASE``` and works in the same way, except that the
 * **WHEN(** _something_ **)**
 * **THEN(** _something_ **)**
 
-These macros map onto ```SECTION```s except that the section names are the _something_s prefixed by "given: ", "when: " or "then: " respectively.
+These macros map onto ```SECTION```s except that the section names are the _something_s prefixed by "given: ", "when: " or "then: " respectively. These macros also map onto the AAA or A<sup>3</sup> test pattern (standing either for [Assemble-Activate-Assert](http://wiki.c2.com/?AssembleActivateAssert) or [Arrange-Act-Assert](http://wiki.c2.com/?ArrangeActAssert)), and in this context, the macros provide both code documentation and reporting of these parts of a test case without the need for extra comments or code to do so.
```
horenmar: It has been that in the original already, but the `_something_s` won't render properly.

Also new documentation is written with explicit linebreaks, to avoid situation like this where a single line of diff represents many lines of actual text, which in turn makes reviewing/diffing hard. 😃

horenmar:
The example is indeed too long. It would be better off significantly shorter, but also showing the expected output of running such scenario.

You could actually even make that compilable via e.g. compiler explorer -- the v3 versions are compilable and runnable there.

docs/test-cases-and-sections.md
matthew-limbinar: @horenmar I have attempted to implement your review comments. Let me know what you think.

matthew-limbinar:
BTW, I don't believe the Mac build failure has anything to do with my doc changes.

horenmar:
> BTW, I don't believe the Mac build failure has anything to do with my doc changes.

And even without looking I can say that there is 99.9% chance of you being right 😃 

The Mac builders are overloaded and there is a brittle test there that I have to fix at some point.

matthew-limbinar:
👍 Do I need to do anything further to get this PR merged, @horenmar?

docs/test-cases-and-sections.md
```diff
@@ -90,15 +90,28 @@ This macro maps onto ```TEST_CASE``` and works in the same way, except that the
...
+executed and validated -- there are additional macros starting with `AND_`:
 
 * **AND_GIVEN(** _something_ **)**
 * **AND_WHEN(** _something_ **)**
 * **AND_THEN(** _something_ **)**
 
-Similar to ```GIVEN```, ```WHEN``` and ```THEN``` except that the prefixes start with "and ". These are used to chain ```GIVEN```s, ```WHEN```s and ```THEN```s together.
+These are used to chain ```GIVEN```s, ```WHEN```s and ```THEN```s together. The `AND_*` clause is placed 
+_inside_ the clause on which it depends. There can be multiple _independent_ clauses that are all _dependent_ 
+on a single outer clause. See [this working example](https://godbolt.org/z/cMjsx4) (with an intentional
```
horenmar: The CE code should link against `trunk` version of Catch2, rather than preview 2. I got the bug that made `trunk` not work fixed last week.

docs/test-cases-and-sections.md
```diff
@@ -90,15 +90,28 @@ This macro maps onto ```TEST_CASE``` and works in the same way, except that the
...
 
 * **AND_GIVEN(** _something_ **)**
 * **AND_WHEN(** _something_ **)**
 * **AND_THEN(** _something_ **)**
 
-Similar to ```GIVEN```, ```WHEN``` and ```THEN``` except that the prefixes start with "and ". These are used to chain ```GIVEN```s, ```WHEN```s and ```THEN```s together.
+These are used to chain ```GIVEN```s, ```WHEN```s and ```THEN```s together. The `AND_*` clause is placed 
+_inside_ the clause on which it depends. There can be multiple _independent_ clauses that are all _dependent_ 
+on a single outer clause. See [this working example](https://godbolt.org/z/cMjsx4) (with an intentional 
+failure to demonstrate the reporting output).
```
horenmar: A short example should still be inline with the documentation, with the link only providing a more belabored example.

Given the linked code, I would use something like

```cpp
SCENARIO( "vector can be sized and resized" ) {
    GIVEN( "An empty vector" ) {
        auto v = std::vector<std::string>{};

        // Validate assumption of the GIVEN clause
        THEN( "The size and capacity are 0" ) {
            REQUIRE( v.size() == 0 );
            REQUIRE( v.capacity() == 0 );
        }

        // Validate one use case for the GIVEN objects
        WHEN( "push_back() is called" ) {
            v.push_back("hullo");

            THEN( "The size changes" ) {
                REQUIRE( v.size() == 1 );
                REQUIRE( v.capacity() >= 1 );
            }
        }
    }
}
```

for the inline example, together with an example of how the BDD macros are then reported.

docs/test-cases-and-sections.md
```diff
@@ -90,15 +90,28 @@ This macro maps onto ```TEST_CASE``` and works in the same way, except that the
...
 
-> `AND_GIVEN` was [introduced](https://github.com/catchorg/Catch2/issues/1360) in Catch 2.4.0.
+> `AND_GIVEN` was [introduced](https://github.com/catchorg/Catch2/issues/1360) in Catch 2.4.0. (The other BDD macros have been around since 1.x.)
```
horenmar:
```suggestion
> `AND_GIVEN` was [introduced](https://github.com/catchorg/Catch2/issues/1360) in Catch 2.4.0. (The other BDD macros were introduced in 1.x.)
```

or just leave it out, the cut out is either 2.4.0 or 2.3.0, and what isn't explicitly documented has been added before that. (The "leave older things out" approach is in use in documentation in other places.)

horenmar:
Otherwise it looks fine.

docs/test-cases-and-sections.md
matthew-limbinar: Done.

matthew-limbinar:
@horenmar, I have applied your suggestions. Let me know what you think!

matthew-limbinar:
@horenmar, just checking in. Anything I can do to get this merged?

horenmar:
No, sorry. I wanted to make some further changes after merging this, but didn't have time to do them, so I put off merging the changes 😃 

I fixed the merge conflict and I am going to merge this now.

PR #2182, opened 2021-03-03
MatcherMethod: Add virtual destructor and default constructors

johnelse:
The lack of virtual destructor raises warnings with Visual Studio 2019
16.7 and later. Declaring this then requires the constructor, copy
constructor and move constructor to be provided.

I notice `-Wnon-virtual-dtor` is explicitly disabled for clang just
above the declaration of this struct, but it's not clear from this
file's history why that was done rather than adding the destructor.


johnelse:
I'm happy to also remove the `clang`-specific warning tweaking if you think it can go as a result of this.

Issue #1712, opened 2019-08-02
Allow multiple reporters with multiple output files


horenmar:
**Description**

There have been requests for specifying multiple reporters to be used for the same test run. This was implemented for a while, but it ran into trouble caused by the fact that there was no way to give them different output files/streams for them to write to. If we reintroduce this possibility, we also need to decide on a syntax that allows different reporters to write to different files/streams.

**Additional context**
Possible syntaxes:
* `-r` takes a list, `-o` takes a list, the size of list in `-r` must be at most 1 longer than the one in `-o`. Reporters are then given the targets from `-o` in the same order they are specified.
Examples: `-r junit xml -o junit.out` -> junit reporter writes to the file `junit.out`, xml reporter writes to stdout.
* `-r` takes a list of reporters, where the name can be followed by a colon and the name of the output file, e.g. `-r junit:junit.out` would say that the junit reporter should write its output to `junit.out`. Stdout would then be specified as either `-`, or without any followup after the name.

We also need to decide what to do if multiple reporters would be writing to stdout -- ignore it? Warn the user? Error out?

JoeyGrajciar:
Personally i would go with second option.
It would be less prone to errors in defining of the reporters order.
Stdout could be specified by leaving out the optional `:<filename>` part.

I would go with error for mutliple reporters writing to stdout.

macetw:
This is a basic fundamental need. CI frameworks (like Jenkins) need this information. Developers also need this information while running on their console, and they're not interested in digging in to XML files fishing out this information. It's also important that the same test logic runs on development workstations as what runs on CI frameworks. Please prioritize this request.

WimLeflere:
@macetw with the current reporters and output file options you can already have different output for CI and developers.
On CI you can for example use the xml reporter and write it to a file.
On the developers machine you can use the console reporter and write to the terminal.
With the same test code/executable.

PR #2183, opened 2021-03-08
Add support for multiple reporters

mjerabek:
## Description

This MR adds support for multiple reporters. CLI syntax: `-r xml:output.xml -r console`. The `-o` option is still available and determines the default output file, which is used by listeners and reporters without their own output file specified. Only one reporter may use the default output.

## TODO

* [x] Deal with different stdout capturing strategies of the reporters.
* [ ] Allow specifying `--use-colour` per reporter.
    * This can wait for its own MR, but the CLI syntax should be decided now. Using `:` as option delimiter (`-r console:out.txt:colour=1`) is problematic because of Windows absolute paths (pathologic example: `-r console:c:\colour=1`).
* [x] Extend Approval tests to test multiple reporters.
* [ ] Improve documentation.
* [ ] Format the patches using `clang-format`.

## GitHub Issues

Addresses #1712


mjerabek:
Approval tests added, but ATM failing because of the unhandled mixed stdout capturing strategies.

src/catch2/catch_config.hpp
```diff
@@ -132,7 +133,7 @@ namespace Catch {
         TestSpec m_testSpec;
         bool m_hasTestFilters = false;
     };
-
+    std::ostream& operator<<(std::ostream &os, Catch::ConfigData::ReporterAndFile const& reporter);
```
horenmar: Should be a hidden friend. Probably doesn't need `Catch::` namespace in front of `ConfigData`.

src/catch2/catch_config.cpp
```diff
@@ -30,6 +31,14 @@ namespace Catch {
         } // unnamed namespace
     } // namespace Detail
 
+    std::ostream& operator<<(std::ostream &os, ConfigData::ReporterAndFile const& reporter) {
```
horenmar: I am surprised clang-format doesn't linebreak this.

tests/SelfTest/IntrospectiveTests/String.tests.cpp
```diff
@@ -35,7 +35,17 @@ TEST_CASE( "StringRef", "[Strings][StringRef]" ) {
         REQUIRE_THROWS(original.c_str());
         REQUIRE_NOTHROW(original.data());
     }
-
+    SECTION( "Copy construction" ) {
```
horenmar: I would prefer "Copy construction is shallow" or "Copy constructor is shallow"

mjerabek:
Ad stdout capturing:

Suppose we have this code:
```c++
TEST_CASE("foo") {
    CHECK(1 == 1);
    std::cout << "Hello\n";
    CHECK(2 == 2);
    std::cout << "World\n";
}
```
and run it as `./test -s` and `./test -s -r console -r xml:out.xml`. In case A (no capturing), the messages and reports from the reporter will be interleaved. In case B, we need to capture the output for the XML reporter, but we also need it to be printed to the real stdout.

Which brings me to another thing - if the non-capturing reporter output is directed to a file (`./test -s -r console:out.txt`), the prints still go to stdout (and not the reporter stream). In other words, non-capturing reporters explicitly *don't care* about the stdout. In this spirit, the interleaving of stdout/reporter messages to console (or the lack of thereof) should not matter and just printing the captured output after the testcase ends should be enough.

Although it likely *will* be confusing to users. OTOH having prints intermixed with assertions is bad form anyway.

**Proposal**: If both capturing and non-capturing reporters are used at the same time, the output is captured and re-printed to stdout/stderr *at the end of the testcase*. More generally, let us define that the relative ordering of print messages w.r.t. the reporter output is *unspecified*. Single-reporter cases are not affected, so this doesn't break compatibility.

src/catch2/catch_config.hpp
mjerabek: I wanted to avoid including `<ostream>`, but it's probably not that important in this case. Reimplemented as a hidden friend. The `Catch::` is gone.

tests/SelfTest/IntrospectiveTests/String.tests.cpp
mjerabek: My primary concern was if it by chance doesn't perform conversion to (a temporary) `std::string` and then construct a new `StringRef` from that (which it didn't, because the conversion operator is `explicit`). So maybe "Copy construction doesn't construct a temporary std::string"?

src/catch2/catch_config.hpp
horenmar: You can have hidden friend declaration separate from definition, as long as you don't mind that for the TU with the definition, the function is not treated as hidden friend.

horenmar:
There is actually multiple things that will need to be handled for multiple reporters in a smarter manner, e.g. what if one of the reporters wants to be informed of all assertion successes, while others do not.

src/catch2/catch_config.cpp
mjerabek: Right. It likely would, I just never ran it. Took this as an opportunity to install the sublime plugin. I will reformat the patches when this is finished.

mjerabek:
> There is actually multiple things that will need to be handled for multiple reporters in a smarter manner, e.g. what if one of the reporters wants to be informed of all assertion successes, while others do not.

That one was easy (or I got it wrong). But there's a catch: AFAICS, the fact if `Listener`s got all assertions or just the failing ones was determined by the used reporter. Now it's based solely on the Listener's preferences. I consider this a fix, but it *might* break existing listeners that do not set the preference explicitly (good thing this goes to a major release).

src/catch2/catch_config.cpp
```diff
@@ -13,12 +13,27 @@
...
+#include <ostream>
 
 namespace Catch {
+    namespace Detail {
+        namespace {
+            class RDBufStream : public IStream {
+                mutable std::ostream m_os;
+            public:
+                // Note that the streambuf must live at least as long as this object.
+                RDBufStream(std::streambuf *sb) : m_os( sb ) {}
```
horenmar: Catch2 now uses doxygen style comments for these things.

src/catch2/catch_config.hpp
```diff
@@ -20,6 +22,25 @@ namespace Catch {
...
+            std::string reporterName;
+
+            // If none, the output goes to the default output.
+            Option<std::string> outputFileName;
+
+            friend bool operator==(ReporterAndFile const& a, ReporterAndFile const& b) {
+                return a.reporterName == b.reporterName && a.outputFileName == b.outputFileName;
+            }
+
+            friend std::ostream& operator<<(std::ostream &os, ReporterAndFile const& reporter) {
```
horenmar: I would rather demote the definition (and the related include) into the .cpp file.

src/catch2/catch_config.hpp
```diff
@@ -20,6 +22,25 @@ namespace Catch {
     struct IStream;
 
     struct ConfigData {
+        struct ReporterAndFile {
+            std::string reporterName;
+
+            // If none, the output goes to the default output.
+            Option<std::string> outputFileName;
+
+            friend bool operator==(ReporterAndFile const& a, ReporterAndFile const& b) {
```
horenmar: Nit: I strongly prefer `lhs` and `rhs` 😃

src/catch2/internal/catch_clara.hpp
```diff
@@ -448,16 +469,21 @@ namespace Catch {
                     m_ref( ref ) {}
 
             public:
-                template <typename T>
+                template <typename T, typename = typename std::enable_if<!Detail::is_unary_function<T>::value>::type>
```
horenmar:
```suggestion
                template <typename T, typename = typename std::enable_if_t<!Detail::is_unary_function<T>::value>>
```

v3 targets C++14 as baseline, so traits can be used with the `_t` suffix. (Sadly, no `_v` for values)

src/catch2/internal/catch_clara.hpp
```diff
@@ -448,16 +469,21 @@ namespace Catch {
...
                     m_hint( hint ) {}
 
-                template <typename LambdaT>
+                template <typename LambdaT, typename = typename std::enable_if<Detail::is_unary_function<LambdaT>::value>::type>
                 ParserRefImpl( LambdaT const& ref, std::string const& hint ):
                     m_ref( std::make_shared<BoundLambda<LambdaT>>( ref ) ),
                     m_hint( hint ) {}
 
+                template <typename LambdaT>
+                ParserRefImpl( accept_many_t, LambdaT const& ref, std::string const& hint ):
```
horenmar: If the "accept many" tag is placed as the last argument, couldn't we do without the sfinae?

src/catch2/internal/catch_commandline.cpp
```diff
@@ -115,16 +116,54 @@ namespace Catch {
                 return ParserResult::runtimeError( "Unrecognised verbosity, '" + verbosity + "'" );
             return ParserResult::ok( ParseResultType::Matched );
         };
-        auto const setReporter = [&]( std::string const& reporter ) {
+        auto const setReporter = [&]( std::string const& reporterSpec ) {
             IReporterRegistry::FactoryMap const& factories = getRegistryHub().getReporterRegistry().getFactories();
 
-            auto lcReporter = toLower( reporter );
+            // clear the default reporter
```
horenmar, line 144: This function feels complex enough that it should be split away and tested separately.

src/catch2/internal/catch_stringref.hpp
```diff
@@ -31,6 +31,8 @@ namespace Catch {
 
     public: // construction
         constexpr StringRef() noexcept = default;
+        constexpr StringRef(StringRef const&) noexcept = default;
+        constexpr StringRef& operator=(StringRef const&) noexcept = default;
```
horenmar: Why?

(Also, this disable moves. While moves aren't optimized for StringRef, I think this could be surprising in the future when StringRef reports to be nonmoveable under some type trait).

src/catch2/reporters/catch_reporter_listening.cpp
```diff
@@ -15,151 +17,157 @@ namespace Catch {
...
+            reporter->getPreferences().shouldReportAllAssertions;
+
+        m_listeners.push_back( std::move( reporter ) );
     }
 
     void ListeningReporter::noMatchingTestCases( std::string const& spec ) {
         for ( auto const& listener : m_listeners ) {
             listener->noMatchingTestCases( spec );
         }
-        m_reporter->noMatchingTestCases( spec );
```
horenmar: I see the idea under unifying listeners and reporters, but we promise that listeners are always called before reporters. We cannot promise this unless we require that all calls to `addListener` are done before `addReporter` calls.

src/catch2/reporters/catch_reporter_listening.cpp
```diff
@@ -15,151 +17,157 @@ namespace Catch {
...
-    void ListeningReporter::addReporter(IStreamingReporterPtr&& reporter) {
-        assert(!m_reporter && "Listening reporter can wrap only 1 real reporter");
-        m_reporter = std::move( reporter );
-        m_preferences.shouldRedirectStdOut = m_reporter->getPreferences().shouldRedirectStdOut;
+    void ListeningReporter::addReporter( IStreamingReporterPtr&& reporter ) {
+        m_preferences.shouldRedirectStdOut |= reporter->getPreferences().shouldRedirectStdOut;
+
+        // We will need to output the captured stdout if there are reporters
+        // that do not want it captured.
+        m_haveNoncapturingListeners |= !reporter->getPreferences().shouldRedirectStdOut;
```
horenmar: I am actually not sure these are the correct semantics...

The basic idea behind capturing is that for reporters with structured output, we cannot let tests just print to stdout and break the format... However, do we care about this if the structured output is written elsewhere? Do we care about also printing the output if there is a reporter with non structured output? What if that reporter is redirected? 😕  I honestly don't know.

src/catch2/catch_config.hpp
horenmar: Also, is this used anywhere that is not the self tests? If not, I would prefer a `StringMaker` specialization there.

src/catch2/reporters/catch_reporter_listening.cpp
```diff
@@ -15,151 +17,157 @@ namespace Catch {
...
         }
-        m_reporter->sectionEnded( sectionStats );
     }
 
     void ListeningReporter::testCaseEnded( TestCaseStats const& testCaseStats ) {
+        if ( m_preferences.shouldRedirectStdOut && m_haveNoncapturingListeners ) {
+            if ( !testCaseStats.stdOut.empty() ) {
+                Catch::cout() << testCaseStats.stdOut << std::flush;
+            }
+            if ( !testCaseStats.stdOut.empty() ) {
```
horenmar:
```suggestion
            if ( !testCaseStats.stdErr.empty() ) {
```

I didn't check the actual name, but...

horenmar:
One more generic concern: How long are the approval tests now? For me, they are already the longest individual CTest test. This adds a lot more of them.

horenmar:
Idea for temporarily workaround for the console colour issue:

Colour support is already kinda broken and only works by accident (and sometimes stops working). There is no sane way to support multiple reporters with Windows colours. Thus, let's disable all colours if there are multiple reporters specified, at least temporarily. Later on the colours can be properly reworked.

src/catch2/catch_config.hpp
mjerabek: Done.

tests/SelfTest/IntrospectiveTests/String.tests.cpp
mjerabek: Changed to "Copy construction is shallow".

src/catch2/catch_config.hpp
mjerabek: Demoted. I'll check where it's needed.

src/catch2/catch_config.hpp
mjerabek: Done.

src/catch2/internal/catch_stringref.hpp
mjerabek: Sigh. True. I just wanted to make it explicit that the class has copy operations, so that the `StringRef( std::string const& stdString )` overload is not selected for copy (which I am frankly not sure would be impossible if the `operator std::string` weren't `explicit`).

I'll remove it.

src/catch2/reporters/catch_reporter_listening.cpp
mjerabek: Can the registered listeners/reporters change mid-operation? I.e., can we get away with just "sorting" the list at some point? I'd rather avoid the code duplication if possible, unless of course it remains the easiest way.

src/catch2/reporters/catch_reporter_listening.cpp
mjerabek: > The basic idea behind capturing is that for reporters with structured output, we cannot let tests just print to stdout and break the format

I don't entirely agree with that. I think we want to capture the output so that we can include it in the structured output in some well-defined manner. I don't see any reason to support `./prints-to-stdout-test -r xml >out.xml`. There *is* a supported way to do this (`./prints-to-stdout-test -r xml:out.xml`, or even `-r xml -o out.xml`).

On the other hand, something like `./prints-to-stdout-test -r xml | process-the-xml` is not _that_ far-fetched anymore. In that case we really don't want the output to be reprinted to stdout.

However, it may be plausible to run `./test -r console -r xml:out.xml`, where I think I would want to see the output. OTOH in that case I can very well specify that explicitly as an opt-in.

Another conundrum is `./test -r console:out.txt -r xml:out.xml`. The stdout output on stdout will do us no good, unless of course it is unrelated to the tests at all and we want it separately. OTOH if we want it to be included in the txt report, it would mean we want the console reporter to *capture* the output and print it in a well-defined place. We could get this with a capturing console reporter, which would be trivial to implement.

**Summary**: Capturing reporters may want to include the captured output in some structured way. Non-capturing reporters do not care about stdio at all. Unless they are a capturing reporter in disguise and the structured way to display the output is to print it at a given place (at the end of testcase or at the place where it would be if the outputs were interleaved). But it never worked that way (`./test -r console -o out.txt`) and if we want that, we should get a capturing console reporter.

And that can be the answer to the "reprinting in report". Just create a `capturing-console` reporter, which will print the captured output at the end of a test case.

In case the stdout/stderr contains an independent diagnostics from the tested codebase, one may want to preserve the stream exactly as-is. We can easily achieve that by adding a capturing reporter that just prints the captured stdout/stderr. So we would have `./test -r xml:out.xml -r 'stdout:&1' -r 'stderr:&2'`, where `&1` and `&2` would be a special path syntax to specify stdout/stderr.

---

**TL;DR**: If at least one capturing reporter is specified, the stdio is captured. If one wishes to include it in a report, they should use an appropriate capturing reporter for that use case. If one wishes to capture the stdout/stderr separately or reprint it to stdout/stderr, they should use special reporters `stdout`/`stderr` which just reprint the captured output to the specified stream.

Tasks:
* [ ] do not reprint anything
* [ ] create a `capturing-console` reporter which will print the captured stdio at the end of each test case
  * to demonstrate the possibility; when per-reporter options are implemented, we
    can make it an option for the console reporter (but that's for later)
* [ ] create `stdout`, `stderr` reporters, which will just print the captured stdout/stderr at the end of each test case
* [ ] add support for `&1`, `&2` output path specifications

src/catch2/internal/catch_clara.hpp
mjerabek: I don't think so. The reason for the sfinae is not the added `accept_many_t` feature, but the ability to pass named lambdas, i.e.
```c++
auto foo = [](std::string opt) { ... };

clara::Opt(foo, "foo")["--foo"]
```

Before, the passed lambda had to be "not convertible to non-const ref".

src/catch2/reporters/catch_reporter_listening.cpp
horenmar: Adding more listeners/reporters after startup is not supported, once we add all registered listeners and selected reporters, that's the final set.

docs/command-line.md
```diff
@@ -131,6 +131,8 @@ The bundled reporters are:
 
 The JUnit reporter is an xml format that follows the structure of the JUnit XML Report ANT task, as consumed by a number of third-party tools, including Continuous Integration servers such as Jenkins. If not otherwise needed, the standard XML reporter is preferred as this is a streaming reporter, whereas the Junit reporter needs to hold all its results until the end so it can write the overall results into attributes of the root node.
 
+This option may be passed multiple times to use multiple (different) reporters  at the same time. See [Reporters](reporters.md#multiple-reporters) for details.
```
horenmar, line 141: Newly written docs should use linebreaks for <80 cols lines.

Also, maybe have the link text be "Multiple Reporters" or "documentation on using multiple reporters"?

docs/reporters.md
```diff
@@ -35,6 +35,17 @@ You see what reporters are available from the command line by running with `--li
 
 By default all these reports are written to stdout, but can be redirected to a file with [`-o` or `--out`](command-line.md#sending-output-to-a-file)
 
+<a id="multiple-reporters"></a>
```
horenmar, line 38: We usually use the heading ids for linking, but it might be a good idea to be inconsistent but have a stable link for cross-documentation linking.

src/catch2/internal/catch_clara.hpp
```diff
@@ -53,7 +54,27 @@ namespace Catch {
...
+        struct accept_many_t {};
+        constexpr accept_many_t accept_many {};
+
         namespace Detail {
+            struct fake_arg {
+                template <typename T>
+                operator T();
+            };
+
+            template<typename... Ts> struct make_void { using type = void; };
```
horenmar: I am 99% sure that we already have this somewhere else, including a `_t` suffixed type alias. Since Clara is no longer a separate project, it should reuse the one that Catch2 already has.

src/catch2/internal/catch_clara.hpp
horenmar: Ok, makes sense. About as much as C++ tmp can make sense :smiley:

src/catch2/reporters/catch_reporter_listening.cpp
horenmar: However I would like to avoid having some sort of `Multireporter::finalize` member function, so maybe it should internally keep track of "last {reporter/listener} insert index" and just insert new reporters/listeners to the right place.

There shouldn't ever be enough reporters/listeners that the potential n^2 performance comes at a meaningful runtime cost.

src/catch2/reporters/catch_reporter_listening.cpp
```diff
@@ -15,151 +17,157 @@ namespace Catch {
...
-        m_preferences.shouldRedirectStdOut = m_reporter->getPreferences().shouldRedirectStdOut;
+    void ListeningReporter::addReporter( IStreamingReporterPtr&& reporter ) {
+        m_preferences.shouldRedirectStdOut |= reporter->getPreferences().shouldRedirectStdOut;
+
+        // We will need to output the captured stdout if there are reporters
+        // that do not want it captured.
+        m_haveNoncapturingListeners |= !reporter->getPreferences().shouldRedirectStdOut;
+
+        m_preferences.shouldReportAllAssertions |=
+            reporter->getPreferences().shouldReportAllAssertions;
```
horenmar: Listeners should probably participate in this too.

Or maybe they should even force this to be on? Given their even based design it seems weird that you can have a listener only receive some assertions... But obviously listeners might also care about completely different events, and then it is pointless to pass them assertions.

I think this might end up being quite surprising for the users, and we might need to consider having listeners default to `shouldReportAllAssertions == true`, but allow them to opt out...

src/catch2/reporters/catch_reporter_listening.cpp
```diff
@@ -15,151 +17,157 @@ namespace Catch {
...
             listener->assertionStarting( assertionInfo );
         }
-        m_reporter->assertionStarting( assertionInfo );
     }
 
     // The return value indicates if the messages buffer should be cleared:
     bool ListeningReporter::assertionEnded( AssertionStats const& assertionStats ) {
-        for( auto& listener : m_listeners ) {
-            static_cast<void>( listener->assertionEnded( assertionStats ) );
+        const bool alwaysReport =
```
horenmar: Can we have different name for this? I remember this name being in more places but I think it sends the wrong message.

src/catch2/reporters/catch_reporter_listening.cpp
horenmar: Ok, so I've thought about this a bit, and I agree in regards to correctness.

I had an idea about having the capture mechanism be transparent in cases where the uncapturing reporter is the one writing to stdout, which could be done (with some mild annoyance) with the capture mechanisms, but then I decided that it seems like too much work.

The reason was that by the time I am down to printf debugging, I am usually kinda desperate and want the output to happen exactly when written and as written, without further surprises. However, then I realized that I don't need to have multiple reporters in these cases 🤷

docs/command-line.md
horenmar: I am gonna do a larger doc rewrite later.

PR #2186, opened 2021-03-12
Fix introduced in version placeholders

### LINKED ISSUE #2082: Fix introduced-in version placeholder to use proper project name
Currently the "introduced in version xyz" documentation placeholder uses format "... in Catch X.Y.Z". It should use "Catch2" instead, because that is the name of the project, and we should keep our documentation consistent.

### Changes needed;
* [x] Update `tools/scripts/releaseCommon.py` to look for, and write out, new placeholder
* [x] Update `docs/contributing.md` to mention the new placeholder
* [x] Update docs to use the new format for consistency

JoeyGrajciar:
Resolves #2082

### DIFF
--- a/tools/scripts/releaseCommon.py
+++ b/tools/scripts/releaseCommon.py
@@ -127,8 +127,8 @@ def updateVersionDefine(version):
 def updateVersionPlaceholder(filename, version):
     with open(filename, 'rb') as file:
         lines = file.readlines()
-    placeholderRegex = re.compile(b' in Catch X.Y.Z')
-    replacement = ' in Catch {}.{}.{}'.format(version.majorVersion, version.minorVersion, version.patchNumber).encode('ascii')
+    placeholderRegex = re.compile(b' in Catch[0-9]? X.Y.Z')
+    replacement = ' in Catch2 {}.{}.{}'.format(version.majorVersion, version.minorVersion, version.patchNumber).encode('ascii')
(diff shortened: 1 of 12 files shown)


horenmar (approved):
lgtm

One thing I am considering is supporting both `in Catch2 X.Y.Z` and `in Catch X.Y.Z` as the placeholder, as it would prevent some potential doc bugs, while having a reasonably small implementation cost.

JoeyGrajciar:
Will add both then, but not sure what bugs you mean.

horenmar:
If a contributor used the old form out of habit, and it got past review, it wouldn't be updated upon release.

tools/scripts/releaseCommon.py
```diff
@@ -127,8 +127,8 @@ def updateVersionDefine(version):
 def updateVersionPlaceholder(filename, version):
     with open(filename, 'rb') as file:
         lines = file.readlines()
-    placeholderRegex = re.compile(b' in Catch X.Y.Z')
-    replacement = ' in Catch {}.{}.{}'.format(version.majorVersion, version.minorVersion, version.patchNumber).encode('ascii')
+    placeholderRegex = re.compile(b' in Catch[0-9]? X.Y.Z')
```
horenmar, line 130: That works 😃

horenmar (approved):

PR #2190, opened 2021-03-17
When used as subproject, treat Catch2 cmake target include paths as SYSTEM

ashaduri:
… to avoid compiler warnings in Catch2 headers.

## Description
When embedding Catch2 into a cmake project using `add_subdirectory()`, `Catch2` target include paths are simply PUBLIC.
If a user project has many compiler warnings enabled, they may get triggered simply by including Catch2 headers.

This PR marks the Catch2 include paths as SYSTEM when it's used as a subproject, so that these include paths are treated as system header locations. This makes the compilers avoid user-enabled warnings from Catch2 headers.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -271,11 +271,20 @@ target_compile_features(Catch2
     cxx_variadic_macros
 )
 
-target_include_directories(Catch2
-  PUBLIC
-    $<BUILD_INTERFACE:${SOURCES_DIR}/..>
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
https://github.com/ChrisThrasher/argon/blob/master/tests/FindCatch2.cmake#L10

This can be fixed from your project as well if you'd like. Take the `Catch2` target's interface include directories and re-apply them with `SYSTEM`.

PR #2195, opened 2021-03-26
Fix compiling v2.x with C++17 + Clang 5 + libstdc++ v5

dimztimz:
This basically tests the combination where the compiler supports most of C++17 but the library does not.


horenmar:
Thanks, this looks good.

PR #2197, opened 2021-03-27
Fix BUILD.bazel in 2.x when used as a submodule

JennToo:
## Description

When Catch2 is imported as a submodule, using an absolute target path is
problematic because it can't be known in advance where the submodule
will be placed.

The `devel` branch already seems to already have a relative rule path,
so a change is not needed for that branch.


JennToo:
On second thought this actually doesn't seem like a good idea. It's better to just use this library through a `WORKSPACE` rule instead.

PR #2199, opened 2021-03-30
Lcc should be ignored

Mardalemer:
## Description

E2K platform has custom GCC-like compiler LCC that hasn't got all GCC's capabilities and current way for warning suppression in macros fails building with LCC.

The motivation is to build catch2 on e2k platform with lcc compiler.


horenmar:
Thanks, good luck.

PR #2200, opened 2021-04-02
Deleted

unsafe-andrew-old:
Replaced a part of if construction with switch/case. It increased lines number but made the code more readable, as for me


dionisioedu (approved):

horenmar:
Between having both `return` and `break` per case, and relying on handling most of the inputs in the implicit fallthrough, I really don't find your version more readable.

unsafe-andrew-old:
Okay

Issue #2202, opened 2021-04-05
Catch2's tests should never be run, if it is loaded as a sub-project


claremacrae:
_Please note: I will submit PRs to fix this on the v2 and v3 branches... the fix is trivial_

**Describe the bug**

Catch2's top-level `CMakeLists.txt` file has code to prevent self-tests being run if the project is being bundled e.g. via `add_subdirectory()` or `FetchContent`.

```cmake
# detect if Catch is being bundled,
# disable testsuite in that case
if(NOT DEFINED PROJECT_NAME)
    set(NOT_SUBPROJECT ON)
endif()
```

However, if `NOT_SUBPROJECT` was already set to `ON` in a parent scope, the self-tests are still run anyway (and also the 20-30 CTest targets are added)

**Expected behavior**

Turn off `NOT_SUBPROJECT` if Catch2 is running as a sub-project

**Reproduction steps**

Steps to reproduce the bug.

```bash
git clone https://github.com/claremacrae/demo-cmake-subproject-error.git

mkdir build
cd    build

cmake ..
cmake --build . --target help
```

Expected output:

```
The following are some of the valid targets for this Makefile:
... all (the default if no target is provided)
... clean
... depend
... rebuild_cache
... edit_cache
... package_source
... package
... Catch2WithMain
... tests
```

Actual output - note the include of `SelfTest`, as well as all the targets generated by including CTest:
```
The following are some of the valid targets for this Makefile:
... all (the default if no target is provided)
... clean
... depend
... install/local
... install/strip
... install
... package_source
... rebuild_cache
... edit_cache
... list_install_components
... package
... Catch2WithMain
... ContinuousCoverage
... ContinuousTest
... ContinuousBuild
... ContinuousStart
... ExperimentalBuild
... ExperimentalCoverage
... ExperimentalSubmit
... ContinuousUpdate
... ExperimentalUpdate
... ExperimentalStart
... ContinuousSubmit
... ExperimentalTest
... NightlySubmit
... NightlyMemCheck
... ExperimentalConfigure
... NightlyCoverage
... NightlyTest
... ContinuousMemCheck
... NightlyBuild
... NightlyConfigure
... NightlyUpdate
... NightlyMemoryCheck
... ContinuousConfigure
... Nightly
... NightlyStart
... ExperimentalMemCheck
... Continuous
... Experimental
... SelfTest
... tests
```

**Platform information:**

 - OS: **Windows NT**
 - Compiler+version: **GCC v2.9.5**
[...]

PR #2204, opened 2021-04-05
Prevent Catch2 v2 tests from running if loaded as a sub-project

claremacrae:
## Description

What: Prevent Catch2 v2's tests from being run, and CTest targets from being added, if Catch2 is configured as a sub-project.

Why: In projects that use Catch2 via `add_subdirectory` or `FetchContent`, this prevents the IDE's lists of targets fro being cluttered up with all CTest's targets - and it also makes CLion's "Run all CTest targets" a lot more useful, as it then only runs the user's tests, and not Catc2 ones as well.

## GitHub Issues

See #2202 - there will be a separate fix for the `devel` branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,8 @@ cmake_minimum_required(VERSION 3.5)
 # disable testsuite in that case
 if(NOT DEFINED PROJECT_NAME)
     set(NOT_SUBPROJECT ON)
+else()
+    set(NOT_SUBPROJECT OFF)
 endif()
 
 # Catch2's build breaks if done in-tree. You probably should not build


claremacrae:
The CI failures are due to GitHub Actions dropping older clang and gc versions - nothing to do with this PR:
See https://github.com/actions/virtual-environments/issues/2950

CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ cmake_minimum_required(VERSION 3.5)
 # disable testsuite in that case
 if(NOT DEFINED PROJECT_NAME)
     set(NOT_SUBPROJECT ON)
+else()
+    set(NOT_SUBPROJECT OFF)
```
horenmar, line 8: I think this would be better off as

`set(NOT_SUBPROJECT NOT DEFINED PROJECT NAME)`, and later turning `if(NOT_SUBPROJECT)` into `if(${NOT_SUBPROJECT})`.

claremacrae: Could you please say why?

I can think of a number of downsides of that:

* It means updating all the uses of the variable need to be updated, increasing the scope of the PR and the chance of error
* It requires anyone making future edits to this project to notice, if they were making edits that used this variable (e.g. if they were updating an old fork), that they would need to update a usage of a variable whose use had been changed in this PR
* The [current usage is idiomatic](https://github.com/search?q=%22NOT_SUBPROJECT%22+filename%3ACMakeLists.txt&type=Code) in lots of CMake projects - to use something different risks people making errors if ever making future edits in the project

horenmar: My thinking here is that there are two ways to go about this.

* The way that is currently in the PR. In any common programming language, it would be a terrible style, in CMake it is less so, because it doesn't really handle boolean expressions in a way one would expect.
* The suggested way, which is less commonly in use, but more closely follows standard programming style in different languages. (CMake also uses it, i.e. for `cmake_dependent_option`)

I weakly prefer the second option, but I am willing to accept either style. I also think there is actually a much more important question to ask: why is the change necessary at all? `if` should already interpret empty string as false, and dereferencing an empty variable provides an empty string. Is the problem here that the same-named variable is defined by the including project?


----------

e: okay I see, the original issue mentions that the project above us could've already set the variable.

PR #2206, opened 2021-04-06
Fix typos in the code base

jbrianceau:
## Description
Fix few typos in the code base.
Note that only documentation and comments are impacted by this change.

### DIFF
--- a/extras/catch_amalgamated.cpp
+++ b/extras/catch_amalgamated.cpp
@@ -3550,7 +3550,7 @@ namespace Catch {
             // Extracts the actual name part of an enum instance
             // In other words, it returns the Blue part of Bikeshed::Colour::Blue
             StringRef extractInstanceName(StringRef enumInstance) {
-                // Find last occurence of ":"
+                // Find last occurrence of ":"
                 size_t name_start = enumInstance.size();
                 while (name_start > 0 && enumInstance[name_start - 1] != ':') {
                     --name_start;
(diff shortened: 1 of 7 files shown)


extras/catch_amalgamated.cpp
```diff
@@ -3550,7 +3550,7 @@ namespace Catch {
             // Extracts the actual name part of an enum instance
             // In other words, it returns the Blue part of Bikeshed::Colour::Blue
             StringRef extractInstanceName(StringRef enumInstance) {
-                // Find last occurence of ":"
```
horenmar: Neither of the `extras/catch_amalgamated.{h,c}pp` files should be touched outside of creating new release.

horenmar:
Thanks, but if you want the changes to propagate to cppwinrt, you should probably run a similar PR against the `v2.x` branch.

extras/catch_amalgamated.cpp
jbrianceau: Oh I missed that, I've just updated the PR to revert the change on this file.

jbrianceau:
> Thanks, but if you want the changes to propagate to cppwinrt, you should probably run a similar PR against the `v2.x` branch.

This pull request is about quite cosmetic changes, so I don't think it's worth doing a backport of it for the `v2.x` branch.
Still, let me know if you think this backport would be nice to have, and I'll create a PR against this branch in that case.

horenmar:
I think that having a separate PR for v2 would be useful.

jbrianceau:
> I think that having a separate PR for v2 would be useful.

Sure, here it is #2208

PR #2211, opened 2021-04-12
avoid "implicit int" functions

bbolli:
The later definitions have a `void` return type, so use the same type in the "do-nothing" alternatives that are used e.g. when compiling for MinGW.


bbolli:
Duplicate of #2213.

PR #2226, opened 2021-04-29
Allow changing the format of StringMaker<float> and StringMaker<double>

pazner:
## Description

Allows the user to change the output of `StringMaker<float>` and `StringMaker<double>` from `std::fixed` to `std::scientific`.

This can be useful when comparing very small numbers, which in fixed notation requires counting zeros, but in scientific notation is clear.

The user can make this change by setting the static `format` variable, e.g. `Catch::StringMaker<float>::format = std::scientific;`.

### DIFF
--- a/docs/tostring.md
+++ b/docs/tostring.md
@@ -127,6 +127,17 @@ inside the `StringMaker` specialization, like so:
 This assertion will fail and print out the `testFloat1` and `testFloat2`
 to 15 decimal places.
 
+`StringMaker<float>` and `StringMaker<double>` default to using fixed output,
+but can be set to use scientific notation by modifying the `format` static
+variable:
+
+```cpp
+        Catch::StringMaker<float>::format = std::scientific;
+        const float testFloat1 = 1e-5f;
+        const float testFloat2 = 1e-6f;
+        REQUIRE(testFloat1 == testFloat2);
+```
+
 ---
 
 [Home](Readme.md#top)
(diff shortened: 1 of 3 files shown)


pazner:
Does this seem like a reasonable addition? Should any changes be made to make it suitable for merging?

pazner:
@horenmar, any chance you could take a quick look at this PR? Thanks!

mjacobse:
As pointed out in https://github.com/catchorg/Catch2/issues/2238#issuecomment-1261152406, the removal of trailing zeros in `fpToString` would cause trailing zeros in the exponent to be cut off. So as suggested there, this part should be bypassed if `format != std::fixed`.

Other than that I really think this pull request should be considered. Scientific output seems pretty vital to me for getting reliably meaningful test output with any non-trivial floating point data. Apart from the issue with small values as pointed out in #2238, for large numbers an excessive amount of meaningless digits will be printed for high precisions with `std::fixed`.

steppi:
Would it be possible for this to be revisited/revived? I've been working on using Catch2 to set up tests for a library of mathematical special functions [xsf](https://github.com/scipy/xsf) and am happy with how things are going, but it would be nice to be able to rely on Catch2's own logging rather than having to write things on my own. This application requires output to be printed in scientific notation with full precision in the mantissas. I'd be happy to take this on if the problem is that there was no one to work on it, since it only needs small adjustments.

steppi:
@pazner, would you like to give me permission to push to your branch here, or do you think it would be better for me to start a new PR from my own branch that includes your commits?

pazner:
@steppi, feel free to push directly to the branch in my fork. I invited you as a collaborator.

steppi:
> @steppi, feel free to push directly to the branch in my fork. I invited you as a collaborator.

Thanks @pazner!

PR #2227, opened 2021-05-06
Catch tests as dll

forry:
## Description

CMake changes only.
 * Added CATCH_EXPERIMENTAL_SHARED option for setting Catch2 to SHARED and eporting every symbol for windows.
 * ~Removed DESTINATION from install command since the defaults works better.~ Given even more destinations because why the hell not (need cmake 3.14 to get rid of it).

So I can compile the tests to separate dlls and explicitly link them to a runner application. This is especially usefull for integration/system level testing.


dimztimz:
This should be closed, it it already implemented.

forry:
Nice

PR #2231, opened 2021-05-11
include <iterator> for using std::back_inserter

vincent-hui:
`<iterator>` should be included explicitly for using std::back_inserter. My test code cannot be compiled without including `<iterator>` on my platform.

Sorry, I do not know where `#include <iterator>` should be added except in this header file. Would you mind telling me where `#include <iterator>` should be added if the single-include `catch.hpp` should not be changed by pull requests?

### DIFF
--- a/single_include/catch2/catch.hpp
+++ b/single_include/catch2/catch.hpp
@@ -5612,6 +5612,8 @@ namespace Catch {
         double clockCost;
     };
 
+#include <iterator>
+	
     template <class Duration>
     struct BenchmarkStats {
         BenchmarkInfo info;


horenmar:
Done in the proper place in the referencing commit.

vincent-hui:
Thank you very much.

When will new https://github.com/catchorg/Catch2/blob/v2.x/single_include/catch2/catch.hpp be generated?

horenmar:
I don't want to promise anything, but for now you could generate it yourself with `scripts/generateSingleHeader.py`

PR #2237, opened 2021-05-23
Don't include CTest in non-development builds.

siretty:
## Description

When Catch2 is used as a CMake subproject (via add_subdirectory) and is not built in development mode (CATCH_DEVELOPMENT_BUILD) CTest is not required for the build, as testing is off. When CTest is included it creates various targets (around 20) which are shown in IDEs such as Clion and pollute the list of configurations.

This patch conditionally includes CTest only if Catch2 is built in development mode.


horenmar:
Thanks

Issue #2103, opened 2020-11-15
FetchContent doesn't seem to work as described in the documentation with Catch2


tjwrona:
I am trying to use `FetchContent` to include Catch2 into a CMake project I am working on, but it does not seem to work. After using `FetchContent` as described in the documentation the `catch_discover_tests()` function is still not found.

Following the documentation it states that `FetchContent` should be usable for including Catch2 into a project as follows:

```cmake
Include(FetchContent)

FetchContent_Declare(
  Catch2
  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
  GIT_TAG        v2.13.1)

FetchContent_MakeAvailable(Catch2)

target_link_libraries(tests Catch2::Catch2)
```

But after following these steps, I am still unable to call `catch_discover_tests()` within my CMake files.

I even tried adding `find_package(Catch2 REQUIRED)` and `include(Catch)` but both of those commands fail as well.

**Platform information:**

 - OS: **Windows 10**
 - Compiler+version: **Visual Studio 2019 16.19.0 Preview 1.0**
 - Catch version: **v2.13.3**

David-Durst:
I think the following should work:
```
Include(FetchContent)

FetchContent_Declare(
  Catch2
  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
  GIT_TAG        v2.13.1)

FetchContent_MakeAvailable(Catch2)

target_link_libraries(tests Catch2::Catch2)

list(APPEND CMAKE_MODULE_PATH ${catch2_SOURCE_DIR}/contrib)
include(CTest)
include(Catch)
catch_discover_tests()
```

Per https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md#automatic-test-registration, if Catch2 has been downloaded but not installed (i.e. what FetchContent_Declare and FetchContent_MakeAvailable does), then "you need to add them to your CMake module path."

It seems that since FetchContent makes the package available during configuration time and before building, the *.cmake files aren't created yet and are still in *.cmake.in form in https://github.com/catchorg/Catch2/tree/devel/CMake.

If a project maintainer has a better solution, let me know!

shaperilio:
> ```
> Include(FetchContent)
> 
> FetchContent_Declare(
>   Catch2
>   GIT_REPOSITORY https://github.com/catchorg/Catch2.git
>   GIT_TAG        v2.13.1)
> 
> FetchContent_MakeAvailable(Catch2)
> 
[...]

PR #2253, opened 2021-06-27
Update CMake doc for automatic test registration

vishesh:
## Description

`FetchContent` doesn't include `contrib` directory as part of `CMAKE_MODULE_PATH`. This results into `include(Catch)` to fail. This patch just updates the documentation describing how to do include the path, so the new users don't have to figure this out themselves.

Source: https://github.com/catchorg/Catch2/issues/2103#issuecomment-730626324

## GitHub Issues

#2103

### DIFF
--- a/docs/cmake-integration.md
+++ b/docs/cmake-integration.md
@@ -88,6 +88,18 @@ include(Catch)
 catch_discover_tests(foo)
 ```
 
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks, lgtm.

PR #2257, opened 2021-07-11
Allow sharded execution for bazel test sharding support

ChewyGumball:
## Description
This PR adds support for splitting test execution across multiple invocations of the test binary (otherwise known as test sharding). The tests to run are split evenly into the number of groups given by the new `Config::shardCount()`, and the group at the index given by the new `Config::shardIndex()` is executed. The intended use is that multiple processes will be run, each with the same shard count, but different shard indices. Sharding works for executing as well as listing tests, and is applied after the list of tests to execute has been created (after filters).

The main motivation is to allow [Bazel test sharding](https://docs.bazel.build/versions/main/test-encyclopedia.html#test-sharding) to work:

```c++
#define CATCH_CONFIG_RUNNER
#include "catch.hpp"

#include <cstdlib>
#include <filesystem>

int main( int argc, char* argv[] )
{
  Catch::Session session;

  int returnCode = session.applyCommandLine( argc, argv );
  if( returnCode != 0 )
        return returnCode;
 
  // writing to session.configData() or session.Config() here 
  // overrides command line args
  // only do this if you know you need to

  char* envShardCount = std::getenv("TEST_TOTAL_SHARDS");
  if (envShardCount != nullptr && session.configData().shardCount != 1) {
        session.configData().shardCount = std::stoul(envShardCount);
  }

  char* envShardIndex = std::getenv("TEST_SHARD_INDEX");
  if (envShardIndex != nullptr && session.configData().shardIndex != 0) {
        session.configData().shardIndex = std::stoul(envShardIndex);
  }

  char* envShardStatusFile = std::getenv("TEST_SHARD_STATUS_FILE");
  if (envShardStatusFile != nullptr) {
      std::filesystem::last_write_time(envShardStatusFile, std::chrono::file_clock::now());
  }

  int numFailed = session.run();
  
  // numFailed is clamped to 255 as some unices only use the lower 8 bits.
  // This clamping has already been applied, so just return it here
  // You can also do any post run clean-up here
  return numFailed;
}
```


tests/TestScripts/testSharding.py
```diff
@@ -0,0 +1,141 @@
...
+def main():
+    self_test_exe, = sys.argv[1:]
+
+    # We want a random seed for the test, but want to avoid 0, because it has special meaning
+    shard_counts = [1, 5]
+    seeds = [random.randint(1, 2 ** 32 - 1), random.randint(1, 2 ** 32 - 1)]
+    tags = [["generators"], ["generators", "matchers"], []]
+    orders = ["rand", "decl", "lex"]
+
+    test_cases = [TestCase(*t) for t in itertools.product(shard_counts, orders, tags, seeds)]
```
ChewyGumball: This test takes a long time to run without having many cores available due to the number of combinations created here. If we want to keep the test run time shorter, we could do one or more of the following:

- have only one seed
- have only `rand` and `decl` orders
- remove one of the non empty tags

docs/command-line.md
```diff
@@ -418,6 +421,16 @@ processing of output.
 `--use-colour yes` forces coloured output, `--use-colour no` disables coloured
 output. The default behaviour is `--use-colour auto`.
 
+<a id="test-sharding"></a>
+## Test Sharding
+<pre>--shard-count <#number of shards>, --shard-index <#shard index to run></pre>
```
horenmar, line 433: A feature version placeholder should be somewhere around here.

docs/command-line.md
```diff
@@ -418,6 +421,16 @@ processing of output.
...
+When `--shard-count <#number of shards>` is used, the tests to execute will be split evenly in to the given number of sets,
+identified by indicies starting at 0. The tests in the set given by `--shard-index <#shard index to run>` will be executed.
+The default shard count is `1`, and the default index to run is `0`. If the shard index is greater than the number of shards,
+the last shard will be run.
```
horenmar: Is there an external reason for this? I would much rather have the runner report an error and exit early if this happens.

src/catch2/internal/catch_sharding.hpp
```diff
@@ -0,0 +1,49 @@
...
+#define CATCH_SHARDING_HPP_INCLUDED
+
+#include <catch2/catch_session.hpp>
+
+#include <cmath>
+
+namespace Catch {
+
+    template<typename CONTAINER>
+    CONTAINER createShard(CONTAINER const& container, IConfig const& config) {
```
horenmar: Nitpick: I would rather see the shard size and shard count passed directly.

src/catch2/internal/catch_sharding.hpp
```diff
@@ -0,0 +1,49 @@
...
+        if (config.shardCount() > 1) {
```
horenmar: I think this would end up being more readable with the braches swapped and the common case be an early exit.

src/catch2/internal/catch_sharding.hpp
```diff
@@ -0,0 +1,49 @@
...
+#ifndef CATCH_SHARDING_HPP_INCLUDED
+#define CATCH_SHARDING_HPP_INCLUDED
+
+#include <catch2/catch_session.hpp>
+
+#include <cmath>
+
+namespace Catch {
+
+    template<typename CONTAINER>
```
horenmar:
```suggestion
    template<typename Container>
```

and the rest of the usage should be changed accordingly.

tests/TestScripts/testSharding.py
horenmar: Yeah, that's waaaay too much work for this sort of test.

What I would prefer to see:
* Add unit tests within SelfTest for checking various edge cases.
* Have this script only check that partitioning was integrated with the runner, running the binary just few times. This means that this script would be running for single seed and single tag selector.

tests/TestScripts/testSharding.py
```diff
@@ -0,0 +1,143 @@
...
+    check_listed_and_executed_tests_match(sharded_tests, executed_tests)
+
+
+def main():
+    self_test_exe, = sys.argv[1:]
+
+    # We want a random seed for the test, but want to avoid 0, because it has special meaning
+    shard_counts = [1, 5]
+    seeds = [random.randint(1, 2 ** 32 - 1), random.randint(1, 2 ** 32 - 1)]
+    tags = [["generators"], ["generators", "matchers"], []]
```
horenmar: Whichever tags end up being used, also add `~benchmarks`, in case some are added in the future for these tags.

Running even single benchmark adds significant runtime through measuring the execution environment.

tests/CMakeLists.txt
```diff
@@ -309,6 +309,9 @@ set_tests_properties(TagAlias PROPERTIES
 add_test(NAME RandomTestOrdering COMMAND ${PYTHON_EXECUTABLE}
   ${CATCH_DIR}/tests/TestScripts/testRandomOrder.py $<TARGET_FILE:SelfTest>)
 
+add_test(NAME TestSharding COMMAND ${PYTHON_EXECUTABLE}
+  ${CATCH_DIR}/tests/TestScripts/testSharding.py $<TARGET_FILE:SelfTest>)
```
horenmar: Depending on how expensive the test ends up being after the changes, this might have to go into `ExtraTests`.

src/catch2/internal/catch_sharding.hpp
```diff
@@ -0,0 +1,49 @@
...
+            auto startIterator = std::next(container.begin(), startIndex);
+            auto endIterator = std::next(container.begin(), endIndex);
+
+            // Since we are calculating the end index with floating point numbers, but flooring 
+            // the value, we can't guarantee that the end index of the last shard lines up exactly
+            // with the end of input container. If we want the last shard, force the end index to
+            // be the end of the input container.
+            if (shardIndex == shardCount - 1) {
+                endIterator = container.end();
+            }
```
horenmar: I don't like the messing around with the floating point and then fixing away the potential error from floating point rounding.

Something like this should calculate the proper indices for a shard without running into these issues

```cpp
size_t testCount = container.size();
size_t shardCount = ...;

const size_t shardSize = testCount / shardCount;
const size_t leftoverTests = testCount % shardCount;

const size_t startIdx = shardIdx * shardSize + std::min(shardIdx, leftoverTests);
const size_t endIdx = (shardIdx + 1) * shardSize + std::min(shardIdx + 1, leftoverTests);
```

(this front-loads the extra tests, so that splitting 7 tests into 5 shards gives you shard of size [2, 2, 1, 1, 1])

horenmar (changes requested):
Hi, didn't have time to go through this sooner.

I like the general idea, but there is a bunch of changes to be done before merging.

docs/command-line.md
ChewyGumball: Do you mean something like in the benchmark sections: 
```
> [Introduced](https://github.com/catchorg/Catch2/pull/1844) in Catch2 2.11.2.
```
?

docs/command-line.md
ChewyGumball: There is no external reason for this. I made the command line parser report an error if this happens from command line arguments as you suggest, and used an assert in the `createShards` function to ensure this precondition in case the config data is modified directly.

docs/command-line.md
horenmar: Yes, but with the right format so that the placeholder is recognized and updated when release is made. See [contributing docs](https://github.com/catchorg/Catch2/blob/devel/docs/contributing.md#technicalities), version tags.

ChewyGumball:
All great comments. I've tried my best to address them in the latest commit. Would you mind taking another look?

tests/SelfTest/IntrospectiveTests/CmdLine.tests.cpp
```diff
@@ -569,6 +569,27 @@ TEST_CASE( "Process can be configured on command line", "[config][command-line]"
...
+        }
+
+        SECTION("shard-index") {
+            CHECK(cli.parse({ "test", "--shard-index=2"}));
+
+            REQUIRE(config.shardIndex == 2);
+        }
+
+        SECTION("Zero shard-count") {
+            auto result = cli.parse({ "test", "--shard-count=0"});
```
horenmar: Without digging too deep, I suspect that providing a negative number would pass, but result in a nonsensically large shard count.

Whether that is or isn't true, there should be a test for it.

src/catch2/internal/catch_commandline.cpp
```diff
@@ -218,6 +227,12 @@ namespace Catch {
             | Opt( config.benchmarkWarmupTime, "benchmarkWarmupTime" )
                 ["--benchmark-warmup-time"]
                 ( "amount of time in milliseconds spent on warming up each test (default: 100)" )
+            | Opt( setShardCount, "shard count" )
+                ["--shard-count"]
+                ( "split the tests to execute into this many groups" )
+            | Opt( config.shardIndex, "shard index" )
```
horenmar: Why is one of these done with custom function and the other one isn't?

docs/command-line.md
```diff
@@ -418,6 +421,18 @@ processing of output.
...
+## Test Sharding
+<pre>--shard-count <#number of shards>, --shard-index <#shard index to run></pre>
+When `--shard-count <#number of shards>` is used, the tests to execute will be split evenly in to the given number of sets,
+identified by indicies starting at 0. The tests in the set given by `--shard-index <#shard index to run>` will be executed.
+The default shard count is `1`, and the default index to run is `0`. It is an error to specify a shard index greater than
+the number of shards.
+
+This is useful when you want to split test execution across multiple processes, as is done with [Bazel test sharding](https://docs.bazel.build/versions/main/test-encyclopedia.html#test-sharding). 
+
+> [Introduced](https://github.com/catchorg/Catch2/pull/2257) in Catch2 X.Y.Z.
```
horenmar: Nitpick: I would prefer this to be under the header for this section.

horenmar:
I probably won't get to the tests till tomorrow.

tests/SelfTest/IntrospectiveTests/Sharding.tests.cpp
```diff
@@ -0,0 +1,41 @@
...
+ */
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators_all.hpp>
+
+#include <catch2/internal/catch_sharding.hpp>
+
+#include <unordered_map>
+#include <vector>
+
+TEST_CASE("Sharding Function") {
```
horenmar:
```suggestion
TEST_CASE("Sharding Function", "[approvals]") {
```

The approval tests are getting too big, I will be cutting them down a bit later, but for now I don't think these need to be included.

horenmar:
Thanks, it looks pretty good now, but I think you went overboard on reducing the test script.

What I would want to see is:

* Pick a positive tag (generators is fine)
* List all tests for that tag without sharding
* Shard the tests into say, 5 shards, list each shard individually, check that they add up into the whole list, and without overlap.
* Shard the tests into the same numbers of shards, check that executing each shard executes the tests listed in previous step.


Since the expensive part is calling into the binary, rather than doing extra processing of the data on the Python side, it is possible to also add extra checks along the way, e.g. checking that the binary obeys `--order` selection when sharding and so on, because that only adds some extra processing in Python.

horenmar (approved):

PR #2269, opened 2021-08-10
Rename contrib/ to extras/ in the docs

numaru:
The directory has been renamed in 918aa32 but the doc was not updated yet.


horenmar:
Good find, thanks.

PR #2278, opened 2021-08-26
Differentiate between debug and release outputs

ror3d:
## Description
I was trying to use Catch2 as an imported library (not a subproject) in my project's cmake (visual studio) and when the build configuration didn't match the one used to build Catch2 it complained about mismatched `_ITERATOR_DEBUG_LEVEL` and `/MDd` vs `/MD`. This PR adds a postfix to the lib files produced by building Catch so that the debug and release versions are differentiated and it can then be imported without the issue.


horenmar:
Sure, why not.

PR #2280, opened 2021-08-28
Fix warning suppressions leaking under clang.exe

jorgenpt:
## Description
When running clang.exe under Windows, catch.hpp leaks warning
suppressions because it uses `#pragma warning(push)` & `#pragma
warning(pop)` around warning suppressions like `#pragma clang diagnostic
ignore "-Wunused-variable"`, instead of using `#pragma clang diagnostic
push` and `#pragma clang diagnostic pop`.

This fixes that by only defining
`CATCH_INTERNAL_START_WARNINGS_SUPPRESSION` and
`CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION` to be the cl.exe variants if
`defined(_MSC_VER) && !defined(__clang__)`.

### DIFF
--- a/include/internal/catch_compiler_capabilities.h
+++ b/include/internal/catch_compiler_capabilities.h
@@ -146,10 +146,6 @@
 // Visual C++
 #if defined(_MSC_VER)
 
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks

PR #2284, opened 2021-09-08
pkgconfig: Add missing entries

Biswa96:
This adds prefix, exec_prefix and libdir fields


Biswa96:
The prefix and exec_prefix are also needed in devel branch. Should I create another PR for devel branch?

horenmar:
Thanks, I don't actually use pkg-config, so bugs there might be around for a long time.

> The prefix and exec_prefix are also needed in devel branch. Should I create another PR for devel branch?

Yes.

PR #2293, opened 2021-09-29
Bring back useful comment

knatten:
In b7b346c3e56030 this conditional was simplified to just
`while( false)` rather than the current one. Then in 3a33315ff8e04, that
change was reverted. I found it hard to understand this
complicated conditional, but after some digging in history I found this
comment which used to be here. It was removed in b7b346c3e56030, but not
restored together with the revert in 3a33315ff8e04. Let's revive it.


horenmar:
Sure, thanks.

PR #2297, opened 2021-10-03
Fixed #2272: Compilation failure with C++20

codeinred:
## Description
This commit fixes issue #2272: Compilation failure with C++20. 

To do this, it removes using-declarations for 

- `std::allocator<T>::address()`,
- `std::allocator<T>::construct()`, 
- `std::allocator<T>::max_size()`, and 
- `std::allocator<T>::destroy()` 

from the CustomAllocator class in Matchers.tests.cpp. [These functions were deprecated in C++17, and removed in C++20](https://en.cppreference.com/w/cpp/memory/allocator), so having a using declaration for them causes compilation error. The functions weren't used in the Catch2 code base in earlier versions of C++, so it's not a breaking change. 

This commit also wraps the use of a comma operator inside the bracket operator in *Message.tests.cpp*. This change is necessary because gcc was issuing an error when compiling tests with `-std=c++20`. 

```cpp
CAPTURE(std::vector<int>{1, 2, 3}[0, 1, 2],
```
Becomes
```cpp
CAPTURE(std::vector<int>{1, 2, 3}[(0, 1, 2)],
```
<details><summary>Click to see compilation error</summary>

```
projects/CMakeFiles/SelfTest.dir/SelfTest/UsageTests/Message.tests.cpp.o 
/usr/bin/c++  -I/home/alecto/cowork/Catch2/include -std=c++20 -Wall -Wextra -Wunreachable-code -Wpedantic -Wmissing-declarations -Werror -MD -MT projects/CMakeFiles/SelfTest.dir/SelfTest/UsageTests/Message.tests.cpp.o -MF projects/CMakeFiles/SelfTest.dir/SelfTest/UsageTests/Message.tests.cpp.o.d -o projects/CMakeFiles/SelfTest.dir/SelfTest/UsageTests/Message.tests.cpp.o -c /home/alecto/cowork/Catch2/projects/SelfTest/UsageTests/Message.tests.cpp
In file included from /home/alecto/cowork/Catch2/include/catch.hpp:54,
                 from /home/alecto/cowork/Catch2/projects/SelfTest/UsageTests/Message.tests.cpp:9:
/home/alecto/cowork/Catch2/projects/SelfTest/UsageTests/Message.tests.cpp: In function ‘void ____C_A_T_C_H____T_E_S_T____76()’:
/home/alecto/cowork/Catch2/projects/SelfTest/UsageTests/Message.tests.cpp:245:40: error: top-level comma expression in array subscript is deprecated [-Werror=comma-subscript]
[...]

Issue #2272, opened 2021-08-24
Compilation failure with C++20


stac47:
I tried to build the tests of Catch2 with latest gcc 11.2.1 and C++20 standard.

I got those two errors:
```
In file included from /home/docker/opensource/Catch2/include/catch.hpp:54,
                 from /home/docker/opensource/Catch2/projects/SelfTest/UsageTests/Message.tests.cpp:9:
/home/docker/opensource/Catch2/projects/SelfTest/UsageTests/Message.tests.cpp: In function 'void ____C_A_T_C_H____T_E_S_T____76()':
/home/docker/opensource/Catch2/projects/SelfTest/UsageTests/Message.tests.cpp:245:40: error: top-level comma expression in array subscript is deprecated [-Werror=comma-subscript]
  245 |     CAPTURE(std::vector<int>{1, 2, 3}[0, 1, 2],
      |                                        ^
```
Which is related to http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1161r3.html
```
/home/docker/opensource/Catch2/projects/SelfTest/UsageTests/Matchers.tests.cpp:245:38: error: 'address' has not been declared in 'class std::allocator<int>'
  245 |             using std::allocator<T>::address;
      |                                      ^~~~~~~
/home/docker/opensource/Catch2/projects/SelfTest/UsageTests/Matchers.tests.cpp:247:38: error: 'construct' has not been declared in 'class std::allocator<int>'
  247 |             using std::allocator<T>::construct;
      |                                      ^~~~~~~~~
/home/docker/opensource/Catch2/projects/SelfTest/UsageTests/Matchers.tests.cpp:249:38: error: 'max_size' has not been declared in 'class std::allocator<int>'
  249 |             using std::allocator<T>::max_size;
      |                                      ^~~~~~~~
/home/docker/opensource/Catch2/projects/SelfTest/UsageTests/Matchers.tests.cpp:250:38: error: 'destroy' has not been declared in 'class std::allocator<int>'
  250 |             using std::allocator<T>::destroy;
      |                                      ^~~~~~~
```
Because it has been removed in C++20 from std::allocator.

Platform:
 - OS: GNU/Linux Ubuntu 20.04
 - Compiler+version: GCC 11.2.1
 - Catch version: HEAD of v2.x branch

Regards,

horenmar:
[...]

PR #2300, opened 2021-10-11
Apply PR #2297 to devel branch

codeinred:
## Description
It turns out that Issue #2272 partially affected the devel branch. When building tests with C++20, the compiler emits a warning that top-level comma expressions in array subscripts are depricated. Warnings are treated as errors, so this caused the build to fail.

This commit adds localized warning suppression [in accordance with this recommendation here.](https://github.com/catchorg/Catch2/pull/2297#discussion_r720848392)

## GitHub Issues
It turns out that Issue #2272 also partially applied to the devel branch. This issue was marked as closed, but the fix was only ever applied to the 2.x release branch, and still exists on the devel branch.


codeinred:
It's important to note that part of the fix applied in #2297 (removing using declarations for deleted functions inside the custom allocator class) wasn't necessary for the devel branch.

horenmar:
Thanks

PR #2313, opened 2021-10-27
Fix clang analyzer warning about FilterGenerator

GHF:
**Describe the bug**
When using a filter generator, clang analyzer (called through clang-tidy) produces a diagnostic about `FilterGenerator::next`.

```
/Users/xw/Projects/mays/build/_deps/catch2-src/single_include/catch2/catch.hpp:4166:42: warning: Call to virtual method 'FilterGenerator::next' during construction bypasses virtual dispatch [clang-analyzer-optin.cplusplus.VirtualCall]
                auto has_initial_value = next();
                                         ^
/Users/xw/Projects/mays/mays/scale_test.cc:164:16: note: Calling 'filter<int, (lambda at /Users/xw/Projects/mays/mays/scale_test.cc:163:25)>'
          100, filter([](int i) { return i != 0; },
               ^
/Users/xw/Projects/mays/build/_deps/catch2-src/single_include/catch2/catch.hpp:4101:97: note: expanded from macro 'GENERATE'
                                 [ ]{ using namespace Catch::Generators; return makeGenerators( __VA_ARGS__ ); } ) //NOLINT(google-build-using-namespace)
                                                                                                ^~~~~~~~~~~
/Users/xw/Projects/mays/build/_deps/catch2-src/single_include/catch2/catch.hpp:4189:67: note: Calling 'make_unique<Catch::Generators::FilterGenerator<int, (lambda at /Users/xw/Projects/mays/mays/scale_test.cc:163:25)>, (lambda at /Users/xw/Projects/mays/mays/scale_test.cc:163:25), Catch::Generators::GeneratorWrapper<int>>'
        return GeneratorWrapper<T>(std::unique_ptr<IGenerator<T>>(pf::make_unique<FilterGenerator<T, Predicate>>(std::forward<Predicate>(pred), std::move(generator))));
                                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/xw/Projects/mays/build/_deps/catch2-src/single_include/catch2/catch.hpp:3934:43: note: Calling constructor for 'FilterGenerator<int, (lambda at /Users/xw/Projects/mays/mays/scale_test.cc:163:25)>'
            return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
[...]

PR #2314, opened 2021-10-27
Fix clang analyzer warning about FilterGenerator in v2.x

GHF:
PR #2313 backported to v2.x since that's the one I'm actually using.

### DIFF
--- a/include/internal/catch_generators_generic.hpp
+++ b/include/internal/catch_generators_generic.hpp
@@ -63,7 +63,7 @@ namespace Generators {
             if (!m_predicate(m_generator.get())) {
                 // It might happen that there are no values that pass the
                 // filter. In that case we throw an exception.
-                auto has_initial_value = next();
+                auto has_initial_value = nextImpl();
                 if (!has_initial_value) {
                     Catch::throw_exception(GeneratorException("No valid value found in filtered generator"));
                 }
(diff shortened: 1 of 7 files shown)


include/internal/catch_generators_generic.hpp
```diff
@@ -51,7 +51,7 @@ namespace Generators {
 
 
     template <typename T, typename Predicate>
-    class FilterGenerator : public IGenerator<T> {
```
horenmar: This is, very very hypothetically, an API breaking change.

I think it is very low risk, but there is a bit uglier but guaranteed non-breaking change that also fixes this: turn the current implementation of `next` into a private helper, e.g. `nextImpl`, and have both the constructor and the new `next` implementation call it directly, e.g.

```
bool next() override {
    return nextImpl();
}
```

GHF: > This is, very very hypothetically, an API breaking change.
> 
> I think it is very low risk, but there is a bit uglier but guaranteed non-breaking change that also fixes this: turn the current implementation of `next` into a private helper, e.g. `nextImpl`, and have both the constructor and the new `next` implementation call it directly, e.g.

Thanks for pointing out that important consideration. I think this is your call. Would you like me to revise this PR for the `nextImpl` route?

horenmar: Yes, sorry for making it implicit only.

I would merge a PR that adds a test or two (e.g. the ones I added to `devel` recently) and goes the `nextImpl` route.

PR #2315, opened 2021-10-28
Fix performance when printing strings w/o newlines

jorgenpt:
## Description
When generating columns from long strings, as e.g. done when we
`REQUIRE(str == other_str)`, every single column would calculate the
length of the current line. This can be very slow if used for large
strings, due to the O(n²) performance. We had a test comparing two 2
megabyte long strings, and even when logging directly to file, this test
took over an hour to run with `-s` on 2.x. With this change, it takes
~4s.

This change makes us only recalculate the length of the current line
when we've consumed the current line, which means that unbroken strings
will only have a single iteration over the line (and that each line will
only be traversed once). `m_end` would always be the same after every
`calcLength` until we encounter the next newline in our output.

### DIFF
--- a/src/catch2/internal/catch_textflow.cpp
+++ b/src/catch2/internal/catch_textflow.cpp
@@ -42,14 +42,16 @@ namespace Catch {
         void Column::iterator::calcLength() {
             m_suffix = false;
             auto width = m_column.m_width - indent();
-            m_end = m_pos;
             std::string const& current_line = m_column.m_string;
-            if ( current_line[m_pos] == '\n' ) {
-                ++m_end;
-            }
-            while ( m_end < current_line.size() &&
-                    current_line[m_end] != '\n' ) {
-                ++m_end;
+            if ( m_end <= m_pos ) {
+                m_end = m_pos;
+                if ( current_line[m_pos] == '\n' ) {
+                    ++m_end;
+                }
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks.

It will take me a while to review, because that part of the codebase is ... undertested ... and been shown to contain hidden landmines.

jorgenpt:
@horenmar I figured -- I tried my best to make this demonstrably not change the functionality at all. Instead of any of this change I considered changing:

```cpp
while ( m_end < current_line.size() &&
        current_line[m_end] != '\n' ) {
```

to

```cpp
const auto max_end = std::min(current_line.size(), width);
while ( m_end < max_end && current_line[m_end] != '\n' ) {
```

instead -- this would make it so it *never* has to iterate the whole string for a single column and generally does less work, but the risk was that `m_end` is used in `assert( m_pos <= m_end );` in `operator*`, and I found it somewhat challenging to verify that the assert would still be valid if `m_end`'s meaning was changed from "end of the next line" to "end of this column". If you end up digging in to it, might be worth considering that approach instead.

horenmar:
So the good part is that I cleaned up the code a bit and added tests. The bad part is that I found some fun bugs and realized that I don't even know what should be the correct behaviour for some of the inputs.

However, this change won't make it worse so I am going to merge it when I have time.

PR #2317, opened 2021-11-04
Handle dynamic stack size on GLIBC > 2.33

### LINKED ISSUE #2178: Catch2 fails to compile on Fedora Rawhide
**Describe the bug**
Tests that use Catch2 fail to compile on Fedora Rawhide.

**Expected behavior**
Tests that use Catch2 compile and execute without any problems on Fedora Rawhide.

**Reproduction steps**
Try to compile the factorial example from the [Catch2 tutorial](https://github.com/catchorg/Catch2/blob/devel/docs/tutorial.md#writing-tests):
```gcc
$ c++ test.cpp
In file included from /usr/include/signal.h:315,
                 from catch.hpp:8034,
                 from test.cpp:2:
catch.hpp:10822:58: error: call to non-'constexpr' function 'long int sysconf(int)'
10822 |     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
      |                                                          ^~~~~~~~~~~
In file included from /usr/include/bits/sigstksz.h:24,
                 from /usr/include/signal.h:315,
                 from catch.hpp:8034,
                 from test.cpp:2:
/usr/include/unistd.h:630:17: note: 'long int sysconf(int)' declared here
  630 | extern long int sysconf (int __name) __THROW;
      |                 ^~~~~~~
In file included from test.cpp:2:
catch.hpp:10881:45: error: size of array 'altStackMem' is not an integral constant-expression
10881 |     char FatalConditionHandler::altStackMem[sigStackSize] = {};
      |                                             ^~~~~~~~~~~~
```

**Platform information:**

 - OS: **latest Fedora Rawhide**
 - Compiler+version: **gcc (GCC) 11.0.0 20210210 (Red Hat 11.0.0-0)**
 - Catch version: **v2.13.4**

**Additional context**
This behaviour seems to be related to this change in `glibc`: https://sourceware.org/git/?p=glibc.git;a=commit;h=6c57d320484988e87e446e2e60ce42816bf51d53

michel-slm:
## Description
On GLIBC 2.33, `MINSIGSTKSZ` is not a constant. This can be detected
at compile time by the presence of either `_SC_SIGSTKSZ_SOURCE` or
`_GNU_SOURCE`.

If either are define, hardcode the size we want rather than compare it,
since we can't do that in a constant expression anymore.

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>

## GitHub Issues
Closes #2178

### DIFF
--- a/extras/catch_amalgamated.cpp
+++ b/extras/catch_amalgamated.cpp
@@ -3750,8 +3750,14 @@ namespace Catch {
     };
 
     // 32kb for the alternate stack seems to be sufficient. However, this value
-    // is experimentally determined, so that's not guaranteed.
+    // is experimentally determines, so that's not guaranteed.
+#if defined(_SC_SIGSTKSZ_SOURCE) || defined(_GNU_SOURCE)
+    // on glibc > 2.33 this is no longer constant, see
+    // https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=85e84fe53699fe9e392edffa993612ce08b2954a;hb=HEAD
+    static constexpr std::size_t sigStackSize = 32768;
+#else
     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
+#endif
 
     static SignalDefs signalDefs[] = {
         { SIGINT,  "SIGINT - Terminal interrupt signal" },


michel-slm:
note: I maintain the [Eternal Terminal](https://github.com/MisterTea/EternalTerminal) package for Fedora, and have had to carry an out-of-tree patch for its bundled Catch2 since Fedora 35 was in development

horenmar:
Thanks for the effort, but the amalg. files (and for v2.x branch the single header file) are only updated on release.

Given that the problem has already been fixed in the devel branch (I actually spent some time wondering wtf, knowing that I fixed this issue already, until I noticed you edited the `catch_amalgamated.cpp` file), you want to regenerate the files and then update the bundled version in the Eternal Terminal package.

PR #2320, opened 2021-11-17
Fix template args to avoid conversion warnings

sketchylizard:
Fix avoids a warning about sign conversion when included from a file compiled with -Wsign-conversion.

## Description
Changed integer template parameter from int to size_t. This avoids a warning when header is included into a file using -Wsign-conversion.


horenmar:
Thanks

PR #2322, opened 2021-11-22
Fix link in event-listeners.md

tstenner:
`catch_reporter_registrars.hpp` was moved; this PR updates the example in `event-listeners.md`


horenmar:
Thanks, I updated the path in sources, but forgot to check the docs.

PR #2329, opened 2021-12-11
Fix minor errors in tutorial docs

soumitradev:
Some code blocks didn't use the updated name Contains()


horenmar:
Thank you, but this change breaks the docs.

Between the last preview release and current devel, `Contains` for strings has been renamed to `ContainsSubstring`, as the name collision between the overload set couldn't have been resolved otherwise.

PR #2333, opened 2021-12-15
Add Bazel `XML_OUTPUT_FILE` processing to `main`

mattyclarkson:
`bazel test` sets the `XML_OUTPUT_FILE` environment variable which
specifies where the JUnit output of the test executable should be
written.

Picking up this variable allows catch2 to naturally integrate into
the Bazel testing ecosystem out-of-the-box.

The environment variable processing is hidden behind the
`CATCH_CONFIG_MAIN_BAZEL_XML_OUTPUT_FILE` define and it _only_
enabled in the `BUILD.bazel`.

This allows the Bazel ecosystem to create a `cc_test` target as
follows:

```
cc_test(
    name = "some_test",
    deps = ["@catch2//:catch2_main"],
    srcs = globs("**/*_test.cpp"),
)
```

The output will be written to `bazel-testlogs/some_test/test.xml`

### DIFF
--- a/src/catch2/internal/catch_main.cpp
+++ b/src/catch2/internal/catch_main.cpp
@@ -33,7 +33,18 @@ int main (int argc, char * argv[]) {
     // and its constructor, as it (optionally) registers leak detector
     (void)&Catch::leakDetector;
 
-    return Catch::Session().run( argc, argv );
+    Catch::Session session;
+
+#if defined(CATCH_CONFIG_MAIN_BAZEL_JUNIT_XML_OUTPUT_FILE)
+    // Bazel expects test output to be written to this file in JUnit format.
+    auto bazelOutputFile = std::getenv("XML_OUTPUT_FILE");
+    if (bazelOutputFile != nullptr) {
+        session.configData().reporterName = "junit";
+        session.configData().outputFilename = bazelOutputFile;
+    }
+#endif
+
+    return session.run( argc, argv );
 }
 
 #endif // !defined(CATCH_AMALGAMATED_CUSTOM_MAIN
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -16,6 +16,7 @@ cc_library(
 # Static library, with main.
 cc_library(
     name = "catch2_main",
+    local_defines = ["CATCH_CONFIG_MAIN_BAZEL_JUNIT_XML_OUTPUT_FILE"],
     srcs = ["src/catch2/internal/catch_main.cpp"],
     deps = [":catch2"],
     visibility = ["//visibility:public"],


mattyclarkson:
RFC: this is quite a punt to allow Catch2 to integrate more tightly into the Bazel ecosystem out-of-the-box.

It's a bit overzealous about respecting `XML_OUTPUT_FILE`, included in the default `main` and currently modifies the Bazel `:catch2_main` target.

Happy to create a _new_ Bazel target (`:catch2_bazel_main`?), move the code whereever and be more defensive.

Also happy for it to be rejected: we can add the patches downstream in the Bazel Central Registry, if needed.

jesseschalken:
Does this override Catch2's default console output when run via Bazel? If so it would be nice if there was output when run with `--test_output=streamed` in addition to writing to `$XML_OUTPUT_FILE`.

mattyclarkson:
> Does this override Catch2's default console output

Yes. Changing the reporter to `junit` from `console` stops the console output.

Catch2 doesn't support mutliple reporters. There is a proposal in #1712 which would enable both JUnit output and console output.

src/catch2/internal/catch_main.cpp
```diff
@@ -33,7 +33,18 @@ int main (int argc, char * argv[]) {
     // and its constructor, as it (optionally) registers leak detector
     (void)&Catch::leakDetector;
 
-    return Catch::Session().run( argc, argv );
+    Catch::Session session;
+
+#if defined(CATCH_CONFIG_MAIN_BAZEL_JUNIT_XML_OUTPUT_FILE)
```
horenmar, line 38: That's a terrible define name. How about `CATCH_CONFIG_BAZEL_SUPPORT`? (Does Bazel have more of these customization env option?)

Also it needs to be documented in docs/configuration.md

src/catch2/internal/catch_main.cpp
```diff
@@ -33,7 +33,18 @@ int main (int argc, char * argv[]) {
...
+    // Bazel expects test output to be written to this file in JUnit format.
+    auto bazelOutputFile = std::getenv("XML_OUTPUT_FILE");
+    if (bazelOutputFile != nullptr) {
+        session.configData().reporterName = "junit";
+        session.configData().outputFilename = bazelOutputFile;
```
horenmar, line 43: This needs to be redone with the multireporter support.

horenmar (changes requested):
Shame Bazel didn't use properly namespaced env variable for this, it could've been enabled unconditionally. `XML_OUTPUT_FILE` is waaay too generic for that though.

Also this likely shouldn't be in the default-main, instead it should internally add another reporter (if the user doesn't want this behaviour, they can just not provide the define/env variable).

Oh and it obviously needs tests. We already have a bunch of separately-compiled tests in `tests/ExtraTests`, so you can look there at how to add new one.

src/catch2/internal/catch_main.cpp
mattyclarkson: Huh, cool. Will look into that.

src/catch2/internal/catch_main.cpp
mattyclarkson: Agreed. I just sorta went for _something_. Happy to spin to `CATCH_CONFIG_BAZEL_SUPPORT` if we don't spin out another Bazel specific `main`.

mattyclarkson:
> Shame Bazel didn't use properly namespaced env variable for this

Yep. I can ask upstream in Bazel world to see it could also export `BAZEL_XML_OUTPUT_FILE`.

> instead it should internally add another reporter

Like, a whole new Bazel reporter, i.e. `--reporter=bazel` that inherits from the JUnit reporter?

> Oh and it obviously needs tests

Yep. Was pushing this up with minimal code to see if it got straight up rejected. I can move forward with filling out he edges.

Thanks for the comments!

horenmar:
> Like, a whole new Bazel reporter, i.e. --reporter=bazel that inherits from the JUnit reporter?

No. I meant that Catch2 should internally build another reporter for Bazel when it detects the env variable, instead of having it be done in the main function.

mattyclarkson:
> instead of having it be done in the main function

So either we add another to the config when it is created or in the [session](https://github.com/catchorg/Catch2/blob/4ff8b27bb6bed4b8b309e56cd269b4f1fbc24e89/src/catch2/catch_session.cpp#L301)?

Something like:

* When config parsed
* If `BAZEL_XML_OUTPUT_FILE` present
* Add JUnit multi-reporter pointing at `BAZEL_XML_OUTPUT_FILE`

_or_

* When session started
* If `BAZEL_XML_OUTPUT_FILE` present
* Add JUnit multi-reporter pointing at `BAZEL_XML_OUTPUT_FILE`

The former might be nicer because dumping the configuration would show the extra JUnit reporter when `BAZEL_XML_OUTPUT_FILE` is present.

horenmar:
Either of these is fine honestly. I would probably go with inserting the new reporter as part of parsed config (I think the `ConfigData` instance could already be initialized with the reporter pair, but I haven't checked closely if that would run into some issues later).

horenmar:
I recently changed how the default reporter is handled, the Bazel reporter can likely be handled in the same way: https://github.com/catchorg/Catch2/commit/fc5552d27b57628f8faf6064ca6b305e2d3bf458

mattyclarkson:
Closing in favour of #2399

Issue #578, opened 2016-01-25
Use of reserved identifiers


melak47:
A quick scan shows at least `__catchResult` and `____C_A_T_C_H____T_E_S_T____` being used, 
but since these identifiers contain double underscores, they are reserved for 
the compiler/stdlib implementation: [[lex.name] 2.10.3](http://eel.is/c++draft/lex.name#3) ([global.names] 17.6.4.3.2 until C++17)

horenmar:
Not really a bug, but more of a language lawyer request.

AndrewPaxie:
LLVM commit b83b23275b745287bf9d3d72a93b593119f53f75 introduces
warning `-Wreserved-identifier`.
See: https://reviews.llvm.org/D93095
This is devastating to projects using Catch that compile with `-Weverything` on the Clang command line.
Example: trompeloeil.

j-stephan:
>  This is devastating to projects using Catch that compile with -Weverything on the Clang command line.

Indeed. Once clang ~~14~~ 13 becomes more widely used a lot of people will run into this.

davidstone:
Trying to set up a new project using the `add_subdirectory` CMake integration. I immediately run into this error because my project builds with this warning. This is using the Catch2 v3 devel branch.

```
In file included from /home/david/tm/Catch2/src/catch2/../catch2/catch_test_macros.hpp:11:
In file included from /home/david/tm/Catch2/src/catch2/../catch2/internal/catch_test_macro_impl.hpp:11:
In file included from /home/david/tm/Catch2/src/catch2/../catch2/internal/catch_assertion_handler.hpp:11:
In file included from /home/david/tm/Catch2/src/catch2/../catch2/catch_assertion_info.hpp:13:
/home/david/tm/Catch2/src/catch2/../catch2/internal/catch_stringref.hpp:101:32: error: identifier '_sr' is reserved because it starts with '_' at global scope [-Werror,-Wreserved-identifier]
    constexpr auto operator "" _sr( char const* rawChars, std::size_t size ) noexcept -> StringRef {
                   ~~~~~~~~~~~~^~~
                   operator""_sr
/home/david/tm/Catch2/src/catch2/../catch2/internal/catch_stringref.hpp:106:28: error: identifier '_catch_sr' is reserved because it starts with '_' at global scope [-Werror,-Wreserved-identifier]
constexpr auto operator "" _catch_sr( char const* rawChars, std::size_t size ) noexcept -> Catch::StringRef {
[...]

PR #2336, opened 2021-12-16
Avoid reserved identifiers in macros

bernhardmgruber:
## Description
This PR renames all macros that are flagged by clang with `-Werror -Wreserved-identifier`. Leading underscores are removed, multi-underscores inside macros reduced to one.

With the proposed changes, Catch2 v2 can be used in a project which enables `-Werror -Wreserved-identifier` with clang-13 and above.

## GitHub Issues
Fixes part of #578.


horenmar:
Thanks. I would've used different names, but these are fine too.

Issue #1579, opened 2019-03-23
How to get test case names in output


pwinston:
I expected one output option would be like this:
    MyTestCase1: 5 assertions passed
    MyTestCase2: 3 assertions passed
    ...
or even just
    MyTestCase1: PASS
    MyTestCase2: PASS
    ....

I figured the log should list the tests that were performed, not just "passed 182 test cases" since that doesn't tell you what tests were in place at the time.

I see the junit and xml reporters to include the test case names. Are there settings that produce a simple list like above? Or is it necessary to transform the xml, or write a reporter? Or is this not a useful output format to ask for for some reason?

pwinston:
I guess this is pretty close `./runner --list-test-names-only || ./runner` but still wondering if there is something like I'm saying, with each test case name followed by the result.

horenmar:
If you want your output to contain passed assertions, you can try the `-s` flag. However, having successful tests be a part of the output leads to an unreadable mess fairly quickly -- I would consider the test suite at my current employer fairly modest, but the summary line is 
```
Passed all 3889 test cases with 28179 assertions.
```
so having a line per test case would make console output pretty unusable.



If you want your own bespoke output format, then that's what the Reporter interface is for.

pwinston:
Thanks for the response. I think `-s` is too much information, that's essentially echoing lines of code, where the test names are hand-chosen and presumably human readable.

I see your point about having 4,000 tests, but put yourself in the shoes of a new user with 20 tests: seeing the test names printed seems reasonable to me. From what I see of Google Test docs they print test names in their default output, I didn't install google test, but looked that way.

I'm not going to press the issue. Just giving the feedback that as a new user I felt 100% sure this type of output must be available in the system, it just seemed logical to me, and spent a long time looking for it. We'll probably just do ` --list-test-names-only` as a way to record what tests were run, I think that's sufficient for our needs so that's good.

biodranik:
+1 to have names of running tests printed as a non-default option.

codypiersall:
This is done as a side effect of passing `-d yes`:

```console
root@microzed-zynq7 ~ ./unittests -d yes
0.000 s: Can instantiate Thing
0.000 s: Structs are the right size
===============================================================================
All tests passed (12 assertions in 2 test cases)
```

guban:
[...]

PR #2343, opened 2022-01-02
Add gtest console reporter

mmatrosov:
## Description

The default console reporter is a good choice when tests are fast and quiet. In many projects that I worked with neither is true. A test might take up to several seconds and the component under test might log something to console. The default reporter does not work good in these cases. Either you cannot understand what is being run now, or the output is cluttered. This desire manifested itself in #1579. Also, seems like the default reporter cares too much about assertions - how many passed, how many failed. However, I've never felt any need in this info in my practice. 

All these problems do not apply to the output of Google Test. That's why I decided to mimic its behavior in a new `gtest` reporter. Please see changes in `reporters.md` for details.

That's how the output looks for `SelfTest`:
* No filters: https://gist.github.com/mmatrosov/ecd15e02d7c1361acc2b229322070b75
* Filter is `"*fail*"`: https://gist.github.com/mmatrosov/8248a77d8b82623475e39dc84bdceba0

The initial code for `catch_reporter_gtest.cpp` was copy-pasted from `catch_reporter_console.cpp`. They don't have much code duplication in HEAD though. You can check out [the diff for the file excluding the first commit](https://github.com/catchorg/Catch2/pull/2343/files/f2d3a584cefc155aea9b3b01bab41c6009d4c05a..845adeee3b6ba281abe448e72433872060f201e6#diff-4fb79bcd900e95c4d6982bf0689d0f597daae9c21c44f2ec2bf6f4a45ed67d55) to see the duplicated code.

This PR is for `v2.x` branch, because I don't understand the state of v3. There are no stable releases for it, the last releaes was more than a year ago, and it is not available in conan-center. But I want to use this feature in the nearest future, that's why I am targeting v2.

## GitHub Issues
Should solve  #1579.


mmatrosov:
I am not sure how to resolve the coverage problems. Should I include the new reporter in the test? If yes, how do I do it?

If I don't add the new reporter to the test I can fix the "project" check by adding some tests for `catch_string_manip.cpp`, but I won't be able to fix the "patch" check I suppose.

mmatrosov:
@horenmar could you please take a look? Seems like the run of `ctest` should produce all the coverage results, but when I run it locally, it only uses the default and the "compact" reporters, even though I see how all existing reporters are covered in codecov: https://app.codecov.io/gh/catchorg/Catch2/compare/2343/tree/include/reporters
But I don't get how coverage for other reporters is collected.

horenmar:
Thanks for the effort, but I do not want to merge gtest format reporter. Nowadays I only want to merge reporters that implement a well documented format used by some other tools.

As to the coverage, reporters are primarily tested through ApprovalTests, which are implemented in `scripts/approvalTests.py`, and they have to be added manually.

sketch34:
> The default reporter does not work good in these cases. Either you cannot understand what is being run now, or the output is cluttered.

Agreed. I am running into this problem with two of my current projects as well.

mmatrosov:
Yeah. I am disappointed this change was not merged. We maintain it as a patch in our custom conan recipe for catch2.

dengchj:
> Yeah. I am disappointed this change was not merged. We maintain it as a patch in our custom conan recipe for catch2.

You shouldn't have just called gtest. If you change the name but the implementation is similar, it will probably be accepted :)

PR #2346, opened 2022-01-05
Allow to specify CATCH_CONFIG_DEFAULT_REPORTER for static library with main()

mmatrosov:
## Description
Currently if a client program uses Catch2-provided `main()` function from the static library `libCatch2WithMain.a`, there is no way to set the default reporter with the `CATCH_CONFIG_DEFAULT_REPORTER` macro, since [the customizable code](https://github.com/catchorg/Catch2/blob/216713a4066b79d9803d374f261ccb30c0fb451f/include/internal/catch_config.hpp#L66) has already been compiled. I added a test to demonstrate this:

1. Checkout the first commit that adds the test: `git checkout 23c486ec`
2. Build and run the test: `mkdir -p build && cmake . -B build -DCATCH_BUILD_EXAMPLES=ON -DCATCH_CONFIG_DEFAULT_REPORTER="\"xml\"" -DCATCH_BUILD_STATIC_LIBRARY=YES && make -C build -j && build/examples/MainFromStaticLib`

See how the default reporter is used, even though `xml` was specified. I forward this option in cmake as a define. See:

3. Checkout the head of the branch: `git checkout default-reporter-static-lib`
4.  Repeat step 2

See how the specified `xml` reporter is now properly used.

This PR is for `v2.x` branch, because I don't understand the state of v3. There are no stable releases for it, the last releaes was more than a year ago, and it is not available in conan-center. But I want to use this feature in the nearest future, that's why I am targeting v2.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -136,6 +136,11 @@ add_executable( 207-Rpt-TeamCityReporter ${EXAMPLES_DIR}/207-Rpt-TeamCityReporte
 #    add_executable( ${name} ${EXAMPLES_DIR}/${name}.cpp $<TARGET_OBJECTS:CatchMain> ${HEADER_DIR}/catch2/catch.hpp )
 #endforeach()
 
+if (CATCH_BUILD_STATIC_LIBRARY)
+    add_executable( MainFromStaticLib ${EXAMPLES_DIR}/100-Fix-Section.cpp )
+    target_link_libraries( MainFromStaticLib Catch2WithMain )
+endif()
+
 foreach( name ${TARGETS_ALL} )
     target_include_directories( ${name} PRIVATE ${HEADER_DIR} ${CATCH_DIR} )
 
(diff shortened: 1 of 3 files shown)


examples/CMakeLists.txt
```diff
@@ -136,6 +136,11 @@ add_executable( 207-Rpt-TeamCityReporter ${EXAMPLES_DIR}/207-Rpt-TeamCityReporte
 #    add_executable( ${name} ${EXAMPLES_DIR}/${name}.cpp $<TARGET_OBJECTS:CatchMain> ${HEADER_DIR}/catch2/catch.hpp )
 #endforeach()
 
+if (CATCH_BUILD_STATIC_LIBRARY)
+    add_executable( MainFromStaticLib ${EXAMPLES_DIR}/100-Fix-Section.cpp )
+    target_link_libraries( MainFromStaticLib Catch2WithMain )
+endif()
```
horenmar: Why have these been added? What do they test?

mmatrosov: This test creates an executable that uses `main()` from the static lib. It allows to see whether the default reporter option is respected. Please see the description of the PR for the details.

The test is not automated though - one has to manually assess its output to see if the right reporter was used. As I did.

horenmar: > The test is not automated though - one has to manually assess its output to see if the right reporter was used.

Right, that's the problem. If the test is not run during CI, then it effectively does not exist in the long term.

This change is kinda annoying to test, because it is a CMake config option, but not impossible. What would need to happen is to write a (Python, for cross-platform compatibility) script that configures a separate CMake build, builds it and checks the output afterwards -- this script can then be registered with CTest as a normal test.

Or, given how simple this change is, just leave it untested -- especially since it won't directly translate to the devel branch, which needs to handle this a bit differently.

mmatrosov: Ok, I see your point. I removed the test.

Issue #2348, opened 2022-01-06
`TEMPLATE_PRODUCT_TEST_CASE` trips up `-Wsign-conversion`


cjdb:
**Describe the bug**
`TEMPLATE_PRODUCT_TEST_CASE`'s internals do something to trip up `-Wsign-conversion`, which causes problems for user code.

**Expected behavior**
No warning should be visible from user code.

**Reproduction steps**
```cpp
// g++-11 -Werror=sign-conversion
#include <catch.hpp>
#include <vector>

TEMPLATE_PRODUCT_TEST_CASE("hello", "", (std::vector), (int)) {}
```
[Compiler Explorer](https://godbolt.org/z/6ra1EoTen)

**Platform information:**

 - OS: Ubuntu 20.04
 - Compiler+version: GCC 11
 - Catch version: v2.13.8 locally; trunk on Compiler Explorer

horenmar:
Okay, I see this is open against v2 branch. AFAIK [this commit should fix it](https://github.com/catchorg/Catch2/commit/8cb8f0b08b0ec35edd346df999f3b854ad05aa41) but it won't apply cleanly.

PR #2349, opened 2022-01-06
suppresses sign conversion warning in `TEMPLATE_PRODUCT_TEST_CASE`

cjdb:
`-Wsign-conversion` was complaining about an implicit sign conversion,
so this commit explicitly converts the unsigned type to a signed one.

Commit also adds a `-Wsign-conversion` to the list of warnings, and
reorders that list so it's easier to spot what's missing (and removed
two duplicate warnings).

Bug: #2348

### DIFF
--- a/.github/workflows/linux-simple-builds.yml
+++ b/.github/workflows/linux-simple-builds.yml
@@ -65,6 +65,22 @@ jobs:
             build_type: Release
             std: 17
             other_pkgs: clang-10
+          - cxx: clang++-10
+            build_type: Debug
+            std: 20
+            other_pkgs: clang-10
+          - cxx: clang++-10
+            build_type: Release
+            std: 20
+            other_pkgs: clang-10
+          - cxx: g++-10
+            build_type: Debug
+            std: 20
+            other_pkgs: g++-10
+          - cxx: g++-10
+            build_type: Release
+            std: 20
+            other_pkgs: g++-10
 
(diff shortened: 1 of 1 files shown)


CMake/MiscFunctions.cmake
```diff
@@ -35,35 +35,34 @@ function(add_warnings_to_targets targets)
 
     if (NOT MSVC)
         set(CHECKED_WARNING_FLAGS
+          "-Wabsolute-value"
```
cjdb: To save time auditing, here's what I did:

1. Add `"-Wsign-compare"` to the end of the list.
2. Sorted the list using VS Code.
3. Removed any duplicates I noticed.

cjdb:
Since this is a lot of work, I'm going to hold off on fixing this up any further till I get direction from a maintainer.

horenmar:
Yeah, this is unmergeable like this.

If you want to pursue these changes, it will need to be merged in small pieces/commits, that are reviewable. e.g. I have no idea what actually changed in the template test registries (if anything) other than formatting.

A good way to start would be to cleanup the warnings + add sign conversions as `no-error` warning, so `-Werror` doesn't apply to it, then start merging fixups for reasonably small chunks. Also I prefer cleaning up type mixings when possible, rather than just adding casts everywhere. This is usually bit more work in the short term, but in the long term is easier to maintain.

PR #2353, opened 2022-01-14
Ignoring VSCode settings dir

rsevero:
## Description
Changing .gitignore file to ignore the default VSCode settings dir.

This is importat so developers and testers can use VSCode to 
work on Catch without having toi worry about accidently 
commiting unrelevant VSCode local settings.


horenmar:
Sure, why not.

PR #2361, opened 2022-02-01
When building a static library, export include/ instead of single_include/

patsytau:
## Description
This pull request makes Catch2WithMain link privately to Catch2 and export the `include/` directory, rather than the `single_include/` directory when building with CATCH_BUILD_STATIC_LIBRARY.

Previously, the amalgamated header, `single_include/catch.hpp` would be included, which is the header file for the header-only version of the library, so there would be no benefit to building the static library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,12 @@ add_library(Catch2::Catch2 ALIAS Catch2)
 # Hacky support for compiling the impl into a static lib
 if (CATCH_BUILD_STATIC_LIBRARY)
   add_library(Catch2WithMain ${CMAKE_CURRENT_LIST_DIR}/src/catch_with_main.cpp)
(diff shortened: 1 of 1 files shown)


horenmar:
No.
The static lib is built from the single header, and it should also target includes to the single header. This is actually what would happen if I merged your PR anyway, because you reuse `$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>` for the include path on installation, but that is the path for the single header -- the separate headers are never installed.

The advantage of the static library is that is built once for multiple consuming projects, not the different include files... in fact, including `single_include/catch2/catch.hpp` might be faster (especially on spinning rust) than including `include/catch.hpp`, because in one case the preprocessor has to deal with opening only one file and then ignoring ~12k inactive lines behind an ifdef, while in the other the processor doesn't have to deal with those lines, but it is forced to find and open 50+ files.

PR #2362, opened 2022-02-02
feature: double -> const auto when static_cast in right

Certseeds:
## Description

double -> const auto when static_cast<double> in right

~## GitHub Issues~

Signed-off-by: Certseeds <51754303+Certseeds@users.noreply.github.com>


horenmar:
Sure, why not.

PR #2364, opened 2022-02-03
fix: a single _ should not be used for a name

### LINKED ISSUE #2363: Catch cannot be used if `_` is a macro / defined (2.13.5-2.13.8)
A single underscore was used as a name here:

https://github.com/catchorg/Catch2/blob/958944d27a2d2fb82aa008377bf4f8752f6b848e/include/internal/catch_run_context.cpp#L455

This breaks tests that are simulating the common practice of `#define`ing a `_` macro that handles translation. Broken in https://github.com/catchorg/Catch2/commit/8f277a54c0b9c1d1024dedcb2dec1d206971e745, breaks pybind11 in https://github.com/pybind/pybind11/pull/3679.

The test in pybind11 is adding a `-D_=1` to ensure that pybind11 doesn't break when `_` is defined to something, this might be a good idea for a test for Catch's code, too.

~~I can't tell where that comment came in, it was after (working) 2.13.3 and is in (broken) 2.13.8. By date, maybe v2.13.6~~. I wish I could use git bisect on tags! :) Edit: first broken version is 2.13.5.

henryiii:
## Description

This avoids using a single underscore for a name, which can conflict with defining `_` for translation. One idea for a test would be to add `-D_=1` when building Catch, to help ensure that this is not rebroken in the future. Another could be to do a static check for a single underscore, though that is allowed in comments and such. Something like this:

```console
$ git grep '\<_\>' src/*
src/catch2/internal/catch_run_context.cpp:        FatalConditionHandlerGuard _(&m_fatalConditionhandler);
```

(that's on `devel`)

## GitHub Issues

Closes #2363.

### DIFF
--- a/include/internal/catch_run_context.cpp
+++ b/include/internal/catch_run_context.cpp
@@ -452,7 +452,7 @@ namespace Catch {
     }
 
     void RunContext::invokeActiveTestCase() {
-        FatalConditionHandlerGuard _(&m_fatalConditionhandler);
+        FatalConditionHandlerGuard _gaurd(&m_fatalConditionhandler);
         m_activeTestCase->invoke();
     }
 


henryiii:
PS: I think this happened on the 3.x development branch too.

PR #2365, opened 2022-02-05
build: proper support for lib suffix in conan

gegles:
## Description

Currently, the conan configuration does not allow for the Debug libraries (suffixed by a `d`) to be found by the using project. This PR simply solves this issue by matching the conan configuration with the CMake config (using `CMAKE_DEBUG_POSTFIX`).


horenmar:
Thanks

PR #2368, opened 2022-02-18
Examples: remove references to catch_default_main

johnbeard:
catch_default_main.hpp was removed in db32550898

The example 000-CatchMain.cpp is no longer compiled, but is still
present in the examples and is still references by other example
files. Remove the file and references to it, as they are confusing.


horenmar:
Thanks.

PR #2133, opened 2020-12-27
Added ElementsAre and UnorderedElementsAre

urielgarciarivas:
## Description
Added `ElementsAre` and `UnorderedElementsAre` generic range matchers which take a range during construction and their `match` method take a range as well.

## GitHub Issues
Addressing issue #2087 .

## Notes
I did not include support to providing a `matcher` as construction argument due to `ElementsAre` would be exactly the same as `AllMatch`.

Issue #2087, opened 2020-11-02
Add new range Matchers using the new generic matcher support


horenmar:
**Description**
v3 contain facilities for type-generic matchers, so that a matcher can be used on an arbitrary input type, as long as it fulfills the correct concept.

Catch2 already [provides some generic matchers](https://github.com/catchorg/Catch2/blob/devel/docs/matchers.md#generic-range-matchers), but it should provide more of them. The most obvious missing ones are

* `ElementsAre`, which would perform check that the input arg is exactly equal to the elements
* `UnorderedElementsAre`, which would perform check that the input arg is equal to the elements, in some permutation.

**Additional context**
Both of the matchers should allow providing either an exact value, or a matcher, as any of the construction arguments.

urielgarciarivas:
I will be working on this issue! I see that there is a similar implementation in `catch_matchers_vector.hpp`.

PR #2377, opened 2022-03-02
Added ElementsAre and UnorderedElementsAre 

johnbeard:
[This is a continuations of the solid start made in #2133 by @Garz4 in 2020, with some modifications]

Added `ElementsAre` and `UnorderedElementsAre` generic range matchers which take a range during construction and their match method take a range as well.

Addresses issue #2087 

As before, no matcher constructor analogous to `ContainsMatcherMatcher` is provided, as this is just the same as `AllMatch`.

Differences to the previous PR:

* Add unit tests and docs
* Differing types are permitted, so you can compare, e.g. a vector of ints to an array of chars, or any other comparable types.
* STL-like predicates are also permitted instead of the default `std::equal_to`


tests/SelfTest/UsageTests/MatchersRanges.tests.cpp
```diff
@@ -548,3 +554,210 @@ TEST_CASE( "The quantifier range matchers support types with different types ret
...
+        }
+        SECTION( "Two equal containers, one random access, one not" ) {
+            const std::array<int, 3> array_int_a{ { 1, 2, 3 } };
+            const std::list<int> list_char_a{ 1, 2, 3 };
+
+#ifdef CONCEPT_TESTING
+            // Verify these types really are different in random access nature
+            static_assert( std::random_access_iterator<decltype(array_int_a)::iterator> );
+            static_assert( !std::random_access_iterator<decltype(list_char_a)::iterator> );
+#endif
```
horenmar: Is there a reason not to have this in C++14 through `std::is_base_of` and `std::iterator_traits`?

tests/SelfTest/UsageTests/MatchersRanges.tests.cpp
```diff
@@ -548,3 +554,210 @@ TEST_CASE( "The quantifier range matchers support types with different types ret
...
+
+        SECTION( "Two equal non-empty containers (close enough)" ) {
+            const std::vector<int> vector_a{ { 1, 2, 3 } };
+            const std::array<char, 3> array_a_plus_1{ { 2, 3, 4 } };
+            CHECK_THAT( vector_a,
+                        ElementsAre( array_a_plus_1, close_enough ) );
+            CHECK_THAT( vector_a,
+                        UnorderedElementsAre( array_a_plus_1, close_enough ) );
+        }
+    }
```
horenmar, line 1056: This should also have (and support) rangeish types that don't have member begin (e.g. they provide ADL-findable begin only), and have different types for begin and end (the support for these is currently spotty, but the goal is to provide it eventually).

Honestly the test utils for this could use some refactoring and cleanup in general, so I don't blame you if you don't want to write the tests, but the impl should reasonably support these.

src/catch2/matchers/catch_matchers_elements_are.hpp
```diff
@@ -8,61 +8,124 @@
...
-    }
-
-    //! Creates a matcher that checks if all elements in a range are equal to all elements in another range in some permutation
-    template<typename RangeLike>
-    UnorderedElementsAreElementsMatcher<RangeLike> UnorderedElementsAre(RangeLike rng) {
-        return { std::forward<RangeLike>(rng) };
-    }
-
-} // namespace Matchers
+    namespace Matchers {
```
horenmar: huh, is this actually the formatting given by clang-format?

I would swear the clang-format had nested namespace indents disabled.

src/catch2/matchers/catch_matchers_elements_are.hpp
```diff
@@ -8,61 +8,124 @@
...
+            UnorderedElementsAreElementsMatcher<RangeLike, std::equal_to<>>>
+        UnorderedElementsAre( RangeLike range ) {
+            return { CATCH_FORWARD( range ), std::equal_to<>{} };
+        }
+
+        /**
+         * Creates a matcher that checks if all elements in a range are equal
+         * to all elements in another range, in some permuation.
+         *
+         * Uses to provided predicate `eq` to do the comparisons
```
horenmar: nitpick: `eq` is not the name of the arg in this function

horenmar:
I didn't check the code in too much detail, but here is a big note on this:

> As before, no matcher constructor analogous to ContainsMatcherMatcher is provided, as this is just the same as AllMatch.

I want these to support matcher inputs on the expected side, so something like this can be written: `REQUIRE_THAT( foo(), ElementsAre(IsOdd(), IsEven(), IsOdd());`. Ideally I would also like something like this to work: `REQUIRE_THAT( foo(), ElementsAre(1, 2, 3));` but that is secondary and I am fine with forcing a different syntax there.

Alternatively, I am willing to hear an argument on why the functionality above should have a different name (or, the functionality in this PR could be renamed to something like `Equals` and `UnorderedEquals`).

johnbeard:
Thank you for the careful review!

> I want these to support matcher inputs on the expected side, so something like this can be written: `REQUIRE_THAT( foo(), ElementsAre(IsOdd(), IsEven(), IsOdd());`.

OK, so just to check, the difference is then that `AllMatch` applies a single matcher to each element:

> `REQUIRE_THAT( foo(), AllMatch( IsOdd() );

And `ElementsAre` uses a variadic form where the matchers are applied element-wise:

> REQUIRE_THAT( foo(), ElementsAre(IsOdd(), IsEven(), IsOdd());

Makes sense to me, though I'll have to have a think about how to do it!

If I can indeed do such a variadic matcher syntax, then it feel like `REQUIRE_THAT( foo(), ElementsAre(1, 2, 3));` would be just one SFINAE away at that point.

src/catch2/matchers/catch_matchers_elements_are.hpp
johnbeard: Already fixed in followup: 7c09700

src/catch2/matchers/catch_matchers_elements_are.hpp
johnbeard: It seems not: `NamespaceIndentation: All` is specified (should be `None` for zero indent) :https://github.com/catchorg/Catch2/blob/61d0f7a9afb88c12f8931f80321e58f2cfbc62db/.clang-format#L16

tests/SelfTest/UsageTests/MatchersRanges.tests.cpp
johnbeard: Because I was in C++20 mode and didn't think of it? 😅 See 9080853 for a re-implementation.

horenmar: Yeah, we still do things the hard way here, C++14 only 😃

horenmar:
> Thank you for the careful review!

Oh don't worry, that will come later. 🙃  And you are correct about the differences between `AllMatch` (`Any`, `None`) matcher and this one.

I do think you might be underestimating the complexity to support both direct-element comparison, and matcher-based comparison. For the latter, most work can be reused from the support code in `&&` and `||` operators, but when I did a quick prototype, the work to intermix them proved to be pretty annoying. Do note that mixing both in single matcher should be also supported, so that e.g. `ElementsAre(1, IsOdd(), 3)` is valid.

johnbeard:
> I do think you might be underestimating the complexity to support both direct-element comparison, and matcher-based comparison.

Probably! But worth a try!

I don't mind to split to the work to `Equals/UnorderedEquals` and something else, or `ElementsAre` and something else. If you'd like to go that way, I have no objection. But it seems like if the constructors look like:

* ElementsAre( other_container )
* ElementsAre( 1, 2, 3 ) // module a set of {}, see below

Then having separate names for each would be a shame and a bit confusing since they're so similar.

Would it make sense to use an initializer-list constructor like std::vector: `ElementsAre{ 1, IsOdd(), 3 }` rather than a variadic constructor? That might make it possible to also do `ElementsAre( {1,2,3}, predicate )`, since the init list is just the first argument. In the same way that the init-list ctor of std::vector has an optional allocator argument. If it's just variadic, it'll certainly be (at least) awkward to separate the last one if it's a predicate.

johnbeard:
I'm still trying to think of how to do this, but it probably will take me a while to figure it out (I also will never complain if someone wants to Just Do It and preempt it, I claim no dibs!).

Maybe it makes more sense to tidy and move to merge what exists here? This interface remains the same?:

```
REQUIRE_THAT( range, ElementsAre(otherRange);
```

Issue #2316, opened 2021-10-30
CATCH_CONFIG_DISABLE fails CAPTURE(i,j)


laoshaw:
**Describe the bug**
when I turn on CATCH_CONFIG_DISABLE, all CAPTURE(i,j) , i.e. more than one arg will err.

**Expected behavior**
No such error

**Reproduction steps**
Build below code with -DCATCH_CONFIG_DISABLE 

```
#include "catch.hpp"

TEST_CASE("generators") {
    auto i = GENERATE(as<std::string>(), "a", "b", "c");
    SECTION("one") {
        auto j = GENERATE(range(8, 11), 2);
        CAPTURE(i, j);
        SUCCEED();
    }
    SECTION("two") {
        auto j = GENERATE(3, 141, 1.379);
        CAPTURE(i, j);
        SUCCEED();
    }
}

// cross-product of two ranges.
TEST_CASE("100x100 ints", "[.][approvals]") {
    auto x = GENERATE(range(0, 100));
    auto y = GENERATE(range(200, 300));
    CHECK(x < y);
}
```
will lead to below errs:

```
generate.cpp:7:21: error: macro "CAPTURE" passed 2 arguments, but takes just 1
    7 |         CAPTURE(i, j);
      |                     ^
In file included from generate.cpp:1:
catch.hpp:17887: note: macro "CAPTURE" defined here
17887 | #define CAPTURE( msg ) (void)(0)
      | 
generate.cpp:12:21: error: macro "CAPTURE" passed 2 arguments, but takes just 1
   12 |         CAPTURE(i, j);
      |                     ^
In file included from generate.cpp:1:
catch.hpp:17887: note: macro "CAPTURE" defined here
17887 | #define CAPTURE( msg ) (void)(0)
      | 
generate.cpp: In function ‘void ____C_A_T_C_H____T_E_S_T____0()’:
generate.cpp:7:9: error: ‘CAPTURE’ was not declared in this scope
    7 |         CAPTURE(i, j);
      |         ^~~~~~~
generate.cpp:12:9: error: ‘CAPTURE’ was not declared in this scope
   12 |         CAPTURE(i, j);
```

**Platform information:**

 - OS: **Linux**
 - Compiler+version: **G++ v9.3**
 - Catch version: **v2.13.7**, the newest

**Additional context**

My fix is below to get it working:

```
17886c17886,17887
< #define CAPTURE( msg ) (void)(0)
---
> #define CAPTURE( ... ) (void)(0)
```

PR #2378, opened 2022-03-02
CAPTURE not variadic when disabled

johnbeard:
This caused errors on CAPTURE(a,b) when CATCH_CONFIG_DISABLE
is set.

Closes: #2316

### DIFF
--- a/include/catch.hpp
+++ b/include/catch.hpp
@@ -433,7 +433,7 @@ using Catch::Detail::Approx;
 #define INFO( msg ) (void)(0)
 #define UNSCOPED_INFO( msg ) (void)(0)
 #define WARN( msg ) (void)(0)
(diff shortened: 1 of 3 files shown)


horenmar:
Thanks, but you should also add a test to `projects/ExtraTests/X02-DisabledMacros.cpp`.

johnbeard:
Re the X01 test: Is that right? The test fails due to that `CATCH_FAIL_CHECK` but I'm not sure if that's deliberate somehow? Or maybe the test should be split with that bit in a `[!shouldfail]`?

horenmar:
I don't see a failing CI build so I am not sure what I mean, but the changes look correct. If nothing comes up, I'll merge this tomorrow.

PR #2380, opened 2022-03-03
Fix name of with-main library in .pc file

### LINKED ISSUE #2379: catch2-with-main.pc disagrees with installed lib name
**Describe the bug**
`catch2-with-main.pc` has the following line in it:
```
Libs: -L${libdir} -lCatch2WithMain
```
So, the expected basename of the library file will be `libCatch2WithMain`.

The installed library is named `libCatch2Main.a`. Thus, linking fails when using pkgconfig.

**Expected behavior**
Linking should succeed, if the correct lib name is given.

**Reproduction steps**
```
$ pkg-config --libs catch2-with-main
-L/home/juckelman/projects/make_world/install/lib64 -lCatch2WithMain -lCatch
$ ls /home/juckelman/projects/make_world/install/lib64
./  ../  cmake/  libCatch2.a libCatch2Main.a
```

uckelman-sf:
## Description

`catch2-with-main.pc` has an incorrect library basename in it, which prevents linking when the .pc file is used. This PR fixes the library name in the .pc file.

## GitHub Issues

This PR fixes #2379.

### DIFF
--- a/CMake/catch2-with-main.pc.in
+++ b/CMake/catch2-with-main.pc.in
@@ -7,4 +7,4 @@ Description: A modern, C++-native test framework for C++14 and above (links in d
 Version: ${pkg_version}
 Requires: catch2 = ${pkg_version}
 Cflags: -I${includedir}
-Libs: -L${libdir} -lCatch2WithMain
+Libs: -L${libdir} -lCatch2Main


horenmar:
Yeah, I think this shows how much I use the pkg-config parts - not at all.

Thanks.

uckelman:
> Yeah, I think this shows how much I use the pkg-config parts - not at all.

This will be getting daily use in a project of mine, so I appreciate having it working now. Thanks for merging it!

PR #2383, opened 2022-03-05
Convert tags to ctest labels

### LINKED ISSUE #1590: Make catch tags work with ctest labels

conr2d:
## Description

This PR allows converting Catch2 tags to CTest labels.

## GitHub Issues

Closes #1590


horenmar:
Okay, so in general I am not against, but I have multiple issues with the specific implementation.

1) Test name does not uniquely identify a `TEST_CASE`. This means that executing `./tests {test-name} --list-tags` can provide tags from multiple different test cases.
2) Executing the test binary for each separate test is pretty large performance impact.

Both of these can be fixed by taking a harder approach towards registering tests, and extracting both test name and its tags at the same time.

In an ideal world, CMake would have functional xml support, and we could just retrieve the information easily through the xml reporter. We are not in an ideal world, so the next best thing is to run `--list-tests`, and parse the output. The format can be parsed by using leading indentation as guide, e.g. 

```
{2 spaces}Test Name
{4 spaces}(Optional) Test location
{6 spaces}(Optional) Tags
```


---

I am also told that the way script is extended with tags-as-properties will scale badly, but I don't know enough CMake to confirm or deny that.

horenmar:
The performance of this approach is such that I won't merge it. More notes here: https://github.com/catchorg/Catch2/pull/2696#issuecomment-1593437575

PR #2384, opened 2022-03-05
Add variadic CAPTURE tests

johnbeard:
## Description

This makes sure that CAPTURE works when called with variadic arguments,
and also works that way when disabled.

This is a port of 5e94498ed04820b1022f0bead0f7d4d5fb7a06dd to the
devel branch.

## GitHub Issues

Tests fix to #2316


horenmar:
Sure, thanks.

PR #2319, opened 2021-11-14
Add AllTrue, AnyTrue, NoneTrue generic range matchers

schallerr:
## Description
This PR adds `AllTrue`, `AnyTrue`, `NoneTrue` generic range matchers, which match a range if all, any, or none of the elements in a range are `true`, respectively.

## GitHub Issues
Addresses issue #2109.

PR #2386, opened 2022-03-06
Refactor MatchersRanges test utils

schallerr:
## Description
Refactor test util classes in MatchersRanges.tests. This is preparatory work for #2319.

- do not hard-code element type and content of mock containers
- prefix members with `m_`
- add const and non-const iterators to `with_mocked_iterator_access`

### DIFF
--- a/tests/SelfTest/UsageTests/MatchersRanges.tests.cpp
+++ b/tests/SelfTest/UsageTests/MatchersRanges.tests.cpp
@@ -14,31 +14,25 @@
 #include <catch2/matchers/catch_matchers_predicate.hpp>
 #include <catch2/matchers/catch_matchers_string.hpp>
 
-#include <array>
 #include <cmath>
+#include <initializer_list>
 #include <list>
 #include <map>
 #include <type_traits>
 #include <vector>
+#include <memory>
 
 namespace {
 
-template <typename Arg, typename... Args>
-auto make_array( Arg&& arg, Args&&... args ) {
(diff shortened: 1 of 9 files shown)


horenmar:
fwiw you can also just store the elements in a vector and use init-list constructor to init the types

tests/SelfTest/UsageTests/MatchersRanges.tests.cpp
```diff
@@ -101,80 +111,96 @@ class has_different_begin_end_types {
...
+template <typename T> struct with_mocked_iterator_access {
+    std::vector<T> m_elements;
 
-struct with_mocked_iterator_access {
-    static constexpr size_t data_size = 5;
-    std::array<int, data_size> elements{ {1, 2, 3, 4, 5} };
-    std::array<bool, data_size> touched{};
-    std::array<bool, data_size> derefed{};
+    // use plain arrays to have nicer printouts with CHECK(...)
+    mutable std::unique_ptr<bool[]> m_touched;
```
horenmar: I noticed that this member is filled out wrong -- `m_touched[0]` will always be false. This was in the code before, but we should either fix it, or remove the member completely... I am not sure if it makes sense to keep it or not, so I will leave it up to you.

schallerr: good catch, I removed it, it was anyway never checked in the tests

PR #2387, opened 2022-03-11
Fix broken Bazel build

Vertexwahn:
Since the CMake header file magic `catch_user_config.hpp.in` was introduced the Bazel work did not work anymore. This PR fixes the Bazel build.

There is also an attempt to get `expand_template` into bazel-skylib:
https://github.com/bazelbuild/bazel-skylib/pull/330

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,14 +1,69 @@
 # Load the cc_library rule.
 load("@rules_cc//cc:defs.bzl", "cc_library")
 
+load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
+
+expand_template(
+    name = "catch_user_config.hpp",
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks. 

Can't say I like having yet another place where the configuration options have to be accounted for manually. Is there no better way to do this in Bazel?

Vertexwahn:
I am not aware of a better way. Other projects introduce a Bazel rule that tries to reproduce CMake behavior (https://github.com/RobotLocomotion/drake/blob/master/tools/workspace/cmake_configure_file.bzl) - but in the end, this does not avoid duplication of parameters.

Yes it is some manual effort to keep it running - I am willing to maintain the Bazel build of Catch 2 for the next six month.

horenmar:
Ok thanks.

PR #2390, opened 2022-03-13
Add eventpp and accessorpp to Catch2 users

wqking:
Add two open source libraries as Catch2 user, eventpp and accessorpp


horenmar:
Thanks

Issue #2328, opened 2021-12-08
Missing # in the filename tag when __FILE__ is a filename


ivan236634452:
**Describe the bug**
When compiler flags are such that \_\_FILE\_\_ macro is replaced with a relative file path (on MSVC no /Z7, /Zi or /ZI and no /FC) then it might be replaced with just a filename with no path separators. In this case tags for tests with --filenames-as-tags do not contain '#' symbol.

The bug is present in the 2.x branch, but not in the main branch where the relevant code has been redesigned.
The culprit is the following code in catch_session.cpp:

```
void applyFilenamesAsTags(Catch::IConfig const& config) {
    auto& tests = const_cast<std::vector<TestCase>&>(getAllTestCasesSorted(config));
    for (auto& testCase : tests) {
        auto tags = testCase.tags;

        std::string filename = testCase.lineInfo.file;
        auto lastSlash = filename.find_last_of("\\/");
        if (lastSlash != std::string::npos) {
            filename.erase(0, lastSlash);
            filename[0] = '#';
        }

        auto lastDot = filename.find_last_of('.');
        if (lastDot != std::string::npos) {
            filename.erase(lastDot);
        }

        tags.push_back(std::move(filename));
        setTags(testCase, tags);
    }
}
```

Adding the following nonintrusive but uninspired adjustment resolves the issue:

```
void applyFilenamesAsTags(Catch::IConfig const& config) {
    auto& tests = const_cast<std::vector<TestCase>&>(getAllTestCasesSorted(config));
    for (auto& testCase : tests) {
        auto tags = testCase.tags;

        std::string filename = testCase.lineInfo.file;
        auto lastSlash = filename.find_last_of("\\/");
        if (lastSlash != std::string::npos) {
            filename.erase(0, lastSlash);
            filename[0] = '#';
        }
        else
        {
            filename.insert(0, "#");
        }

        auto lastDot = filename.find_last_of('.');
        if (lastDot != std::string::npos) {
            filename.erase(lastDot);
        }

        tags.push_back(std::move(filename));
        setTags(testCase, tags);
    }
}
```

**Expected behavior**
Symbol # always prefixes filename tags.

**Reproduction steps**
Compiler flags (MSVC): no /Z7, /Zi, /ZI, /FC
Catch flags: [#test_file] --filenames-as-tags
Place test_file.cpp with any test cases in the root project directory and try to run them.
In the current 2.x branch no tests would be found.

PR #2393, opened 2022-03-17
Fix filename as tag bug

boing102:
## Description
When a test file isn't provided with "/" in the path, the filename isn't registered as a tag.
Fixing by providing a condition for this edge case.

This is the first time I'm contributing to this project (trying to learn c++). I wasn't sure where to add a test for this change, so would appreciated if someone could point to place.

## GitHub Issues
This PR fixes https://github.com/catchorg/Catch2/issues/2328

### DIFF
--- a/include/internal/catch_session.cpp
+++ b/include/internal/catch_session.cpp
@@ -127,6 +127,10 @@ namespace Catch {
                     filename.erase(0, lastSlash);
                     filename[0] = '#';
                 }
+                else
+                {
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks. As the code currently is, it is pretty much impossible to write a test -- the function that trims out the filename would have to be split into its own header, and then the tests could be written.

Given that this change is only relevant to v2 branch, I don't think this is worth the effort.

PR #2405, opened 2022-04-01
Get Bazel build working again

Vertexwahn:
This PR enables the Bazel build again.

Including generated headers using `<...>` instead of `"..."` does unfortunately not work using Bazel.

I am not  100% sure but I think including via `"..."` should be preferred in cpp files since this guarantees that not a system installed catch header file is included. I would suggest to change everything to `"..."`

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -4,7 +4,7 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
 load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
 
 expand_template(
-    name = "catch_user_config.hpp",
+    name = "catch_user_config_hpp",
     out = "catch2/catch_user_config.hpp",
     substitutions = {
         "#cmakedefine CATCH_CONFIG_ANDROID_LOGWRITE": "",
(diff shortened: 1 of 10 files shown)


horenmar:
> I am not 100% sure but I think including via "..." should be preferred in cpp files since this guarantees that not a system installed catch header file is included. I would suggest to change everything to "..."

No.

The difference in the two include styles is that for `#include "foo"` the compiler first attempts to find `foo` in some implementation specific manner, and if that fails, it rewrites it into `#include <foo>` and then attempts to resolve that (in some implementation specific manner). [This can lead to some fun](https://godbolt.org/z/6MrPnnhss).

This also means that `#include "catch2/catch_user_config.hpp"` is strictly looser than `#include <catch2/catch_user_config.hpp>`, and if it works, then the issue likely is that the `catch_user_config.hpp` file is created in a relative directory, because in practice, the difference for GCC is that `""` include paths first check relative to the file being compiled, while `<>` include paths check compared to `-Idir` paths and built-in system paths.

The other option is that Bazel provides `-iquoteDir`, in which case it really shouldn't. (`-iquote` appends path to search locations for `""` style includes, but not to `<>` style includes)

Vertexwahn:
@horenmar I have seen some libs that prefer "quote" includes over <system> includes ([Eigen](https://eigen.tuxfamily.org/), [GoogleTest]( https://github.com/google/googletest), [{fmt}](https://github.com/fmtlib/fmt) which makes things easier for Bazel - but I think changing the code because of Bazel is not a good option.

 I try to figure out if there is a way that works with keeping `#include <catch2/catch_user_config.hpp>`.

I initially tested this on Windows 10/11 and Visual Studio 2022 and it works there with `#include <catch2/catch_user_config.hpp>`.
But it does not work on Ubuntu 20.04 using GCC 9.

Vertexwahn:
https://github.com/catchorg/Catch2/pull/2399 will fix all issues

Issue #2392, opened 2022-03-16
Command line option to skip benchmarks


nandanvasudevan:
**Description**
Enable or disable benchmarks with a command line option

**Additional context**
I don't always want to run benchmarks. In the current situation using any `ifdef`s or `#define`s, triggers a complete re-compilation which is really time-consuming and something I really want to avoid.
It would be really helpful if there was a command line option to disable benchmarks.

## Possibilities with existing
Lets look at what the current options to achieve this are.
### Using Tags
Right now, one plausible way is to keep benchmarks in a separate `TEST_CASE` with a tag that reads 'Benchmark' and then disallow running of that tag which [is supported](https://github.com/catchorg/Catch2/blob/38879296a7a0a1787a85d470047fd14d8f6d09aa/docs/command-line.md?plain=1#L110). 
However, this will almost always lead to code duplication since the setup code will need to be repeated just for the benchmark.
### Using Sections
By keeping a benchmark as a section in `TEST_CASE` and then running only these sections seems possible.
However, there is option to specify, run all except this section.
This wouldn't work.

PR #2408, opened 2022-04-11
Added --skip-benchmarks command-line option.

danifeist:
This PR adds the feature requested by https://github.com/catchorg/Catch2/issues/2392 "Command line option to skip benchmarks" by introducing the command-line option "--skip-benchmarks" so that Catch runs no benchmarks when this option is set. I'd be glad about feedback.


docs/command-line.md
```diff
@@ -370,6 +372,12 @@ See [The LibIdentify repo for more information and examples](https://github.com/
 Will cause the executable to print a message and wait until the return/ enter key is pressed before continuing -
 either before running any tests, after running all tests - or both, depending on the argument.
 
+<a id="skip-benchmarks"></a>
+## Skip all benchmarks
+<pre>--skip-benchmarks</pre>
+
```
horenmar, line 378: This should have a version placeholder (see contributing.md for how it looks).

docs/command-line.md
```diff
@@ -370,6 +372,12 @@ See [The LibIdentify repo for more information and examples](https://github.com/
...
+When this option is set, all benchmarks of each and every Catch test case will be skipped so that it appears as if there were no benchmark defined.
```
horenmar: How about

```suggestion
This flag tells Catch2 to skip running all benchmarks. Benchmarks in this
case mean code blocks in `BENCHMARK` and `BENCHMARK_ADVANCED` macros, not
test cases with the `[!benchmark]` tag.
```

I think this is clearer. (Also new docs use Catch2 and line length of 80, to play nicer with diffs)

src/catch2/benchmark/catch_benchmark.hpp
```diff
@@ -96,7 +96,10 @@ namespace Catch {
             template <typename Fun, std::enable_if_t<!Detail::is_related<Fun, Benchmark>::value, int> = 0>
                 Benchmark & operator=(Fun func) {
                 fun = Detail::BenchmarkFunction(func);
-                run();
+                auto const* cfg = getCurrentContext().getConfig();
+                if (!cfg->skipBenchmarks()) {
+                    run();
+                }
```
horenmar, line 102: I think this should happen before we create the function wrapper.

horenmar (changes requested):
This also needs tests.

It should be enough to pick some TEST_CASE that's already in the SelfTest project and has benchmarks, and then add new CTest test that invokes `SelfTest` with `--skip-benchmarks` and test spec for that one test. The test then checks that the benchmark was skipped by looking for the benchmark output.

Or add a new TEST_CASE just for this, that way is more resilient to later changes.

src/catch2/benchmark/catch_benchmark.hpp
danifeist: you'd place `fun = Detail::BenchmarkFunction(func);` inside the if-clause, did i got that right?

horenmar: Yeah. There is no reason to make the allocation if it won't be used afterwards.

tests/CMakeLists.txt
```diff
@@ -359,6 +359,18 @@ add_test(NAME CheckConvenienceHeaders
...
+  COMMAND
+    $<TARGET_FILE:SelfTest> "Skip benchmark macros"
+      --reporter console
+      --skip-benchmarks
+)
+set_tests_properties("Benchmarking::SkipBenchmarkMacros"
+  PROPERTIES
+    PASS_REGULAR_EXPRESSION "All tests passed \\(2 assertions in 1 test case\\)"
+    FAIL_REGULAR_EXPRESSION "benchmark name"
+)
```
horenmar, line 372: I tightened up the tests a bit. This way the tests fail if either the assertions disappear for some reason, or if the "benchmark begins" event gets sent (even if it isn't executed afterwards).

horenmar:
Thanks.

PR #2410, opened 2022-04-15
Centralize compiler detection for GCC, clang, MSVC and MINGW32.

### LINKED ISSUE #2094: Refactor and centralize compiler detection
Currently Catch2 detects compilers ad-hoc, at the place where the code needs to be compiler-specific. This brings a lot of trouble with more obscure compilers, because compilers like to masquerade for _different_ compilers, such as Clang defining GCC-version macros, ICC defining both, IBM XL self-reporting as Clang, and so on. Because compiler-specific sections of code are often indeed compiler-specific, not detecting compilers properly can be a problem, e.g. Clang will complain that it found unknown pragma if we try to disable GCC-specific warning without checking that Clang is not used.

The solution is to perform more rigorous compiler detection, and centralize it, so that updates don't have to be done in multiple places.

HoseynHeydari:
## Description
Macros detecting compilers `GCC`, `clang`, `MSVC` and `MINGW32` added to `catch_compiler_capabilities.hpp` .
New macros are replaced with former conditions that detect the compiler.
If detecting compilers needs more conditions we can just change macros
at `catch_compiler_capabilities.hpp` instead of multi places in code.

## GitHub Issues
Closes #2094

### DIFF
--- a/src/catch2/benchmark/catch_benchmark.hpp
+++ b/src/catch2/benchmark/catch_benchmark.hpp
@@ -12,6 +12,7 @@
 
 #include <catch2/interfaces/catch_interfaces_config.hpp>
 #include <catch2/internal/catch_compiler_capabilities.hpp>
+#include <catch2/internal/catch_compiler_detections.hpp>
 #include <catch2/internal/catch_context.hpp>
 #include <catch2/interfaces/catch_interfaces_reporter.hpp>
 #include <catch2/internal/catch_unique_name.hpp>
--- a/src/catch2/benchmark/catch_benchmark.hpp
+++ b/src/catch2/benchmark/catch_benchmark.hpp
@@ -12,7 +12,6 @@
 
 #include <catch2/interfaces/catch_interfaces_config.hpp>
 #include <catch2/internal/catch_compiler_capabilities.hpp>
-#include <catch2/internal/catch_compiler_detections.hpp>
 #include <catch2/internal/catch_context.hpp>
 #include <catch2/interfaces/catch_interfaces_reporter.hpp>
 #include <catch2/internal/catch_unique_name.hpp>
--- a/src/catch2/internal/catch_compiler_detections.hpp
+++ b/src/catch2/internal/catch_compiler_detections.hpp
@@ -1,0 +1,26 @@
+
+//              Copyright Catch2 Authors
+// Distributed under the Boost Software License, Version 1.0.
+//   (See accompanying file LICENSE_1_0.txt or copy at
+//        https://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef CATCH_COMPILER_DETECTIONS_HPP_INCLUDED
+#define CATCH_COMPILER_DETECTIONS_HPP_INCLUDED
+
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__ICC) && !defined(__CUDACC__) && !defined(__LCC__)
+#define CATCH_COMPILER_GCC
+#endif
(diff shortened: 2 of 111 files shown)


src/catch2/benchmark/catch_benchmark.hpp
```diff
@@ -12,6 +12,7 @@
 
 #include <catch2/interfaces/catch_interfaces_config.hpp>
 #include <catch2/internal/catch_compiler_capabilities.hpp>
+#include <catch2/internal/catch_compiler_detections.hpp>
```
HoseynHeydari, line 15: This include is not necessary.
Should be removed.

HoseynHeydari:
Some includes should be removed before merging. :| 
I try to handle it by review but there were a lot of files.

HoseynHeydari:
Some lines missing while renaming macro from `CATCH_COMPILER_MSC` to `CATCH_COMPILER_MSVC`.
After testing with clang and MSVC compiler on my machine I will push new changes.

src/catch2/internal/catch_compiler_detections.hpp
```diff
@@ -0,0 +1,26 @@
+
+//              Copyright Catch2 Authors
+// Distributed under the Boost Software License, Version 1.0.
+//   (See accompanying file LICENSE_1_0.txt or copy at
+//        https://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef CATCH_COMPILER_DETECTIONS_HPP_INCLUDED
+#define CATCH_COMPILER_DETECTIONS_HPP_INCLUDED
+
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__ICC) && !defined(__CUDACC__) && !defined(__LCC__)
```
horenmar, line 10: This approach isn't going to scale well, as pretty much every new or obscure compiler masquerades for GCC, so this line is going to keep getting longer and longer.

Instead, let's have a macro `CATCH_COMPILER_DETECTED`, that get's set when a compiler was picked. This would simplify GCC (and also clang) to something like 

```cpp
#if defined(__GNUC__) and !defined(CATCH_COMPILER_DETECTED)
#define CATCH_COMPILER_GCC
#endif
```

src/catch2/internal/catch_compiler_detections.hpp
```diff
@@ -0,0 +1,26 @@
...
+#define CATCH_COMPILER_GCC
+#endif
+
+#if defined(__clang__)
```
horenmar, line 14: This definitely doesn't filter out all the different compilers that like to pretend they are Clang, e.g. IBM XL. See above for a good way to fix this.

horenmar:
I think this PR is currently trying to do too much. I'd recommend starting by merging the reworked compiler detection macros and the code can be converted to use them iteratively later on.

tests/SelfTest/UsageTests/Condition.tests.cpp
```diff
@@ -223,12 +225,12 @@ TEST_CASE( "Comparisons with int literals don't warn when mixing signed/ unsigne
 // Disable warnings about sign conversions for the next two tests
 // (as we are deliberately invoking them)
 // - Currently only disabled for GCC/ LLVM. Should add VC++ too
-#ifdef  __GNUC__
```
HoseynHeydari: This line means if compiler is `GNUC` or `clang`.
But I just replaced it by `CATCH_COMPILER_GCC` cuases compiler warning -> error.
I'm searching for this situations in code.

PR #2420, opened 2022-05-07
ObjectStorage: port away from std::aligned_storage

### LINKED ISSUE #2419: std::aligned_storage is deprecated in C++23
`ObjectStorage` is using `std::aligned_storage`, but C++23 deprecates it, causing warnings in projects that already use `-std=c++2b`.

marcmutz:
It's deprecated in C++23. Just use alignas on a char array, wrapped in
a struct to avoid decaying to char*, which is the canonical
implementation of aligned_storage:
https://en.cppreference.com/w/cpp/types/aligned_storage#Possible_implementation

Fixes https://github.com/catchorg/Catch2/issues/2419

Catch3 is not affected.

### DIFF
--- a/include/internal/benchmark/catch_constructor.hpp
+++ b/include/internal/benchmark/catch_constructor.hpp
@@ -19,8 +19,6 @@ namespace Catch {
             template <typename T, bool Destruct>
             struct ObjectStorage
             {
-                using TStorage = typename std::aligned_storage<sizeof(T), std::alignment_of<T>::value>::type;
-
                 ObjectStorage() : data() {}
 
                 ObjectStorage(const ObjectStorage& other)
@@ -64,7 +62,7 @@ namespace Catch {
                 }
 
 
-                TStorage data;
+                struct { alignas(T) unsigned char data[sizeof(T)]; }  data;
             };
         }
 


single_include/catch2/catch.hpp
```diff
@@ -7439,7 +7437,7 @@ namespace Catch {
                     return *static_cast<T*>(static_cast<void*>(&data));
                 }
 
-                TStorage data;
+                struct { alignas(T) unsigned char data[sizeof(T)]; }  data;
```
horenmar: https://github.com/catchorg/Catch2/blob/v2.x/docs/contributing.md#using-github

> When submitting a pull request please do not include changes to the single include. This means do not include them in your git commits!

include/internal/benchmark/catch_constructor.hpp
```diff
@@ -64,7 +62,7 @@ namespace Catch {
                 }
 
 
-                TStorage data;
+                struct { alignas(T) unsigned char data[sizeof(T)]; }  data;
```
horenmar, line 65: Why the anonymous struct?

marcmutz: That's the expansion of `aligned_storage::type`, basically. I presume it's to avoid the result from decaying to a `uchar*`: https://en.cppreference.com/w/cpp/types/aligned_storage#Possible_implementation

I'll update the commit message.

single_include/catch2/catch.hpp
marcmutz: Done.

include/internal/benchmark/catch_constructor.hpp
horenmar: Hmmm, makes sense I guess.

horenmar:
Ok, thanks.

PR #2426, opened 2022-05-18
Avoid "warning C4702: unreachable code" with MSVC

hanickadot:
Small fix to avoid "warning C4702: unreachable code" on MSVC which can with /W4 can lead to error and compilation failure.


horenmar:
Thanks, but I've decided to fix this in a different way.

hanickadot:
Thanks! :)

Issue #2306, opened 2021-10-15
Unused variable warnings when compiling with nvcc + gcc


MatthiasKohl:
When compiling CUDA files with nvcc (CUDA 11.3 and GCC 9.4), when including catch2 and declaring a simple test case, I'm getting warnings such as `variable "[...]autoRegistrar1" was declared but never referenced`.
Is there a way to get rid of these warnings?

**Platform information:**

 - OS: **Linux x86_64**
 - Compiler+version: **NVCC 11.3 / GCC 9.4.0**
 - Catch version: **v2.13.0** [included as part of CMake build, with cmake 3.17.0]

MatthiasKohl:
Other issues similar to this have been opened (#2132,  #1552) but I believe they're not using a fairly recent GCC.

PR #2427, opened 2022-05-18
Suppress NVCC unused variable warnings

ahans:
## Description
This defines `CATCH_INTERNAL_START_WARNINGS_SUPPRESSION` and `CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION` when compiled with `nvcc` to suppress warnings about unused variables.

## GitHub Issues
[#2306 ](https://github.com/catchorg/Catch2/issues/2306)

### DIFF
--- a/src/catch2/internal/catch_compiler_capabilities.hpp
+++ b/src/catch2/internal/catch_compiler_capabilities.hpp
@@ -58,8 +58,9 @@
 #endif
 
 #if defined(__CUDACC__) && !defined(__clang__)
-#    define CATCH_INTERNAL_START_WARNINGS_SUPPRESSION _Pragma( "nv_diag_suppress 177" )
-#    define CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION  _Pragma( "nv_diag_default 177" )
+#    define CATCH_INTERNAL_START_WARNINGS_SUPPRESSION _Pragma( "nv_diagnostic push" )
+#    define CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION  _Pragma( "nv_diagnostic pop" )
+#    define CATCH_INTERNAL_SUPPRESS_UNUSED_VARIABLE_WARNINGS _Pragma( "nv_diag_suppress 177" )
 #endif
 
 #if defined(__clang__) && !defined(_MSC_VER)
--- a/src/catch2/internal/catch_compiler_capabilities.hpp
+++ b/src/catch2/internal/catch_compiler_capabilities.hpp
@@ -57,6 +57,11 @@
 
 #endif
 
+#if defined(__CUDACC__) && !defined(__clang__)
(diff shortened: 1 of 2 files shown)


src/catch2/internal/catch_compiler_capabilities.hpp
```diff
@@ -57,6 +57,11 @@
 
 #endif
 
+#if defined(__CUDACC__) && !defined(__clang__)
+#    define CATCH_INTERNAL_START_WARNINGS_SUPPRESSION _Pragma( "nv_diag_suppress 177" )
+#    define CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION  _Pragma( "nv_diag_default 177" )
```
horenmar: So, this would work, but wouldn't properly slot into how the macros are used across Catch2. If I am [reading the docs correctly](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#nv-diagnostic-pragmas), then the correct approach would be something like this


```suggestion
#    define CATCH_INTERNAL_START_WARNINGS_SUPPRESSION _Pragma( "nv_diagnostic push" )
#    define CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION  _Pragma( "nv_diagnostic pop" )
#    define CATCH_INTERNAL_SUPPRESS_UNUSED_VARIABLE_WARNINGS _Pragma ( "nv_diag_suppress 177" )
```

ahans: Thanks for your comment! Unfortunately, your suggestion makes the warnings come back. The reason is that in the place where the "unused" variables are introduced we do not use `CATCH_INTERNAL_SUPPRESS_UNUSED_VARIABLE_WARNINGS`, but `CATCH_INTERNAL_SUPPRESS_GLOBALS_WARNINGS`. Apparently, other compilers than nvcc would emit a different warning here. I see two solutions to this:

1. Define `CATCH_INTERNAL_SUPPRESS_GLOBALS_WARNINGS` with the suppression of 177 instead of `CATCH_INTERNAL_SUPPRESS_UNUSED_VARIABLE_WARNINGS`.
2. Leave the `#define`s as you suggested and add `CATCH_INTERNAL_SUPPRESS_UNUSED_VARIABLE_WARNINGS` to the macros in `catch_test_registry.hpp` that introduce the variables nvcc complains about.

I updated my PR to implement option 2), since I think that is the cleaner approach. After all, nvcc's warning is about a variable that is "declared but never referenced" and nothing related to global variables. Let me know what you think!

horenmar: 2. is the right solution, but it might be worth it to open up a bug upstream that a type with opaque constructor triggers unused variable warning. Usually compilers don't warn on this, because they assume that the opaque constructor contains side-effects that the user cares for.

PR #2436, opened 2022-05-24
Update documentation - add pkg-config examples.

pkubanek:
## Description

Use -std=c++14 (instead of c++11). Documents how to integrate with
pkg-config for non-CMake projects.


horenmar:
Thanks, the changes to the migration docs are useful.

The changes to examples are not, because the simple cli call doesn't handle linking the library and so on, but the changes don't make it any worse, so I'll merge it as is and fix those later.

PR #2442, opened 2022-06-02
Fix crash when running with --list-listeners and no registered listeners

loximann:
## Description
Running with `--listeners` without registered listeners caused Catch2 programs to crash.

```
$ ./a.out --list-listeners
zsh: segmentation fault (core dumped)  ./a.out --list-listeners
```
The reason was due to dereferencing an iterator pointing past the end of a vector when the vector is empty.

### DIFF
--- a/src/catch2/reporters/catch_reporter_combined_tu.cpp
+++ b/src/catch2/reporters/catch_reporter_combined_tu.cpp
@@ -154,6 +154,12 @@ namespace Catch {
 
     void defaultListListeners( std::ostream& out,
                                std::vector<ListenerDescription> const& descriptions ) {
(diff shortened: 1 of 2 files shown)


loximann:
Should I add a test case to cover listing when no listeners are registered?

horenmar:
>  Should I add a test case to cover listing when no listeners are registered?

Yes.

loximann:
I modified the fix so that defaultListListeners is single return (which in general I think makes things cleaner). With that, no test cases are needed for full coverage; however I added a test case to avoid regressions.

horenmar:
Please keep the early exit.

And coverage isn't really important, the regression check is.

PR #2448, opened 2022-06-05
Compatility fixes for GCC5

Morwenn:
I use GCC5 in my projects as the oldest GCC compiler to be C++14 conformant. Of course there are a few small issues, but I believe it is a good "oldest" target for C++14.

This PR contains a few changes to fix Catch2v3 compilation with GCC5, which has been broken since v3.0.0-preview5:
- One that seems linked to some C++17 change to inherited constructors, backported as a DR all the way back to C++11.
- One that's apparently linked to changes to `std::pair` constructor though I did not explore further to understand where it originally came from.

### DIFF
--- a/src/catch2/reporters/catch_reporter_automake.hpp
+++ b/src/catch2/reporters/catch_reporter_automake.hpp
@@ -18,6 +18,8 @@ namespace Catch {
 
     class AutomakeReporter final : public StreamingReporterBase {
     public:
+        // GCC5 compat: we cannot use inherited constructor, because it
+        //              doesn't implement backport of P0136
         AutomakeReporter(ReporterConfig&& _config):
             StreamingReporterBase(CATCH_MOVE(_config))
         {}
--- a/src/catch2/reporters/catch_reporter_automake.hpp
+++ b/src/catch2/reporters/catch_reporter_automake.hpp
@@ -8,7 +8,9 @@
 #ifndef CATCH_REPORTER_AUTOMAKE_HPP_INCLUDED
 #define CATCH_REPORTER_AUTOMAKE_HPP_INCLUDED
 
+#include <catch2/interfaces/catch_interfaces_reporter.hpp>
 #include <catch2/reporters/catch_reporter_streaming_base.hpp>
+#include <catch2/internal/catch_move_and_forward.hpp>
 
 #include <string>
 
@@ -16,7 +18,9 @@ namespace Catch {
 
     class AutomakeReporter final : public StreamingReporterBase {
     public:
-        using StreamingReporterBase::StreamingReporterBase;
+        AutomakeReporter(ReporterConfig&& _config):
+            StreamingReporterBase(CATCH_MOVE(_config))
+        {}
         ~AutomakeReporter() override;
 
         static std::string getDescription() {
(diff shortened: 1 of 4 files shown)


src/catch2/reporters/catch_reporter_automake.hpp
```diff
@@ -8,15 +8,19 @@
...
 #include <string>
 
 namespace Catch {
 
     class AutomakeReporter final : public StreamingReporterBase {
     public:
-        using StreamingReporterBase::StreamingReporterBase;
+        AutomakeReporter(ReporterConfig&& _config):
+            StreamingReporterBase(CATCH_MOVE(_config))
+        {}
```
horenmar, line 25: This is really ugly, and I give it about 3 months before someone refactors it back to the original code.

Morwenn: You mean to the current code, this change merely refactors it to the original code :')

Morwenn: More seriously, I don't think there's many ways to work around the issue. If it was one of my projects I would probably go with the following solution:

```cpp
// GCC5 compat: work around inherited constructor issues
template<typename... Args>
AutomakeReporter(Args&&... args):
    StreamingReporterBase(std::forward<Args>(args)...)
{}
```

Compared to what I submitted, it has the advantage that you don't need to change the constructor when the parent's constructor changes, which makes the maintenance easier. On the other hand it probably makes error messages slightly harder to read.

src/catch2/reporters/catch_reporter_automake.hpp
```diff
@@ -8,15 +8,19 @@
...
+#include <catch2/internal/catch_move_and_forward.hpp>
 
 #include <string>
 
 namespace Catch {
 
     class AutomakeReporter final : public StreamingReporterBase {
     public:
-        using StreamingReporterBase::StreamingReporterBase;
+        AutomakeReporter(ReporterConfig&& _config):
```
horenmar, line 23:
```suggestion
        // GCC5 compat: we cannot use inherited constructor, because it
        //              doesn't implement backport of P0136
        AutomakeReporter(ReporterConfig&& _config):
```

src/catch2/reporters/catch_reporter_cumulative_base.hpp
```diff
@@ -88,7 +90,9 @@ namespace Catch {
         using TestCaseNode = Node<TestCaseStats, SectionNode>;
         using TestRunNode = Node<TestRunStats, TestCaseNode>;
 
-        using ReporterBase::ReporterBase;
+        CumulativeReporterBase(ReporterConfig&& _config):
```
horenmar, line 95:
```suggestion
        // GCC5 compat: we cannot use inherited constructor, because it
        //              doesn't implement backport of P0136
        CumulativeReporterBase(ReporterConfig&& _config):
```

src/catch2/reporters/catch_reporter_streaming_base.hpp
```diff
@@ -8,15 +8,19 @@
...
+#include <catch2/internal/catch_move_and_forward.hpp>
 
 #include <vector>
 
 namespace Catch {
 
     class StreamingReporterBase : public ReporterBase {
     public:
-        using ReporterBase::ReporterBase;
+        StreamingReporterBase(ReporterConfig&& _config):
```
horenmar, line 23:
```suggestion
        // GCC5 compat: we cannot use inherited constructor, because it
        //              doesn't implement backport of P0136
        StreamingReporterBase(ReporterConfig&& _config):
```

Morwenn:
Thanks for the continued support :)

PR #2450, opened 2022-06-06
Fix implicit long to double conversion

joe-noel-dev:
## Description

The existing code raised a compiler warning when compiled with `-Wimplicit-int-float-conversion` using clang.


sketchylizard:
I also ran into this and was about to submit the same fix. Hopefully this will be approved and merged soon.

PR #2459, opened 2022-06-13
Change Bazel XML support to depend upon BAZEL_TEST

BrandonThomasJonesARM:
Relating to the original issue https://github.com/catchorg/Catch2/pull/2399 here;

Bazel has now been updated to have a cleaner environment variable for catch2 to know when to output to JUnit XML. 
Seen here: https://github.com/bazelbuild/bazel/pull/15393

Instead of relying on `CATCH_CONFIG_BAZEL_SUPPORT` we can now inspect the environment variable `BAZEL_TEST`. This is better as it allows for 'native' detection of both JUnit XML and Bazel. Originally not done as the detection mechanism was using too generic of a environment name.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -98,10 +98,12 @@ is equivalent with the out-of-the-box experience.
 
 
 ## Bazel support
-When `CATCH_CONFIG_BAZEL_SUPPORT` is defined, Catch2 will register a `JUnit`
-reporter writing to a path pointed by `XML_OUTPUT_FILE` provided by Bazel.
+When `CATCH_CONFIG_BAZEL_SUPPORT` is defined or when `BAZEL_TEST=1` (which is set by the Bazel inside of a test environment), 
+Catch2 will register a `JUnit` reporter writing to a path pointed by `XML_OUTPUT_FILE` provided by Bazel.
 
 > `CATCH_CONFIG_BAZEL_SUPPORT` was [introduced](https://github.com/catchorg/Catch2/pull/2399) in Catch2 3.0.1.
+
+> `CATCH_CONFIG_BAZEL_SUPPORT` was [deprecated](https://github.com/catchorg/Catch2/pull/2459) in Catch2 X.Y.Z.
 
 ## C++11 toggles
 
--- a/src/catch2/catch_config.cpp
+++ b/src/catch2/catch_config.cpp
@@ -15,6 +15,9 @@
 
 namespace Catch {
 
+    // Requires declaration for -Wmissing-declarations.
+    bool provideBazelReporterOutput();
+
     bool operator==( ProcessedReporterSpec const& lhs,
                      ProcessedReporterSpec const& rhs ) {
         return lhs.name == rhs.name &&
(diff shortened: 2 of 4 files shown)


horenmar:
Thanks, but the old approach should be kept around for backwards compatibility, until next release that breaks backcompat.

This means that Catch2 should check the environment variable for path to write the JUnit XML report to if either it was compiled with `CATCH_CONFIG_BAZEL_SUPPORT`, or if `BAZEL_TEST` is set in the environment. (I am not sure whether the value of the `BAZEL_TEST` env var should also be checked for being "1" and not e.g. "0", but I would err on the side of yes it should.)

You should then add a deprecation notice to the description of `CATCH_CONFIG_BAZEL_SUPPORT` (use placeholder similar to the "introduced in" one), and to `docs/deprecations.md`.

BrandonThomasJonesARM:
Understood, I think these are fair points. I will make the changes and update the PR. Thanks for taking a look.

docs/configuration.md
```diff
@@ -98,10 +98,10 @@ is equivalent with the out-of-the-box experience.
 
 
 ## Bazel support
-When `CATCH_CONFIG_BAZEL_SUPPORT` is defined, Catch2 will register a `JUnit`
-reporter writing to a path pointed by `XML_OUTPUT_FILE` provided by Bazel.
+When `CATCH_CONFIG_BAZEL_SUPPORT` is defined or when `BAZEL_TEST=1` (which is set by the Bazel inside of a test environment), 
+Catch2 will register a `JUnit` reporter writing to a path pointed by `XML_OUTPUT_FILE` provided by Bazel.
 
-> `CATCH_CONFIG_BAZEL_SUPPORT` was [introduced](https://github.com/catchorg/Catch2/pull/2399) in Catch2 3.0.1.
+> `CATCH_CONFIG_BAZEL_SUPPORT` was [introduced](https://github.com/catchorg/Catch2/pull/2399) in Catch2 3.0.1. It will be deprecated in Catch2 X.Y.Z.
```
horenmar:
```suggestion
> `CATCH_CONFIG_BAZEL_SUPPORT` was [introduced](https://github.com/catchorg/Catch2/pull/2399) in Catch2 3.0.1.

> `CATCH_CONFIG_BAZEL_SUPPORT` was [deprecated](https://github.com/catchorg/Catch2/pull/2459) in Catch2 X.Y.Z.
```

src/catch2/catch_config.hpp
```diff
@@ -101,6 +101,7 @@ namespace Catch {
         bool listTags() const;
         bool listReporters() const;
         bool listListeners() const;
+	bool hasBazel() const;
```
horenmar: This doesn't need to be exported (also the whitespace there is broken), it should be just a local-linkage function in `catch_config.cpp`.

horenmar: This doesn't need to be exported (also the whitespace there is broken), it should be just a local-linkage function in `catch_config.cpp`.

src/catch2/catch_config.cpp
```diff
@@ -59,27 +59,27 @@ namespace Catch {
             } );
         }
 
-#if defined( CATCH_CONFIG_BAZEL_SUPPORT )
-        // Register a JUnit reporter for Bazel. Bazel sets an environment
-        // variable with the path to XML output. If this file is written to
-        // during test, Bazel will not generate a default XML output.
-        // This allows the XML output file to contain higher level of detail
-        // than what is possible otherwise.
+	if(hasBazel()){
```
horenmar: I don't like the name much, how about `provideBazelReporterOutput`?

horenmar: I don't like the name much, how about `provideBazelReporterOutput`?

src/catch2/catch_config.cpp
```diff
@@ -113,6 +113,25 @@ namespace Catch {
     bool Config::listReporters() const      { return m_data.listReporters; }
     bool Config::listListeners() const      { return m_data.listListeners; }
 
+    bool Config::hasBazel() const {
+#ifdef CATCH_CONFIG_BAZEL_SUPPORT
+        return true;
+#endif
```
horenmar: This will cause annoying warning issues about unreachable code. The simplest solution, albeit a bit annoying, is to turn this into an `#else` and then have the code below also be conditionally compiled.

horenmar: This will cause annoying warning issues about unreachable code. The simplest solution, albeit a bit annoying, is to turn this into an `#else` and then have the code below also be conditionally compiled.

src/catch2/catch_config.cpp
```diff
@@ -113,6 +113,25 @@ namespace Catch {
...
+	
+#    if defined( _MSC_VER )
+        // On Windows getenv throws a warning as there is no input validation,
+        // since the switch is hardcoded, this should not be an issue.
+#        pragma warning( push )
+#        pragma warning( disable : 4996 )
+#    endif
+
+	return std::getenv("BAZEL_TEST") != nullptr;
```
horenmar: If you can, try to use clang-format, we provide a setting file in-repo. Otherwise, Catch2 doesn't use tabs, only spaces.

horenmar: If you can, try to use clang-format, we provide a setting file in-repo. Otherwise, Catch2 doesn't use tabs, only spaces.

tests/ExtraTests/CMakeLists.txt
```diff
@@ -137,6 +137,18 @@ set_tests_properties(CATCH_CONFIG_BAZEL_REPORTER-1
     LABELS "uses-python"
 )
 
+# We must now test this works without the build flag.
+add_executable( BazelReporterNoCatchConfig ${TESTS_DIR}/X30-BazelReporter.cpp )
+target_link_libraries(BazelReporterNoCatchConfig Catch2_buildall_interface)
```
horenmar: This doesn't need the buildall interface, as it explicitly does not touch the compile-time config.

Just linking against `Catch2WithMain` should be enough.

horenmar: This doesn't need the buildall interface, as it explicitly does not touch the compile-time config.

Just linking against `Catch2WithMain` should be enough.

horenmar:
looks good now, thanks.

PR #2463, opened 2022-06-17
CMake: Exclude empty header dirs from install

### LINKED ISSUE #2457: 3.0.1 installs empty directories include/catch2/benchmark/internal and include/catch2/generators/internal

ferdnyc:
## Revised description

Replacing the first version of this PR (which had the weakness, as @horenmar correctly noted, that it would cause any header files placed in the excluded dirs to be left out of the install), this version... well, at least it isn't susceptible to **that**.

You may **HATE** this, and I'd completely understand, but the only thing I can say is that I've tested it heavily, it works perfectly, and the resulting install is identical to the old one _except_ for the lack of empty directories.

There is one new caveat. Because it makes heavy use of the `cmake_path()` command, it only works in CMake 3.20+. (It does a version check, and the old globbed install command is still there and will be used if the CMake version is too old.)

With CMake 3.20+, though, it takes a completely different approach.

There are ample comments in the (added) source, so I won't talk it up any more as it's easier to just read the code. There are also `message(DEBUG)` commands scattered throughout, for now, so if you run `cmake` with the `--log-level=DEBUG` flag you can watch it process every header file, which might help build confidence in the logic.

Logic which, again, **I'm** very confident in, myself... but I also recognize that it's an awfully complex method of installing a bunch of header files, and I would completely accept a reaction along the lines of, "Oh, hell no!" :rofl: 

## GitHub Issues
Fixes #2457

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -263,6 +263,12 @@ set(REPORTER_SOURCES
 )
 set(REPORTER_FILES ${REPORTER_HEADERS} ${REPORTER_SOURCES})
 
+set(ALL_HEADERS
+  ${BENCHMARK_HEADERS}
+  ${INTERNAL_HEADERS}
+  ${REPORTER_HEADERS}
+)
+
 # Fixme: STATIC because for dynamic, we would need to handle visibility
 # and I don't want to do the annotations right now
 add_library(Catch2 STATIC
@@ -342,7 +348,6 @@ if (NOT_SUBPROJECT)
         ${CMAKE_INSTALL_LIBDIR}
     )
 
-
     install(
       EXPORT
         Catch2Targets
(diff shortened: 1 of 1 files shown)


ferdnyc:
The AppVeyor failure (on only one instance of the matrix) is due to a unit test failing with "Could not find 'BazelReporter.exe'", which I'm pretty sure has nothing to do with me.

horenmar:
No, this approach means that if the subfolders start containing headers, installs will break. This is categorically worse than dropping in some empty folders few levels deep in `header-install-prefix/catch2/...`.

ferdnyc:
@horenmar 

> No, this approach means that if the subfolders start containing headers, installs will break. This is categorically worse than dropping in some empty folders few levels deep in `header-install-prefix/catch2/...`.

[I tend to agree, personally.](https://github.com/catchorg/Catch2/issues/2457#issuecomment-1159304289)

ferdnyc:
@horenmar 

The other common approach is to explicitly list all of the headers in the `CMakeLists.txt` file, instead of globbing for them.

The CMake developers have always been down on globbing as a means of file discovery, since it so frequently breaks on rebuilds when new files are added. (That's less of a concern these days, as most builds now are done from-scratch each time by CI, but in the past it's been enough of an issue that it's sort of scarred them for life.) Meson carries the exact same [explicit bias against globbing](https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard).

In fact, I only just noticed, the Catch2 `CMakeLists.txt` _already has_ a list (or, lists) of headers, so there's no need to glob for them. The install can be done using `install(FILES...)`, based on the explicit lists.

It'll just take a tiny bit of processing, to ensure that the directory hierarchy is correctly preserved, since each `install(FILES...)` command can only target a single destination directory. Still, shouldn't be too bad. I'll adjust the PR.

ferdnyc:
@horenmar PR branch completely rewritten, see new introductory comment.

eli-schwartz:
> Meson carries the exact same [explicit bias against globbing](https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard).

Meson *also* just merged support for a rather clever feature of autotools... "nobase". Or in meson,

```
install_headers(
    'foo.h',
    'bar.h',
    'subdir/baz.h',
    preserve_path: true,
)
```
which avoids flattening, while not requiring... ***70***??? lines of scripting logic. :D

PR #2465, opened 2022-06-18
Update commercial-users.md

NexWeb:
Using Catch2 and now Catch3 for our client's as well as our own projects and loving it. Especially the easy C++ BDD setup and implementation.


horenmar:
Thanks

PR #2477, opened 2022-07-04
Fix nvcc compile error

### LINKED ISSUE #2476: nvcc compile error
**Describe the bug**
The `const` version of `Catch::Benchmark::Detail::ObjectStorage::stored_object` does not compile with nvcc (invalid type cast). The problem is solved by first casting to `const void*` and then to `void*` using `const_cast`.

I will make a PR ASAP.

nielskm:
Fix invalid cast error when compiling with nvcc. Fixes #2476


horenmar:
Thanks

PR #2481, opened 2022-07-18
Fix running the tests with shared library on Windows.

dimztimz:
## Description

Without this fix, the test executable fails because it can not find the dll of Catch2.


horenmar:
Thanks

PR #2485, opened 2022-07-24
Modify the install directories into convention.

hwhsu1231:
## Description

Before the modification, `*.dll` files will be installed to the `lib` folder together with `*.lib` files. However, `*.dll` files should be placed in the `bin` folder on Windows platform by convention. Therefore, I modified the install directories of `LIBRARY`, `ARCHIVE`, `RUNTIME`.


horenmar:
Thanks

PR #2487, opened 2022-07-25
Specify __cdecl on main/wmain, as required by x86 (fixes #2486).

### LINKED ISSUE #2486: CATCH_CONFIG_MAIN isn't __cdecl, causing x86 build failures
**Describe the bug**
See title

**Expected behavior**
#define CATCH_CONFIG_MAIN works on x86.

**Reproduction steps**
Building on x86:
```c++
#define CATCH_CONFIG_MAIN
#include <catch2/catch.hpp>

TEST_CASE("passes")
{
    constexpr int x{ 1 };
    constexpr int y{ 1 };
    REQUIRE(x == 1);
}

Current workaround:
```c++
#define CATCH_CONFIG_RUNNER
#include <catch2/catch.hpp>

// Using #define CATCH_CONFIG_MAIN would be nicer, but it fails to specify __cdecl, causing a build failure on x86.
extern "C" int __cdecl wmain(int argc, wchar_t* argv[])
{
    return Catch::Session{}.run(argc, argv);
}

TEST_CASE("passes")
{
    constexpr int x{ 1 };
    constexpr int y{ 1 };
    REQUIRE(x == 1);
}
```

**Platform information:**
 - OS: **Windows NT**
 - Compiler+version: **MSVC v19.30.30795.3**
 - Catch version: **v2.13.8**

davidmatson:


include/internal/catch_default_main.hpp
```diff
@@ -15,7 +15,7 @@
 
 #if defined(CATCH_CONFIG_WCHAR) && defined(CATCH_PLATFORM_WINDOWS) && defined(_UNICODE) && !defined(DO_NOT_USE_WMAIN)
 // Standard C/C++ Win32 Unicode wmain entry point
-extern "C" int wmain (int argc, wchar_t * argv[], wchar_t * []) {
+extern "C" int __cdecl wmain (int argc, wchar_t * argv[], wchar_t * []) {
```
davidmatson: Are there compilers that support wmain but don't have __cdecl? If so, we'd need to chain to another define, such as:

```c++
#ifndef CDECL
#ifdef MSVC
#define CDECL __cdecl
#else
#define CDECL
#endif
#endif
```

horenmar: I don't think there are any.

single_include/catch2/catch.hpp
```diff
@@ -17531,7 +17531,7 @@ namespace Catch {
 extern "C" int wmain (int argc, wchar_t * argv[], wchar_t * []) {
 #else
 // Standard C/C++ main entry point
-int main (int argc, char * argv[]) {
+int __cdecl main (int argc, char * argv[]) {
```
horenmar: No changes to `single_include` as part of PRs!
```suggestion
int main (int argc, char * argv[]) {
```

horenmar:
I was gonna say that I am surprised this passes CI, but v2 used to use travis for non-Windows CI and that is dead.

Sprinkling `__cdecl` around will not compile with non-MSVC compilers, so it will need to be conditional. Alternatively, the warning that causes the compilation failure with /WX can be supressed.

davidmatson:
> Sprinkling `__cdecl` around will not compile with non-MSVC compilers, so it will need to be conditional.

Done - thanks for taking a look!

include/internal/catch_default_main.hpp
```diff
@@ -13,12 +13,20 @@
 
 #ifndef __OBJC__
 
+#ifndef CDECL
```
horenmar: I don't like unnamespaced macros, so I would prefer this to be

```suggestion
#ifndef CATCH_INTERNAL_CDECL
```

davidmatson: Sure; done.

include/internal/catch_default_main.hpp
```diff
@@ -13,12 +13,20 @@
 
 #ifndef __OBJC__
 
+#ifndef CATCH_INTERNAL_CDECL
+#ifdef _MSC_VER
```
dimztimz, line 17: IMO it would be a good idea to additionally constrain this even more to just 32bit x86 with the [predefined macro _M_IX86](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170).

> `_M_IX86` Defined as the integer literal value 600 for compilations that target x86 processors. This macro isn't defined for x64 or ARM compilation targets.

davidmatson: I believe the convention, when specifying __cdecl, is always to specify it though rather than making it arch-specific. (I believe the requirement is for all architectures, it's just technically no different on some so it doesn't cause a problem in practice if it's not there.)

For example:
"On ARM and x64 processors, __cdecl is accepted but typically ignored by the compiler."
https://docs.microsoft.com/en-us/cpp/cpp/cdecl?view=msvc-170

PR #2488, opened 2022-07-26
Fix compatibility with cmake 3.8

cetitec-dje:
## Description
Fixed bugs in cmake lists file to make it compatible with older cmake version 3.8.
* project description is introduced in 3.9, reference https://cmake.org/cmake/help/latest/command/project.html
* policy cmp0069 introduced in 3.9, reference https://cmake.org/cmake/help/latest/policy/CMP0069.html

The minimum version required is set to 3.5 but the cmake lists file is not compatible with 3.8 which we use in our project. 
Tested with cmake 3.8.


horenmar:
The Mac CI failure is unrelated, it was caused by the target runner versions being deprecated.

I will merge this, but fair warning: I will move up the minimum supported CMake version relatively soon (likely after the next release).

Issue #2367, opened 2022-02-17
Empty tag in test case produces crash when generating junit report


thorsten-ottosen:
**Describe the bug**

If I write a  test case like so:

`TEST_CASE("Test requirements of some type", "[][type]")
{...}`

where we have an empty tag "[]", the program crashes when we try to generate junit output. Normal xml output works apparently.

**Expected behavior**
I would expect no crash and that the empty tag is in the output like it is for the normal xml.

**Reproduction steps**

Build your test executable in debug mode and run

`Test.exe -r junit`

This should trigger an assertion in std::string about calling front() on an empty string.

**Platform information:**
- OS: windows 10
- Compiler+version: Microsoft (R) C/C++ Optimizing Compiler Version 19.30.30709 for x64
- Catch version: v2.13.8

horenmar:
I am not convinced that empty tags are useful, but they shouldn't cause crashes (I think they should be rejected).

horenmar:
I can't reproduce this with current `devel` branch.

```cpp
TEST_CASE( "Empty tag", "[][type]" ) { REQUIRE( true ); }

$ ./tests/SelfTest
Errors occurred during startup!
  Empty tags are not allowed
```

However I am unhappy about the behaviour of `./tests/SelfTest []` if I delete the test, as it runs all the tests -- I believe it should be a cli input error and not match every test.

thorsten-ottosen:
You are right that it is of no use. I would be happy that it was rejected as invalid.

PR #2499, opened 2022-08-07
Make empty tag filter match untagged tests

### LINKED ISSUE #2498: Allow including untagged tests with other tag filters
### Description
I believe it is currently not possible to run untagged tests at the same time as some hidden tags.

Say I have tests written like this:
```c++
TEST_CASE("Foo") { ... }
TEST_CASE("Bar") { ... }
TEST_CASE("Baz", [.optional]) { ... }
TEST_CASE("Qux", [.different-optional]) { ... }
```

I would like to be able to invoke my test exe such that I can run a combination of untagged tests and hidden tests at the same time.

#### 1. Leverage the empty tag `[]` 

Currently, running `./myTests []` is the same as running `./myTests`, however the empty tag doesn't behave as expected when combined with `,`.

E.g. current behavior on the file above:
1. `./myTests []` will run `Foo` and `Bar`
2. `./myTests [],[.optional]` will only run `Baz`
3. `./myTests [],[.optional],[.different-optional]` will only run `Baz` and `Qux`

Intuitively, I would expect that since I'm using the `,` separator, all the tests ran in 1 would have been included in 2 and 3.

#### 2. Give untagged tests some new default tag

This unfortunately runs the risk of colliding with existing user tags, e.g. using `[default]` for this is probably no acceptable. But if for some reason, solution 1. above is not feasible, maybe we can find some unique enough default tag pattern such that the impact to existing users will be minimal.

#### 3. Add a new cli option
This is least likely to disrupt any existing workflows, but also not really consistent with the current cli as no other options impact the filter behavior.

Something like `./myTests [.optional] --include-untagged`

itsfrank:
### Why this change
As I noted in https://github.com/catchorg/Catch2/issues/2498 I think that it is currently impossible to run untagged tests with a hidden tag at the same time.

E.g. With a test file like this:
```c++
TEST_CASE("Foo") { ... }
TEST_CASE("Bar", "[.optional]") { ... }
```

There is no cli filter that will allow both tests to run at once. There are no filters that match untagged tests, and adding a filter to match `[.optional]` will exclude any untagged tests.

### What this change does
This change adds explicit support for using the empty tag `[]` in the cli.

Now, the `[]` tag will match all untagged tests; so in the above, a user can run `./tests [],[.optional]` to run both `Foo` and `Bar`

### How I made this change
I basically created a tag with an empty string when `[]` is passed to the cli. Previously, when the `token` within the braces was empty, the `addTagPattern` would early exit and add no tags.

Then in `TagPattern::matches` I added a condition where if the tag is empty, it will only match a test case if it's `tags` vector is also empty.

### Changes to tests
I added 2 tests that verify the behaviour I intended to add was correct; I am happy to add more if that is desired

I also had to modify an existing test in [CmdLine.tests.cpp](https://github.com/catchorg/Catch2/compare/devel...itsFrank:Catch2:make-empty-tag-matches-untagged-checks?expand=1#diff-1b7772ae8a576d129d833dade328f8b816563196a1e5fc933764c09bcfc62a0f) since that test made the assumption that the empty tag `[]` produced no filter entries and matched no tests.

After these changes, all tests pass :)

### Concerns
One conern may be how this change would interact with tests marked with the emtpy tag like so:
```c++
TEST_CASE("Foo", "[][Bar]") { ... } 
```

I tested this and it seems like the empty tag in test definitions is not allowed, I got the following error when trying to define a test with it:
```
../../../tests/SelfTest/IntrospectiveTests/TestSpecParser.tests.cpp:25: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  Empty tags are not allowed
```
So fortunately, this should be a non-issue.

## GitHub Issues
Will probably close https://github.com/catchorg/Catch2/issues/2498
May also address https://github.com/catchorg/Catch2/issues/2367


src/catch2/internal/catch_test_spec_parser.cpp
```diff
@@ -214,25 +214,23 @@ namespace Catch {
...
             } else {
+                m_currentFilter.m_required.emplace_back(Detail::make_unique<TestSpec::TagPattern>(".", m_substring));
                 m_currentFilter.m_required.emplace_back(Detail::make_unique<TestSpec::TagPattern>(token, m_substring));
             }
         }
+        if (m_exclusion) {
+            m_currentFilter.m_forbidden.emplace_back(Detail::make_unique<TestSpec::TagPattern>(token, m_substring));
+        } else {
+            m_currentFilter.m_required.emplace_back(Detail::make_unique<TestSpec::TagPattern>(token, m_substring));
+        }
```
itsfrank, line 233: the diff is a bit strange here, but this should just be removing `if (!token.empty()) {` and shifting everything else 1 indent level to the left.

tests/SelfTest/IntrospectiveTests/TestSpecParser.tests.cpp
```diff
@@ -53,3 +53,26 @@ TEST_CASE("Parsed tags are matched case insensitive",
...
+
+TEST_CASE( "Untagged tests are matched by empty tag",
+           "[test-spec][test-spec-parser]" ) {
+    auto spec = parseAndCreateSpec( "[]" );
+
+    Catch::TestCaseInfo testCase(
+        "", { "fake test name" }, dummySourceLineInfo );
+
+    REQUIRE( spec.matches( testCase ) );
+}
```
itsfrank, line 65: This may be redundant with the test `"empty tag"` section in CmdLine.tests.cpp

itsfrank:
Note: I'll be happy to edit the documentation if this change in behavior is deemed acceptable

src/catch2/catch_test_spec.cpp
```diff
@@ -42,6 +42,9 @@ namespace Catch {
     {}
 
     bool TestSpec::TagPattern::matches( TestCaseInfo const& testCase ) const {
+        if ( m_tag.empty() ) {
+            return testCase.tags.empty();
+        }
```
itsfrank, line 47: this might deserve a comment since it's really not obvious why we are doing this check

itsfrank:
Closing since in #2498 we found an existing solution to underlying issue that led me to opening this PR. After that, we could not think of reasonable use-cases for the empty tag filter.

PR #2503, opened 2022-08-21
fixes catchorg/Catch2#2401

### LINKED ISSUE #2401: With exactly 256 tests I get a warning saying there are no tests (using CMake)
```
CMake Warning at <hidden>/Catch2/v2.13.7/Catch2_Release/lib/cmake/Catch2/CatchAddTests.cmake:46 (message):
  Test executable
  '<hidden>'
  contains no tests!
```

Adding another test makes the error go away. :smile:

eiszapfen2000:
Fix return code when catch2 list options are used. The devel branch does not suffer from this problem, only v2.x.
All tests passed locally.

Closes #2401

### DIFF
--- a/include/internal/catch_session.cpp
+++ b/include/internal/catch_session.cpp
@@ -293,7 +293,7 @@ namespace Catch {
 
             // Handle list request
             if( Option<std::size_t> listed = list( m_config ) )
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks.

Issue #2366, opened 2022-02-08
GENERATE doesn't display which values caused test failures in the test report 


fschuh:
**Description**

The GENERATE macro runs a given test multiple times with different parameters, but the test report doesn't seem to provide context or information on which of those parameters are causing failures in the test.

**Additional context**

As an example, let's consider the following simple test:

```
    TEST_CASE("Test with generators")
    {
        int value = GENERATE(1, 3, 5);
        REQUIRE_FALSE(testFunction(value));
    }
```

If the test fails with generated parameter values 1 and 3, this is what the test report looks like:

```
-------------------------------------------------------------------------------
Test with generators
-------------------------------------------------------------------------------
C:\TestProject1\Tests1.cpp(131)
...............................................................................

C:\TestProject1\Tests1.cpp(134): FAILED:
  REQUIRE_FALSE( testFunction(value) )
with expansion:
  !true

-------------------------------------------------------------------------------
Test with generators
-------------------------------------------------------------------------------
C:\TestProject1\Tests1.cpp(131)
...............................................................................

C:\TestProject1\Tests1.cpp(134): FAILED:
  REQUIRE_FALSE( testFunction(value) )
with expansion:
  !true

-------------------------------------------------------------------------------
```
As we can see, there aren't any hints whatsoever that the values 1 and 3 were the reasons of those failures, while 5 was good.  
If there isn't a way to have Catch2 display the generated parameters on the report, is this something that would be feasible to add?
Are there any reasons why this information doesn't show up in the test report?

neobrain:
This is a handy feature that I found myself looking for previously too. I don't know what's blocking Catch2 from supporting it currently, but a workaround is to use dynamic sections with manual for-loops:
```cpp
    TEST_CASE("Test with generators")
    {
        for (int value : { 1, 3, 5 }) {
            DYNAMIC_SECTION("VALUE IS " << value) {
                REQUIRE_FALSE(testFunction(value));
            }
        }
    }
```

fschuh:
Another workaround is to use the CAPTURE macro.  
However, that adds extra noise to the test code, and you have to remember to use it on every test.  
It would be nice if we could get the values to show up in the report without having to add additional boilerplate to the test.

tanlin2013:
Are there any updates for this feature, after a year?

brianjmaher:
Hi, I stumbled on this issue after getting frustrated with the intrusive code changes required to bake the generator values into my section names. Are there any updates on this?

Issue #2453, opened 2022-06-08
Generator::currentElementAsString breaks compilation of a lot of our tests


stefanhaller:
In 48f3226974, a new method `currentElementAsString` was added to generators. As far as I can tell, it isn't used for anything.

The problem is that this breaks compilation of a lot of our tests, with no obvious workaround. The reason is that we define `CATCH_CONFIG_FALLBACK_STRINGIFIER` to a non-existing symbol in order to catch mistakes where failing `CHECK`s wouldn't stringify correctly. This is very useful, and actually I think it should be the default.

But this now causes compilation failures for `GENERATE` statements where the type that's generated isn't stringifiable.

To illustrate, here's a common pattern that we tend to use a lot in our tests:
```
using T = std::tuple<InputType1, InputType2, SomeHelper, ExpectedOutput1, ExpectedOutput2>;
const auto [input1, input2, helper, expectedOutput1, expectedOutput2] = GENERATE(values<T>({
  { ... },
}));

CAPTURE(input1, input2);

// ...
```

In this example, `InputType1` and `InputType2` are stringifiable, and have to be because of the `CAPTURE`. `SomeHelper` might, for example, be a `std::function`. `ExpectedOutput1` and `ExpectedOutput2` only need to be stringifiable if they appear in a `CHECK` statement, which they often do, but don't necessarily have to.

I'd appreciate advice how to work around this. I suppose I could subclass `FixedValuesGenerator` with an overwritten `stringifyImpl` method (which would only `assert(false)`?), but this seems like a lot of work.

For now we revert 48f3226974 locally, but this is not a satisfactory long-term solution either, of course.

horenmar:
That's an interesting use case for `CATCH_CONFIG_FALLBACK_STRINGIFIER`, but one that I am not convinced should be supported.

The reason for `currentElementAsString()` (and also `currentElementIndex()` added at about the same time) is to eventually allow the reporters to also include current generator state (either as element index, or as element string) in their output. This requires the ability to stringify the generator's elements, and `Detail::stringify` and the related machinery is how Catch2 does that.

stefanhaller:
> That's an interesting use case for `CATCH_CONFIG_FALLBACK_STRINGIFIER`, but one that I am not convinced should be supported.

It has saved us so many times that I wouldn't want to live without it.

[...]

PR #2509, opened 2022-08-27
Report generator states for failed assertions

gotschmarcel:
## Description

As a user of catch I often came across the situation that I wanted to know what the generator values were when a test failed. I'm aware of the CAPTURE API, but I always wanted something automatic.

I'm not entirely certain what kind of format would be preferred, so I'm happy to change and implement it for other reporters too. Currently, I've only implemented this for the console reporter.

Some approval tests don't pass even though I haven't touched these reporters. Any help is appreciated.

## GitHub Issues

#2366

### DIFF
--- a/src/catch2/generators/catch_generators.hpp
+++ b/src/catch2/generators/catch_generators.hpp
@@ -211,7 +211,7 @@ namespace Detail {
     // Note: The type after -> is weird, because VS2015 cannot parse
     //       the expression used in the typedef inside, when it is in
     //       return type. Yeah.
-    auto generate( StringRef generatorName, StringRef definition, SourceLineInfo const& lineInfo, L const& generatorExpression ) -> decltype(std::declval<decltype(generatorExpression())>().get()) {
+    auto generate( StringRef generatorName, StringRef argumentsDescription, SourceLineInfo const& lineInfo, L const& generatorExpression ) -> decltype(std::declval<decltype(generatorExpression())>().get()) {
         using UnderlyingType = typename decltype(generatorExpression())::type;
 
         IGeneratorTracker& tracker = acquireGeneratorTracker( generatorName, lineInfo );
@@ -220,7 +220,7 @@ namespace Detail {
         }
 
         auto const& generator = static_cast<IGenerator<UnderlyingType> const&>( *tracker.getGenerator() );
-        getResultCapture().trackGeneratorState(GeneratorInfo(definition, lineInfo, generator.currentElementAsString()));
+        getResultCapture().trackGeneratorState(GeneratorInfo(generatorName, argumentsDescription, lineInfo, generator.currentElementAsString()));
         return generator.get();
     }
 
--- a/src/catch2/generators/catch_generators.hpp
+++ b/src/catch2/generators/catch_generators.hpp
@@ -8,7 +8,6 @@
 #ifndef CATCH_GENERATORS_HPP_INCLUDED
 #define CATCH_GENERATORS_HPP_INCLUDED
 
-#include "catch2/interfaces/catch_interfaces_capture.hpp"
 #include <catch2/catch_tostring.hpp>
 #include <catch2/interfaces/catch_interfaces_generatortracker.hpp>
 #include <catch2/internal/catch_generator_info.hpp>
--- a/src/catch2/generators/catch_generators.hpp
+++ b/src/catch2/generators/catch_generators.hpp
@@ -8,8 +8,10 @@
 #ifndef CATCH_GENERATORS_HPP_INCLUDED
 #define CATCH_GENERATORS_HPP_INCLUDED
 
+#include "catch2/interfaces/catch_interfaces_capture.hpp"
 #include <catch2/catch_tostring.hpp>
 #include <catch2/interfaces/catch_interfaces_generatortracker.hpp>
+#include <catch2/internal/catch_generator_info.hpp>
 #include <catch2/internal/catch_source_line_info.hpp>
 #include <catch2/internal/catch_stringref.hpp>
 #include <catch2/internal/catch_move_and_forward.hpp>
@@ -210,7 +212,7 @@ namespace Detail {
     // Note: The type after -> is weird, because VS2015 cannot parse
     //       the expression used in the typedef inside, when it is in
     //       return type. Yeah.
-    auto generate( StringRef generatorName, SourceLineInfo const& lineInfo, L const& generatorExpression ) -> decltype(std::declval<decltype(generatorExpression())>().get()) {
+    auto generate( StringRef generatorName, StringRef definition, SourceLineInfo const& lineInfo, L const& generatorExpression ) -> decltype(std::declval<decltype(generatorExpression())>().get()) {
         using UnderlyingType = typename decltype(generatorExpression())::type;
 
         IGeneratorTracker& tracker = acquireGeneratorTracker( generatorName, lineInfo );
@@ -219,6 +221,7 @@ namespace Detail {
         }
 
         auto const& generator = static_cast<IGenerator<UnderlyingType> const&>( *tracker.getGenerator() );
+        getResultCapture().trackGeneratorState(GeneratorInfo(definition, lineInfo, generator.currentElementAsString()));
         return generator.get();
     }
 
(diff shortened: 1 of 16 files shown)


src/catch2/interfaces/catch_interfaces_reporter.cpp
```diff
@@ -5,6 +5,7 @@
 //        https://www.boost.org/LICENSE_1_0.txt)
 
 // SPDX-License-Identifier: BSL-1.0
+#include "catch2/interfaces/catch_interfaces_capture.hpp"
```
gotschmarcel: Those probably shouldn't be here, I believe my IDE auto imported those 🙄

src/catch2/interfaces/catch_interfaces_reporter.hpp
```diff
@@ -8,10 +8,12 @@
 #ifndef CATCH_INTERFACES_REPORTER_HPP_INCLUDED
 #define CATCH_INTERFACES_REPORTER_HPP_INCLUDED
 
+#include "catch2/interfaces/catch_interfaces_capture.hpp"
```
gotschmarcel: Those probably shouldn't be here, I believe my IDE auto imported those 🙄

src/catch2/internal/catch_run_context.hpp
```diff
@@ -8,6 +8,7 @@
 #ifndef CATCH_RUN_CONTEXT_HPP_INCLUDED
 #define CATCH_RUN_CONTEXT_HPP_INCLUDED
 
+#include "catch2/interfaces/catch_interfaces_capture.hpp"
```
gotschmarcel: Those probably shouldn't be here, I believe my IDE auto imported those 🙄

src/catch2/reporters/catch_reporter_console.cpp
```diff
@@ -153,6 +154,21 @@ class ConsoleAssertionPrinter {
                    << '\n';
         }
     }
+    void printGeneratorStates() const {
+        if ( stats.generatorInfos.empty() ) {
+            return;
+        }
+
+        stream << "with " << pluralise(stats.generatorInfos.size(), "generator"_sr) << "\n";
```
gotschmarcel: The newline should be just a character. Same below. Will fix.

src/catch2/internal/catch_run_context.cpp
```diff
@@ -319,6 +319,16 @@ namespace Catch {
         return tracker;
     }
 
+    void RunContext::trackGeneratorState( GeneratorInfo info ) {
+        // Avoid redundant entries, in case a generator is used within a loop.
```
gotschmarcel, line 323: I added the unique name to the info, so this comparison should be fine even if we have more than one generator with the same arguments on the same line. Both of them should show up. This should be tested. Where would the test go? I assume some test that runs through via the approval tests.

src/catch2/reporters/catch_reporter_console.cpp
gotschmarcel: Fixed.

src/catch2/internal/catch_run_context.hpp
gotschmarcel: Fixed

src/catch2/interfaces/catch_interfaces_reporter.hpp
gotschmarcel: Fixed

src/catch2/interfaces/catch_interfaces_reporter.cpp
gotschmarcel: Fixed

src/catch2/interfaces/catch_interfaces_capture.hpp
```diff
@@ -43,6 +44,7 @@ namespace Catch {
         virtual void sectionEndedEarly( SectionEndInfo const& endInfo ) = 0;
 
         virtual auto acquireGeneratorTracker( StringRef generatorName, SourceLineInfo const& lineInfo ) -> IGeneratorTracker& = 0;
+        virtual void trackGeneratorState( GeneratorInfo info ) = 0;
```
gotschmarcel: Should probably be `const&`

gotschmarcel: Fixed

tests/SelfTest/Baselines/console.sw.approved.txt
```diff
@@ -228,6 +228,9 @@ Generators.tests.cpp:<line number>: PASSED:
   REQUIRE( counter < 7 )
 with expansion:
   3 < 7
+with 1 generator
+  line:267: GENERATE(1, 2)
```
gotschmarcel, line 232: Unfortunately, we can't print the variable name the generator value is assigned to, like in `CAPTURE`. This is why I decided to print the `GENERATE` call including the arguments. Since this can be quite long, I put the value into a separate line. Another option would be to elide long descriptions in the middle, e.g. `GENERATE(values{1, 2, ..., 10 })`.
We could also offer an optional user description, but this would probably need a dedicated macro to not introduce a breaking change. Alternatively, something like `GENERATE(1, 2) << "UserDescription"` could work.

src/catch2/generators/catch_generators.hpp
```diff
@@ -219,6 +220,7 @@ namespace Detail {
         }
 
         auto const& generator = static_cast<IGenerator<UnderlyingType> const&>( *tracker.getGenerator() );
+        getResultCapture().trackGeneratorState(GeneratorInfo(generatorName, argumentsDescription, lineInfo, generator.currentElementAsString()));
```
gotschmarcel, line 223: I was wondering if there's maybe a "cheaper" or better way to only get the generator states, if the test fails.

src/catch2/internal/catch_run_context.cpp
```diff
@@ -319,6 +319,16 @@ namespace Catch {
         return tracker;
     }
 
+    void RunContext::trackGeneratorState( GeneratorInfo const& info ) {
+        // Avoid redundant entries, in case a generator is used within a loop.
+        if ( std::find( m_generatorInfos.cbegin(),
```
gotschmarcel, line 324: Since `info.name` is already unique, we could use `find_if` based on the name only. Saves extra some comparisons.

gotschmarcel:
@horenmar is there anything I can do to make this PR move forward? I'm happy to address change requests or fix tests. In case this implementation is approaching the problem in the completely wrong way, I'm also happy to decline this PR.

horenmar:
I don't think this is a workable approach. The issue I have with it is that I don't see how it would support something like the `random` generator in a useful manner.

I will write up what I think is a better one later this week, but the jumping off point is a recently added generator feature: https://github.com/catchorg/Catch2/commit/48f322697434292206db570e8fbe1e3ac8728d0b, but there is also an issue with this approach: #2453, that I am not 100% sure on the resolution-of.

horenmar:
My full thoughts about this:

* The description has to be dynamic, to properly capture the current output element from the generator.
* The description has to happen on-demand, so that we do not pay stringification when the reporter does not use it. There is some downside to this with multiple reporter potentially each paying the cost, but that should be relatively rare, and if it comes down to it, we can cache the stringified elements.
* This means that the basic implementation has to wrap pointers to generators that are currently active in the tracker tree, and the reporters can retrieve the string values from them. However, we should not expose the tracker tree to reporters directly, because that is waaay too low level as an interface for that.
* I am not decided whether the interface exposed to reporters should be done via free function `getActiveGenerators`, which would return limited handles to all the currently active generators (and their relative ordering), or by extending `AssertionStats`. Both have some pros and cons.

gotschmarcel:
Thanks for putting some thought into this! I already assumed that `currentElementAsString` was designed for this purpose, which is why I based my first version already on this. I'm not sure what you mean by "the description has to be dynamic, [...]." Do you mean something beyond `currentElementAsString`? 

Good point on doing this all on demand. I was wondering the [same](https://github.com/catchorg/Catch2/pull/2509#discussion_r956783962)

I guess the problem with extending `AssertionStats` is that we would always need to provide something there, even if the assertion succeeded and some reporters wouldn't benefit from it. It does offer a clean interface though to pass encapsulated information to the reporter. Maybe it would be ok to provide a function to fetch the states as part of `AssertionStats`.

Do you think it's worth for me to further look into this? I'm happy to put more effort into this feature and help you out with it, if you believe it makes sense.

horenmar:
Right, sorry. I initially only skimmed the PR, and when I found out that you are now taking string of the args to `GENERATE`, I assumed you do that to retrieve that for the reporters. I read it fully again this time, and turns out I was wrong in that assumption.

The cost to extending `AssertionStats` is relatively low. If it contains something relatively like `std::vector<GeneratorWrapper>`, where `GeneratorWrapper` is a wrapper over a single pointer to the generator, the cost is only 1 allocation per `AssertionStats`, _if_ there are any active generators. This is generally low enough that we don't mind the cost.

The other option is to add `getActiveGenerators` function to `AssertionStats`, that retrieves the generators on-demand. The advantage is that this will not add an extra allocation if the reporter does not ask for the generators, the downside is that if multiple reporters ask, we pay multiple allocations... or we have to introduce caching and `mutable` and overall the maintenance burden is unlikely to be worth the saved alloc.

In either case `GeneratorWrapper` should provide only a very limited access to the generator interface, retrieving the current element index, and asking for current element string.

-----

Also I think that the default action of a reporter should be to just use generator indices, and there being an option to get the full element strings. The full element strings are likely to be quite noisy in practice, so they should not be used unless the user asks for them.

PR #2515, opened 2022-09-07
Add missing include for VxWorks build.

Rijom:
## Description
std::min is defined in algorithm. It appears to be transitively included for most platforms. For VxWorks (a real-time operating system), however, this explicit include is required.


horenmar:
Thanks

PR #2516, opened 2022-09-08
add versioning to shared libs

jonathanspw:
I'm not sure what you have planned for ABI/API compatibility so versioning it to major.minor might be more appropriate than just major.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -304,7 +304,7 @@ endif()
 set_target_properties(Catch2 PROPERTIES 
                           DEBUG_POSTFIX "d"
                                 VERSION ${PROJECT_VERSION}
-                              SOVERSION ${PROJECT_VERSION_MAJOR})
+                              SOVERSION ${PROJECT_VERSION})
 
 # depend on bunch of C++11 and C++14 features to have C++14 enabled by default
 target_compile_features(Catch2
@@ -354,7 +354,7 @@ set_target_properties(Catch2WithMain
   PROPERTIES
     OUTPUT_NAME "Catch2Main"
         VERSION ${PROJECT_VERSION}
-      SOVERSION ${PROJECT_VERSION_MAJOR}
+      SOVERSION ${PROJECT_VERSION}
 )
 set_target_properties(Catch2WithMain PROPERTIES DEBUG_POSTFIX "d")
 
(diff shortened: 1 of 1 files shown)


horenmar:
ABI stability: NOPE
API stability: semver (modulo bugs and Hyrum's law)

dimztimz:
Becauase there is no ABI stability whatsoever, `SOVERSION` should contain all three parts of the version i.e. `SOVERSION` should be the same with `PROJECT_VERSION`.

jonathanspw:
So even minor versions will potentially change ABI?

That's pretty contradictory to how most ABI versioning works.  Can ABI stability be maintained within X.X releases so that x.x.X can be considered bugfix only, or at least non-breaking?

horenmar:
> So even minor versions will potentially change ABI?

Yes. Catch2 is meant to be used as a static library. As such, it should not be installed globally and ABI stability is a non-concern.

For various reasons, we also support building it as a dynamic library, but there is a number of limitations to it, e.g. symbols are exported only implicitly. Providing ABI stability is a maintenance cost I am not willing to pay.

src/CMakeLists.txt
```diff
@@ -301,7 +301,10 @@ if (ANDROID)
     target_link_libraries(Catch2 INTERFACE log)
 endif()
 
-set_target_properties(Catch2 PROPERTIES DEBUG_POSTFIX "d")
+set_target_properties(Catch2 PROPERTIES 
+                          DEBUG_POSTFIX "d"
+                                VERSION ${PROJECT_VERSION}
+                              SOVERSION ${PROJECT_VERSION_MAJOR})
```
horenmar:
```suggestion
                              SOVERSION ${PROJECT_VERSION})
```

src/CMakeLists.txt
```diff
@@ -350,6 +353,8 @@ target_link_libraries(Catch2WithMain PUBLIC Catch2)
 set_target_properties(Catch2WithMain
   PROPERTIES
     OUTPUT_NAME "Catch2Main"
+        VERSION ${PROJECT_VERSION}
+      SOVERSION ${PROJECT_VERSION_MAJOR}
```
horenmar:
```suggestion
      SOVERSION ${PROJECT_VERSION}
```

jonathanspw:
Thanks for the tweaks and getting this merged!

PR #2517, opened 2022-09-09
improve `-ffile-prefix-map` detection

timblechmann:
## Description

the current implementation has two problems:
* `clang-cl` does not know `-ffile-prefix-map`, but it identifies as "Clang", so the compiler will warn about an unknown compiler option
* xcode's clang however does not identify as "Clang", but as "AppleClang". so `-ffile-prefix-map` is not passed although it is supported by the compiler

### DIFF
--- a/CMake/CatchMiscFunctions.cmake
+++ b/CMake/CatchMiscFunctions.cmake
@@ -115,6 +115,6 @@ endfunction()
 function(add_build_reproducibility_settings target)
   # Make the build reproducible on versions of g++ and clang that supports -ffile-prefix-map
   if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
-    add_cxx_flag_if_supported_to_targets("-ffile-prefix-map=${CATCH_DIR}=." "${target}")
+    add_cxx_flag_if_supported_to_targets("-ffile-prefix-map=${CATCH_DIR}/=" "${target}")
   endif()
 endfunction()
--- a/CMake/CatchMiscFunctions.cmake
+++ b/CMake/CatchMiscFunctions.cmake
@@ -8,9 +8,10 @@
 
 include(CheckCXXCompilerFlag)
 function(add_cxx_flag_if_supported_to_targets flagname targets)
-    check_cxx_compiler_flag("${flagname}" HAVE_FLAG_${flagname})
+    string(MAKE_C_IDENTIFIER ${flagname} flag_identifier )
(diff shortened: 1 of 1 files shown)


jbadwaik:
I think it's a bug with CMake if they identify two different compilers, clang and clang-cl, as the same.

https://gitlab.kitware.com/cmake/cmake/-/issues/22668

horenmar:
This will need some fixes to the path normalization code in approval tests.

timblechmann:
> This will need some fixes to the path normalization code in approval tests.

do you have any pointers for the entry points to do these fixes?

horenmar:
@timblechmann Approval tests run the `SelfTest` binary and compare the output to a known good one. The implementation is in `tools/scripts/approvalTests.py`, and the problematic part is in `normalizePath` function, which looks for the absolute path to the binary in the text output, and if it finds it, it removes prefix.

`-ffile-prefix-map` also removes the prefix, but uses a slightly different prefix than the approval test normalization. This leads me to an idea; we could instead change the mapping used by `-ffile-prefix-map` to `${CATCH_DIR}/=`, so that the resulting path is the same as what the approval tests want.

CMake/CatchMiscFunctions.cmake
```diff
@@ -112,9 +113,8 @@ endfunction()
 # Adds flags required for reproducible build to the target
 # Currently only supports GCC and Clang
 function(add_build_reproducibility_settings target)
-# Make the build reproducible on versions of g++ and clang that supports -ffile-prefix-map
-  if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND NOT ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 8) OR
-     ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND NOT ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 10))
-    target_compile_options(${target} PRIVATE "-ffile-prefix-map=${CATCH_DIR}=.")
+  # Make the build reproducible on versions of g++ and clang that supports -ffile-prefix-map
+  if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
+    add_cxx_flag_if_supported_to_targets("-ffile-prefix-map=${CATCH_DIR}=." "${target}")
```
horenmar:
```suggestion
    add_cxx_flag_if_supported_to_targets("-ffile-prefix-map=${CATCH_DIR}/=" "${target}")
```

Issue #663, opened 2016-05-21
XML reporter leaves empty file if the process is killed by a signal


horenmar:
According to the [reporter documentation](https://github.com/philsquared/Catch/blob/master/docs/command-line.md#choosing-a-reporter-to-use) the xml reporter is streaming and thus I would expect at least partial output if the process is killed a by SIGINT. This is true if I run the binary as `./a.out -r xml`, but if I run the binary as `./a.out -r xml -o outfile`, I end up with an empty file.

In contrast, if I use the default or `compact` reporter, the output on stdout and in the outfile will be the same, which is a partial output until the signal was encountered.

---

Using `JUnit` reporter seems to trigger an assertion:

```
a.out: ./catch.hpp:8734: virtual void Catch::CumulativeReporterBase::testCaseEnded(const Catch::TestCaseStats &): Assertion `m_sectionStack.size() == 0' failed.
```

horenmar:
Strictly speaking this can be worked-around by redirecting std-out, but it should be investigated anyway.

philsquared:
This is potentially fixed by 84af6bc955491d4b2a4c1d3c61e0b3516ce9f298 - but needs testing

horenmar:
This seems to be fixed, just throwing in `std::endl`s is a bit of hack solution, so I am gonna tag this with revisit.

PR #2518, opened 2022-09-10
FileStream: enable automatic flushing

timblechmann:
## Description

when a test executable is killed by a signal (e.g. when executed by ctest with timeout), the reporter files are not flushed. this can lead to incomplete (or empty) report files.
to avoid this we enable automatic flushing via `std::unitbuf`

compare #663


horenmar:
Eh, sure.

PR #2521, opened 2022-09-14
Suppress -Wuseless-cast Warning

### LINKED ISSUE #2520: useless-cast warning from THROW macro family
**Describe the bug**
We get the following error message (with -Werror) when using `REQUIRE_NOTHROW` on a void function.
```
.../Catch2/src/catch2/internal/catch_test_macro_impl.hpp:79:13: error: useless cast to type 'void' [-Werror=useless-cast]
   79 |             static_cast<void>(__VA_ARGS__); \
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.../Catch2/src/catch2/catch_test_macros.hpp:128:34: note: in expansion of macro 'INTERNAL_CATCH_NO_THROW'
  128 |   #define REQUIRE_NOTHROW( ... ) INTERNAL_CATCH_NO_THROW( "REQUIRE_NOTHROW", Catch::ResultDisposition::Normal, __VA_ARGS__ )
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~
mycode.cpp:47:41: note: in expansion of macro 'REQUIRE_NOTHROW'
```

**Expected behavior**
No Warning.

**Reproduction steps**
```

void op();

TEST_CASE("") {
    REQUIRE_NOTHROW(op());
}
```
Although I was not able to reproduce it on godbolt.

**Platform information:**
 - OS: **Windows 10**
 - Compiler+version: **g++.exe (Rev1, Built by MSYS2 project) 12.2.0**
 - We have `-Wuseless-cast` in our warning list
 - Catch version: **v3.1.0**

**Additional context**
We just migrated from Catch 2.13.8 with the single include. Before the Code produced no warning.

I'm also willing to fix this with
```
                if constexpr ( std::is_same_v<decltype( __VA_ARGS__ ),      \
                                              void> ) {                     \
                    __VA_ARGS__;                                            \
                } else {                                                    \
                    static_cast<void>( __VA_ARGS__ );                       \
                }                                                           \
```

HazardyKnusperkeks:
## Description

Suppresses a warning, because we hit the warning in our test code (with -Werror).

## GitHub Issues

Closes #2520.

### DIFF
--- a/src/catch2/internal/catch_test_macro_impl.hpp
+++ b/src/catch2/internal/catch_test_macro_impl.hpp
@@ -76,7 +76,10 @@
     do { \
         Catch::AssertionHandler catchAssertionHandler( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, CATCH_INTERNAL_STRINGIFY(__VA_ARGS__), resultDisposition ); \
         try { \
+            CATCH_INTERNAL_START_WARNINGS_SUPPRESSION \
+            CATCH_INTERNAL_SUPPRESS_USELESS_CAST_WARNINGS \
             static_cast<void>(__VA_ARGS__); \
+            CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION \
             catchAssertionHandler.handleExceptionNotThrownAsExpected(); \
         } \
         catch( ... ) { \
@@ -91,7 +94,10 @@
         Catch::AssertionHandler catchAssertionHandler( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, CATCH_INTERNAL_STRINGIFY(__VA_ARGS__), resultDisposition); \
         if( catchAssertionHandler.allowThrows() ) \
             try { \
+                CATCH_INTERNAL_START_WARNINGS_SUPPRESSION \
+                CATCH_INTERNAL_SUPPRESS_USELESS_CAST_WARNINGS \
                 static_cast<void>(__VA_ARGS__); \
+                CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION \
                 catchAssertionHandler.handleUnexpectedExceptionNotThrown(); \
             } \
             catch( ... ) { \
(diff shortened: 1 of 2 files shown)


src/catch2/internal/catch_test_macro_impl.hpp
```diff
@@ -108,7 +114,10 @@
         Catch::AssertionHandler catchAssertionHandler( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, CATCH_INTERNAL_STRINGIFY(expr) ", " CATCH_INTERNAL_STRINGIFY(exceptionType), resultDisposition ); \
         if( catchAssertionHandler.allowThrows() ) \
             try { \
+                CATCH_INTERNAL_START_WARNINGS_SUPPRESSION \
+                CATCH_INTERNAL_SUPPRESS_USELESS_CAST_WARNINGS \
                 static_cast<void>(expr); \
```
HazardyKnusperkeks, line 119: Any reason this isn't a variadic macro?

HazardyKnusperkeks:
I also noticed that there is a .clang-format, but at least the from me touched code isn't clang-formated at all. Should that stay, or should I at least format the code I've added/changed?

src/catch2/internal/catch_test_macro_impl.hpp
horenmar: Legacy, it was unchanged since Catch v1, which supported C++98 compilers (which do not guarantee variadic macros).

It should be fine to change it to a variadic macro.

horenmar:
clang-format was added later in development, and it doesn't have the right options to faithfully replicate the desired formatting. The general policy is to use it on newly added/changed code, unless it messes up the overall formatting too much (and at some point figure out a way to deal with this 😃 ).

src/catch2/internal/catch_test_macro_impl.hpp
horenmar: However, upon closer inspection, I don't think this will change anything. The user-facing macro is IIRC `THROWS_AS(expr, exceptionType)`, which cannot be made meaningfully variadic either way, so there would be no benefit to changing this here.

PR #2523, opened 2022-09-14
Raise the minimum CMake version to 3.10

dimztimz:
The CMake file states 3.5 as minimal version, but 3.8 was actually minimal, see patch https://github.com/catchorg/Catch2/commit/47d56f28a9801911c048d011b375e5631dbb658f. Before that patch 3.9 was needed. This patch really fixes compatibility with 3.5.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,23 +30,21 @@ if (CMAKE_BINARY_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
     message(FATAL_ERROR "Building in-source is not supported! Create a build dir and remove ${CMAKE_SOURCE_DIR}/CMakeCache.txt")
 endif()
 
-if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.9)
+if(CMAKE_VERSION VERSION_GREATER 3.8)
   # Enable IPO for CMake versions that support it
   cmake_policy(SET CMP0069 NEW)
-  project(Catch2
-    VERSION 3.1.0 # CML version placeholder, don't delete
-    LANGUAGES CXX
-    # HOMEPAGE_URL is not supported until CMake version 3.12, which
-    # we do not target yet.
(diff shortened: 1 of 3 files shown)


horenmar:
No, thanks. When I have time for Catch2 again, I am going to instead bump up the minimum required version of CMake.

dimztimz:
Which version of Cmake do you want? My intention was to clean the file of two calls to `project()` inside if condition. That looks kinda hacky. Is 3.8 Acceptable?

dimztimz:
This patch does two things:

1. Cleans some code, looks better
2. Actually fixes a small bug. The bug is that the call to cmake_minimum_version currently is incorrect.

horenmar:
Do you have good reason for wanting to bump it up to 3.8 specifically? I looked at default packages in LTS ubuntu as a benchmark, bionic has 3.10.2, focal 3.16.3. Based on this I would target 3.16 on my own, but I am willing to hear arguments for other versions.

The cleanup makes sense, if it is still needed after the bump. I did the fastest thing I could when making the change.

dimztimz:
Well, I will list useful features that can be used.

- v3.6 Adds support for Clang-tidy static analysis
- v3.8 Adds support for C++17, support for Cpplint static analysis
- v3.9 Improves command `project()`
- v3.10 Adds support for Cppcheck static analysis.
- v3.12 Improves command `project()`.
- v3.14 Simplifies calls to command `install()`.

Personally, I'd bump it to 3.8 now, 3.10 in couple of months and I'd go above 3.10 when Ubuntu 18.04 goes out of support (precisely, in Extended Security Maintenance).

horenmar:
Ok, I am willing to keep it at 3.10 for now.

dimztimz:
I changed the PR so it raises the CMake minimum version to 3.10, and also it cleans the calls to `project()`.

horenmar:
Thanks.

Issue #2462, opened 2022-06-17
Add C++17 support that "just works" via package managers


davidmatson:
**Description**
I'd like to use code such as the following in a project that gets Catch2 via vcpkg and have it just work (not require a port overlay or patch to the vcpkg repo):
```c++
inline std::string_view test()
{
    throw std::runtime_error{"bad"};
}

TEST_CASE("Log exception")
{
    REQUIRE(test() == "good");
}
```

**Additional context**
I thought this was a vcpkg bug, but they've indicated it's an upstream bug:
https://github.com/microsoft/vcpkg/issues/25236

Similar questions have come up here before, but there isn't a solution yet:
#2210
#2046

I'm hoping we can start by getting agreement between Catch2 and vcpkg on what side the right fix would be here. (I think the starting position is both sides think the problem is on the other side, which probably needs to be resolved first in order to get to a solution.)

ferdnyc:
@davidmatson 
> I think the starting position is both sides think the problem is on the other side

I think that may be overstating it. I have yet to see anything indicating that either side _agrees_ there's a problem at all. (I'm not sure I do, either.)

The vcpkg people suggested a _**TWO-line**_ workaround that would solve this for you. The only reason that's not acceptable is your extremely narrow constraint (emphasis mine):

> I'd like to use code such as the following in a project that gets Catch2 via vcpkg **and have it just work (not require a port overlay or patch to the vcpkg repo)**

...Meeting that requirement probably involves _way_ more work than the 2-line fix you rejected, so what's the compelling argument in favor of expending all that additional effort? I get that _you_ would "like to..." do that, but to be blunt... why should any of the rest of us care?

Beyond vcpkg, there's also the option of using a different package management system. (In choosing to use one, any one, you implicitly incorporate all of its quirks and limitations into your project as well.) As someone mentioned in one of the other threads, Conan packages support an almost limitless array of feature dimensions, and separate packages can be published for any/every combination of them. That means it's possible to choose a specific variant/build that matches your requirements, however complex they may be.

vcpkg have explicitly said they _don't_ support that, at least for C++ standard specifically, which may mean that Catch2 should just be considered incompatible with vcpkg. I can't speak for anyone else, but that's not something I'd consider a bug in Catch2. (Or at least, not one I'd expend any effort addressing.) It downloads as source in mere seconds, and compiles locally in... tens of seconds.  `¯\_(ツ)_/¯`

davidmatson:
[...]

PR #2525, opened 2022-09-14
Add C++ version for bazel

dnmiller:
## Description
When adding Catch2 to the bazel workspace, the bazel build of the static libraries will fail if the C++ version of the project is less than C++14 due to use of `std::enable_if_t`. This calls out the C++ version explicitly in the compilation.

## GitHub Issues
Sort of related to #2462 if you're using bazel to manage dependencies instead of a package manager.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -78,6 +78,7 @@ cc_library(
     linkstatic = True,
     visibility = ["//visibility:public"],
     deps = [":catch2_generated"],
+    copts = ["-std=c++14"],
 )
 
 # Static library, with main.
@@ -88,4 +89,5 @@ cc_library(
     linkstatic = True,
     visibility = ["//visibility:public"],
     deps = [":catch2"],
+    copts = ["-std=c++14"],
 )


horenmar:
How does Bazel handle it if the user wants to use C++17?

dnmiller:
> How does Bazel handle it if the user wants to use C++17?

This actually only applies to the library binaries - the user can still use C++17 for their own targets.

Vertexwahn:
Adding `copts = ["-std=c++14"]` does not make sense. Since GCC and Visual Studio handle it in a different way, we have configs for this. See [`.bazelrc`](https://github.com/catchorg/Catch2/blob/devel/.bazelrc).

`bazel build --config=gcc9 //...` will add `-std=c++2a` - if you need `-std=c++14 ` than add a config to `bazel.rc` e.g. `build:my_config --cxxopt=-std=c++14`. You can then use `bazel build --config=my_config //...`. `copts = ["-std=c++14"]` will lead to compiler warnings on Visual Studio.

horenmar:
This should be no longer needed.

dnmiller:
So I have a use case where the bazel config strategy doesn't work - an old embedded compiler that doesn't know `-std=c++14`, but unit tests that run on a PC with gcc-10. 

Maybe this is more of a bazel deficiency. In the meanwhile I can patch `BUILD.bazel` in my `WORKSPACE` file.

PR #2527, opened 2022-09-17
Fix building as shared library with MSVC.

### LINKED ISSUE #2482: Tests fail to link with v3.1.0 on windows when built as shared library
**Describe the bug**
Version `v3.1.0` enables building `catch2` as a shared lib but does not export/import `static` symbols in header files.
Without proper annotations, `link.exe` on windows will not automatically import `static` variables even when they are exported by `CMake`. Currently `Selftest` cannot be built against a shared version of `catch2`.
The following linker error is produced:
```
InternalBenchmark.tests.cpp.obj : error LNK2001: unresolved external symbol "private: static class Catch::IMutableContext * Catch::IMutableContext::currentContext" (?currentContext@IMutableContext@Catch@@0PEAV12@EA)
Benchmark.tests.cpp.obj : error LNK2001: unresolved external symbol "private: static class Catch::IMutableContext * Catch::IMutableContext::currentContext" (?currentContext@IMutableContext@Catch@@0PEAV12@EA)
ToStringGeneral.tests.cpp.obj : error LNK2001: unresolved external symbol "public: static int Catch::StringMaker<float,void>::precision" (?precision@?$StringMaker@MX@Catch@@2HA)
ToStringGeneral.tests.cpp.obj : error LNK2001: unresolved external symbol "public: static int Catch::StringMaker<double,void>::precision" (?precision@?$StringMaker@NX@Catch@@2HA)
tests\SelfTest.exe : fatal error LNK1120: 3 unresolved externals
ninja: build stopped: subcommand failed.
```

**Expected behavior**
No linker errors when linking against a shared library build of `catch2`.

**Reproduction steps**
Compile and build the project as a shared library:
```
mkdir build && cd build
cmake .. -DBUILD_SHARED_LIBS=ON -DCATCH_DEVELOPMENT_BUILD=ON -GNinja
cmake --build .
````

**Platform information:**

 - OS: **Windows NT**
 - Compiler+version: **MSVC 19.30**
 - Catch version: **v3.1.0**

**Additional context**
CMake's `WINDOWS_EXPORT_ALL_SYMBOLS` exports all symbols but `link.exe` only auto links functions without the necessary `__declspec(dllimport)` annotation. `Selftest` relies on the above mentioned `static` variables that are not automatically linked even though exported.

The cleanest solution at the moment is to:
- Create a new header file
```
----- catch2/export.hpp ------
#pragma once

#if defined(CATCH_STATIC)         // Using static.
#  define CATCH_SYMEXPORT
#elif defined(CATCH_STATIC_BUILD) // Building static.
#  define CATCH_SYMEXPORT
#elif defined(CATCH_SHARED)       // Using shared.
#  ifdef _WIN32
#    define CATCH_SYMEXPORT __declspec(dllimport)
#  else
#    define CATCH_SYMEXPORT
#  endif
#elif defined(CATCH_SHARED_BUILD) // Building shared.
#  ifdef _WIN32
#    define CATCH_SYMEXPORT __declspec(dllexport)
#  else
#    define CATCH_SYMEXPORT
#  endif
#else
#  define CATCH_SYMEXPORT         // Using static or shared.
//#  error define CATCH_STATIC or CATCH_SHARED preprocessor macro to signal test library type being linked
#endif
```
- Include `catch2/export.hpp` in necessary header files (Minimum `catch2/internal/catch_context.hpp` and `catch2/catch_tostring.hpp`) as well as to `src/CMakeLists.txt`
- Annotate `catch2/internal/catch_context.hpp:31` with `CATCH_SYMEXPORT`
- Annotate `catch2/catch_tostring.hpp:299 and 305` with `CATCH_SYMEXPORT`
- Add necessary compile definitions to CMake
```
target_compile_definitions(Catch2 PRIVATE CATCH_SHARED_BUILD INTERFACE CATCH_SHARED)
```

dimztimz:
Fixes #2482.

For some more info see https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html.

> For global data symbols, __declspec(dllimport) must still be used when compiling against the code in the .dll.

See also this https://cmake.org/cmake/help/latest/prop_tgt/DEFINE_SYMBOL.html . That is where the define `Catch2_EXPORTS` comes from.

### DIFF
--- a/CMake/CatchConfigOptions.cmake
+++ b/CMake/CatchConfigOptions.cmake
@@ -67,6 +67,7 @@ set(_OtherConfigOptions
 foreach(OptionName ${_OtherConfigOptions})
   AddConfigOption(${OptionName})
 endforeach()
+set(CATCH_CONFIG_SHARED_LIBRARY ${BUILD_SHARED_LIBS})
 
 set(CATCH_CONFIG_DEFAULT_REPORTER "console" CACHE STRING "Read docs/configuration.md for details. The name of the reporter should be without quotes.")
 set(CATCH_CONFIG_CONSOLE_WIDTH "80" CACHE STRING "Read docs/configuration.md for details. Must form a valid integer literal.")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -409,6 +409,10 @@ if (CATCH_BUILD_EXAMPLES OR CATCH_BUILD_EXTRA_TESTS)
         $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/generated-includes>
         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
     )
+    target_compile_definitions(Catch2_buildall_interface
+      INTERFACE
+        CATCH_CONFIG_STATIC
+    )
     target_compile_features(Catch2_buildall_interface
       INTERFACE
         cxx_alignas
--- a/src/catch2/catch_tostring.hpp
+++ b/src/catch2/catch_tostring.hpp
@@ -296,13 +296,13 @@ namespace Catch {
     template<>
     struct StringMaker<float> {
         static std::string convert(float value);
-        static int precision;
+        CATCH_EXPORT static int precision;
     };
 
     template<>
     struct StringMaker<double> {
         static std::string convert(double value);
-        static int precision;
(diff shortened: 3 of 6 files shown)


horenmar:
Thanks. I am willing to merge this now, but I'd also like to see a PR that adds a test build for the dynamic library.

PR #2528, opened 2022-09-18
Bazel build: Update of skylib

Vertexwahn:
This PR only affects the Bazel Build of Catch2.

- Update of skylib
- Add config for gcc11
- Remove not needed load of `cc_rules`
- Proper handling of `CATCH_CONFIG_SHARED_LIBRARY`


Vertexwahn:
@horenmar Can you please review?

Vertexwahn:
@horenmar Currently the Bazel build of Catch2 is broken. This PR is going to fix this (proper handing of `CATCH_CONFIG_SHARED_LIBRARY`)

.bazelrc
```diff
@@ -1,4 +1,10 @@
+build --enable_platform_specific_config
+
 build:gcc9 --cxxopt=-std=c++2a
+build:gcc11 --cxxopt=-std=c++2a
 build:clang13 --cxxopt=-std=c++17
 build:vs2019 --cxxopt=/std:c++17
-build:vs2022 --cxxopt=/std:c++17
\ No newline at end of file
+build:vs2022 --cxxopt=/std:c++17
```
horenmar, line 7: Is there no better way to do this, than to list all the different compiler versions?

Vertexwahn: Short answer: no;

Long answer: it is nice for every Bazel-based project if it can be built using `bazel build //...` (`//...` means all targets - i.e. build all targets). Since some projects (e.g. C++ compilers) need special treatment to support different C++ versions you need specialized flags - one clean and minimal invasive way is to use Bazel configs, e.g. `bazel build --config=<config_name> //...`.

 `build --enable_platform_specific_config` tells Bazel to autoselect a config named `windows`, `linux` or `macos` depending on the operating system you are using. This way `bazel build //...` on Ubuntu 22.04 will select the `gcc11` config. Nevertheless, you can still run on a Linux distro that only has Clang available, or you config uses a flag not recognized by the compiler. Therefore, there can and will be cases where enable platform specific config feature does not work, but in some cases it can make a `bazel build //...` work.

horenmar: Very well.

PR #2530, opened 2022-09-19
Add support for building with Meson

mikecrowe:
The Meson[1] build system makes it easier incorporate third-party libaries into a project if they also build using Meson.

Let's add a minimal Meson build that's compatible with the CMake build, along with a GitHub workflow to check that it builds and that at least the simplest SelfTest runs.

The handling of catch_user_config.hpp is inspired by BUILD.bazel and doesn't attempt to support any configuratons options. Such features could be added later.

Meson strongly discourages using wildcards to specify sources, so the source and header lists are copied from CMakeLists.txt.

Add a new test workflow to test the Meson builds. I was unable to get these tests to pass with Ubuntu 20.04, so they use Ubuntu 22.04.

I'm neither a CMake nor a Meson expert, but the results seem to work for me.

[1] https://mesonbuild.com/

### DIFF
--- a/.github/workflows/linux-meson-builds.yml
+++ b/.github/workflows/linux-meson-builds.yml
@@ -1,0 +1,43 @@
+name: Linux builds (basic) using meson build system
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: meson ${{matrix.cxx}}, C++${{matrix.std}}, ${{matrix.build_type}}
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        cxx:
+          - g++-11
+          - clang++-11
+        build_type: [debug, release]
+        std: [14, 17]
+        include:
+          - cxx: clang++-11
+            other_pkgs: clang-11
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Prepare environment
+      run: sudo apt-get install -y meson ninja-build ${{matrix.other_pkgs}}
+
+    - name: Configure build
+      env:
+        CXX: ${{matrix.cxx}}
+        CXXFLAGS: -std=c++${{matrix.std}} ${{matrix.cxxflags}}
+      # Note: $GITHUB_WORKSPACE is distinct from ${{runner.workspace}}.
+      #       This is important
+      run: |
+        meson -Dbuildtype=${{matrix.build_type}} ${{runner.workspace}}/meson-build
+
+    - name: Build tests + lib
+      working-directory: ${{runner.workspace}}/meson-build
+      run: ninja
+
+    - name: Run tests
+      working-directory: ${{runner.workspace}}/meson-build
+      # Hardcode 2 cores we know are there
+      run: |
+        meson test --verbose
--- a/meson.build
+++ b/meson.build
@@ -9,9 +9,9 @@ project(
   'catch2',
   'cpp',
   version : '3.1.0',
-  license: 'Boost 1.0')
-pkg = import('pkgconfig')
-cpp = meson.get_compiler('cpp')
+  license: 'BSL-1.0',
+  meson_version: '>=0.49.0'
+)
 
 subdir('src/catch2')
 subdir('tests')
--- a/meson.build
+++ b/meson.build
@@ -8,7 +8,7 @@
 project(
   'catch2',
   'cpp',
-  version : '3.1.0',
+  version : '3.1.0', # CML version placeholder, don't delete
   license: 'BSL-1.0',
   meson_version: '>=0.49.0'
 )
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,17 @@
+#              Copyright Catch2 Authors
+# Distributed under the Boost Software License, Version 1.0.
+#   (See accompanying file LICENSE_1_0.txt or copy at
+#        https://www.boost.org/LICENSE_1_0.txt)
+
+# SPDX-License-Identifier: BSL-1.0
+
+project(
+  'catch2',
+  'cpp',
+  version : '3.1.0',
+  license: 'Boost 1.0')
+pkg = import('pkgconfig')
+cpp = meson.get_compiler('cpp')
+
+subdir('src/catch2')
+subdir('tests')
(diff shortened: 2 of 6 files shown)


mikecrowe:
These changes are being improved based on review comments in https://github.com/mesonbuild/wrapdb/pull/643 . Once that review is complete I'll update this PR.

horenmar:
Why should the build be part of this project, rather than just a recipe in wrapdb?

mikecrowe:
> Why should the build be part of this project, rather than just a recipe in wrapdb?

I hoped that it would be more likely to stay up to date if it was included in Catch2 itself, just like the `BUILD.bazel` file is. However, if you're not interested in that then it can stay separate.

eli-schwartz:
It's indeed more trusted (all content from the original project), easier to use (vs. patching in a secondary source), more likely to remain updated, and works better as a git submodule, or when updating to arbitrary commit checkpoints as the wrapdb will never add support for those.

If, as implied by e.g. the bazel support, the Catch2 project is open to hosting integrations directly, that's the superior solution and the externally patched-in overlay can be dropped from the wrapdb (which may still host a metadata-only record).

horenmar:
I have two issues with keeping Meson build support in-repo.

1) I don't use Meson, so if it breaks I will have to go and learn new build tool
2) Because it uses explicit file lists, it adds another place to edit when adding another file.

Neither of these are deal breakers, but they are not ideal.

The positive side is that the configuration can be done sanely (unlike Bazel), and that there will be few CI jobs to keep the build functional.

eli-schwartz:
Speaking with my @mesonbuild core committer hat on, you are more than welcome to ping me and ask for help at any time.

Duplicating file lists across build systems isn't particularly fun, but one possible option is to have a text file with one source filename per line, and read it into each build system. For meson this would be something like:

```meson
fs = import('fs')

sources = files(
    fs.read('srclist.txt').split()
)
```

(Meson will automatically reconfigure if that file list changes, because it's treated as a build definition file.)

horenmar:
> Speaking with my https://github.com/mesonbuild core committer hat on, you are more than welcome to ping me and ask for help at any time.

Very well.

horenmar:
@mikecrowe Is this ready for review?

mikecrowe:
> @mikecrowe Is this ready for review?

I think so. Would you like me to squash it down to a single commit first?

meson.build
```diff
@@ -0,0 +1,17 @@
...
+# Distributed under the Boost Software License, Version 1.0.
+#   (See accompanying file LICENSE_1_0.txt or copy at
+#        https://www.boost.org/LICENSE_1_0.txt)
+
+# SPDX-License-Identifier: BSL-1.0
+
+project(
+  'catch2',
+  'cpp',
+  version : '3.1.0',
```
horenmar: This should either slurp the version from the source files, or the update script in `tools/scripts/releaseCommon.py` needs to be updated to update this file as well.

tests/meson.build
```diff
@@ -0,0 +1,71 @@
...
+  'SelfTest/UsageTests/Tricky.tests.cpp',
+  'SelfTest/UsageTests/VariadicMacros.tests.cpp',
+  'SelfTest/UsageTests/MatchersRanges.tests.cpp',
+  'SelfTest/UsageTests/Matchers.tests.cpp'
+]
+
+# This isn't as good as the CMake tests, but it proves that we've
+# actually put something in the library files.
+self_test = executable(
+  'self_test',
```
horenmar: Why does this use different filename than the SelfTest binary in CMake build?

meson.build
horenmar: Related, does Meson support the idea of "in-development" versions? I generally dislike the fact that we use the 3.1.0 version number for what is the development version of next release, but most tools don't support development versions too well.

mikecrowe: I think that there are two alternatives here:

1. Call a script that uses the code in `releaseCommon.py` to read the version number out of `catch_version.cpp`. (This may require a more recent version of Meson than is in Debian 11, which complicates things.)
2. Teach `releaseCommon.py` to also put the version number in `meson.build`. This looks like it would be straightforward and Meson would be no worse that CMake.

Regarding in-development versions:

It looks like if we were to require a sufficiently-new version of Meson (newer than is in Debian 11) then it would be possible to use something like:

```meson
project(
  'catch2',
  'cpp',
  version : run_command('git', 'describe', check: true).stdout().strip(),
  license: 'BSL-1.0',
  meson_version: '>=0.49.0'
)
```
so that the version number ends up being generated automatically by Git. The problem with this is that the release scripts would need to patch this with the actual version when producing the tarball and ZIP file releases.

See also: https://stackoverflow.com/questions/59201214/can-the-meson-project-version-be-assigned-dynamically

tests/meson.build
mikecrowe: No good reason. I'll fix it.

meson.build
mikecrowe: I believe that I've implemented option 2.

meson.build
```diff
@@ -0,0 +1,17 @@
...
+# Distributed under the Boost Software License, Version 1.0.
+#   (See accompanying file LICENSE_1_0.txt or copy at
+#        https://www.boost.org/LICENSE_1_0.txt)
+
+# SPDX-License-Identifier: BSL-1.0
+
+project(
+  'catch2',
+  'cpp',
+  version : '3.1.0', # CML version placeholder, don't delete
```
horenmar, line 11: As an FYI, the "CML" stands for "CMakeList(s)". 😃 
Feel free to leave it in though.

meson.build
horenmar: Yeah, given that's what we already do for other places, adding it to `releaseCommon` was the right choice.

meson.build
mikecrowe: Aha! I spent a while searching trying to work out what it stood before and then decided to just copy it exactly. :) Thanks.

.github/workflows/linux-meson-builds.yml
```diff
@@ -0,0 +1,43 @@
...
+      matrix:
+        cxx:
+          - g++-11
+          - clang++-11
+        build_type: [debug, release]
+        std: [14, 17]
+        include:
+          - cxx: clang++-11
+            other_pkgs: clang-11
+
```
mikecrowe, line 19: I reduced the build matrix whilst I was trying to make this workflow work. Would you like me to try widening it to more compiler versions now it seems to be working?

horenmar: Nah, this is fine.

PR #2537, opened 2022-09-24
Display type name in TEMPLATE_LIST_TEST_CASE output

cschreib:
## Description
Currentlly, when using `TEMPLATE_LIST_TEST_CASE` to create a templated test case, the test output refers to each type in the list by index:
```
Template test case with test types specified inside std::tuple - MyTypes - 0
Template test case with test types specified inside std::tuple - MyTypes - 1
Template test case with test types specified inside std::tuple - MyTypes - 2
```

This is easy to handle when the type list contains few elements, but does not scale very well to long type lists, and requires looking at the test code to figure out which index corresponds to which type. This PR changes the output to also show the name of the type:
```
Template test case with test types specified inside std::tuple - MyTypes - 0 - int
Template test case with test types specified inside std::tuple - MyTypes - 1 - char
Template test case with test types specified inside std::tuple - MyTypes - 2 - float
```

This is currently done using `typeid(T).name()`. In libstdc++, the name returned by this function is mangled, so we also have to de-mangle it using standard compiler functions. On MSVC no de-mangling is required. I don't know about other compilers. 

There are other options available to get a human-readable name for a type, including macro tricks with `__PRETTY_FUNCTION__` et al. I figured these were less portable, so shied away from them.

## GitHub Issues
None


tophyr:
Does this work without RTTI? I'd be interested in your opinion of https://github.com/catchorg/Catch2/pull/2771/commits/1bc18d6d99c1163506715fdb05e23fd3f0ed53ce which aims at the same goal, but doesn't necessitate RTTI. I think we could integrate your `typeid` ideas into that fairly easily, also.

cschreib:
I think it does require RTTI. In [snitch](https://github.com/snitch-org/snitch/blob/main/include/snitch/snitch_type_name.hpp) I ended up using the `__PRETTY_FUNCTION__` trick, which doesn't require RTTI and works well on all three major compilers. I would actually recommend switching to this other method.

tophyr:
Thanks for the idea @cschreib! I have integrated your ideas, along with the RTTI default exhibited here and the original customization-point idea I had, in https://github.com/catchorg/Catch2/pull/2771.

PR #2538, opened 2022-09-28
Unused variable removed

evsergey:
## Description
I had warnings from CLang about the variable being assigned, but never used


horenmar:
Thanks, that looks like a left-over from the old way reporter specs were handled.

PR #2539, opened 2022-10-03
meson cleanups with muon

neheb:
muon: https://sr.ht/~lattis/muon/ has a formatter that can be used to clean up meson build files. Similar to clang-format.


meson.build
```diff
@@ -8,9 +8,9 @@
 project(
   'catch2',
   'cpp',
-  version : '3.1.0', # CML version placeholder, don't delete
+  version: '3.1.0', # CML version placeholder, don't delete
```
eli-schwartz, line 11: This would have to be updated at https://github.com/catchorg/Catch2/blob/4db8b50aab9f91f31b2953455091e257942ec188/tools/scripts/releaseCommon.py#L96-L97

neheb: dunno, this is just a formatting PR.

eli-schwartz: I presume the formatting PR would prefer to not have this formatting be constantly overwritten and reverted to the previous formatting by a version bumping script...

neheb: yeah I see what you mean. Updated.

src/catch2/meson.build
```diff
@@ -132,133 +109,155 @@ internal_headers = [
...
+  'catch_tag_alias_autoregistrar.cpp',
+  'catch_test_case_info.cpp',
+  'catch_test_spec.cpp',
+  'catch_timer.cpp',
+  'catch_tostring.cpp',
+  'catch_totals.cpp',
+  'catch_version.cpp',
+)
 
 reporter_headers = [
```
horenmar, line 260: Why are some changed to `files` and some are not?

--e--

Actually I see it is a headers vs sources split, but same question.

neheb: I mentioned it in the commit message. split() is used later on. That only works for strings, not files.

horenmar:
OK, thanks.

PR #2540, opened 2022-10-03
fix compilation on cygwin

### LINKED ISSUE #2511: strnlen and Cygwin compatability - catch_tostring.hpp
I was trying to make autodiff (https://autodiff.github.io/installation/) using cmake and the most recent version of Catch2 on Cygwin, but would always run into an error at the cmake --build . --target install. I finally narrowed down my issue thanks to this error message : 

/usr/local/include/catch2/catch_tostring.hpp: In static member function ‘static std::string Catch::StringMaker<char [SZ]>::convert(const char)’:
/usr/local/include/catch2/catch_tostring.hpp:214:33: error: there are no arguments to ‘strnlen’ that depend on a template parameter, so a declaration of ‘strnlen’ must be available [-fpermissive]

I think the issue is that in the file catch_tostring.hpp there is a call to 'strnlen' that is unfortunately not supported in Cygwin (https://github.com/microsoft/GSL/issues/331). 

I have tried to debug my issue by editing this file and changing the strnlen calls. 

I understand there are several limitations to using cygwin and I am posting this here in case anyone else has this issue. Please let me know if this is not a bug and there was an issue in my implementation.

### LINKED ISSUE #2529: Build Error on Cygwin
**Describe the bug**

I get the following error using gcc 11.3.0 on cygwin:
```
/cygdrive/c/projects/clingo/third_party/catch/src/catch2/../catch2/catch_tostring.hpp:214:33: error: there are no arguments to 'strnlen' that depend on a template parameter, so a declaration of 'strnlen' must be available [-fpermissive]
  214 |                 StringRef( str, strnlen( str, SZ ) ) );
```

What about using `strnlen_s` instead? The only problem I see is that additional macros would have to be defined but since the templates are fully instantiated, the required code could be moved to the source file.

There might also be another header providing a `strnlen`. I could dig a little deeper if help is required.

**Platform information:**

 - OS: **Windows + Cygwin**
 - Compiler+version: **GCC v11.3.0**
 - Catch version: **v3.1.0**

rkaminsk:
## Description

Use std::char_traits<char>::find instead of strnlen for better portability.

## GitHub Issues

- Closes  #2529.
- Closes #2511.

### DIFF
--- a/src/catch2/catch_tostring.hpp
+++ b/src/catch2/catch_tostring.hpp
@@ -13,7 +13,6 @@
 #include <cstddef>
 #include <type_traits>
 #include <string>
-#include <string.h>
 
 #include <catch2/internal/catch_compiler_capabilities.hpp>
 #include <catch2/internal/catch_config_wchar.hpp>
@@ -41,6 +40,13 @@ namespace Catch {
 
     namespace Detail {
 
+        inline std::size_t catch_strnlen(const char *str, std::size_t n) {
+            auto ret = std::char_traits<char>::find(str, n, '\0');
+            if (ret != nullptr) {
+                return ret - str;
+            }
+            return n;
+        }
 
         constexpr StringRef unprintableString = "{?}"_sr;
 
@@ -208,28 +214,24 @@ namespace Catch {
     template<size_t SZ>
     struct StringMaker<char[SZ]> {
         static std::string convert(char const* str) {
-            // Note that `strnlen` is not actually part of standard C++,
-            // but both POSIX and Windows cstdlib provide it.
             return Detail::convertIntoString(
-                StringRef( str, strnlen( str, SZ ) ) );
+                StringRef( str, Detail::catch_strnlen( str, SZ ) ) );
         }
     };
     template<size_t SZ>
     struct StringMaker<signed char[SZ]> {
         static std::string convert(signed char const* str) {
-            // See the plain `char const*` overload
             auto reinterpreted = reinterpret_cast<char const*>(str);
             return Detail::convertIntoString(
-                StringRef(reinterpreted, strnlen(reinterpreted, SZ)));
+                StringRef(reinterpreted, Detail::catch_strnlen(reinterpreted, SZ)));
         }
     };
     template<size_t SZ>
     struct StringMaker<unsigned char[SZ]> {
         static std::string convert(unsigned char const* str) {
-            // See the plain `char const*` overload
             auto reinterpreted = reinterpret_cast<char const*>(str);
             return Detail::convertIntoString(
-                StringRef(reinterpreted, strnlen(reinterpreted, SZ)));
+                StringRef(reinterpreted, Detail::catch_strnlen(reinterpreted, SZ)));
         }
     };
 


extras/catch_amalgamated.cpp
```diff
@@ -6,7 +6,7 @@
 // SPDX-License-Identifier: BSL-1.0
 
 //  Catch v3.1.0
-//  Generated: 2022-07-17 20:14:05.885021
```
dimztimz: You should not regenerate this file on PR. It's generated only once before releasing a version by the main maintainer.

extras/catch_amalgamated.hpp
```diff
@@ -6,7 +6,7 @@
 // SPDX-License-Identifier: BSL-1.0
 
 //  Catch v3.1.0
-//  Generated: 2022-07-17 20:14:04.055157
```
dimztimz: Same, should not be generated on PR.

src/catch2/catch_tostring.hpp
```diff
@@ -41,6 +41,13 @@ namespace Catch {
 
     namespace Detail {
 
+        inline std::size_t catch_strnlen(const char *str, std::size_t n) {
+            auto ret = std::memchr(str, '\0', n);
```
dimztimz: Better to use [std::char_traits<char>::find()](https://en.cppreference.com/w/cpp/string/char_traits/find). That has proper types and not `void*`. it's also `constexpr` if needed.

src/catch2/catch_tostring.hpp
```diff
@@ -13,7 +13,7 @@
 #include <cstddef>
 #include <type_traits>
 #include <string>
-#include <string.h>
+#include <cstring>
```
dimztimz: This header is probably not needed at all if the char_traits thingy is used, see comment bellow.

src/catch2/catch_tostring.hpp
```diff
@@ -41,6 +41,13 @@ namespace Catch {
 
     namespace Detail {
 
+        inline std::size_t catch_strnlen(const char *str, std::size_t n) {
+            auto ret = std::memchr(str, '\0', n);
+            if (ret != nullptr) {
+                return static_cast<const char *>(ret) - str;
```
dimztimz: No need for static cast if you use the `char_traits` thingy.

dimztimz (changes requested):

extras/catch_amalgamated.cpp
rkaminsk: I was not sure if the CI tests would generate it because it seems to be tested. It can easily be undone because it is a separate commit.

src/catch2/catch_tostring.hpp
rkaminsk: If you prefer the `std::char_traits` solution, I can change it. It is part of the `<string>` header.

extras/catch_amalgamated.hpp
rkaminsk: Removed.

extras/catch_amalgamated.cpp
rkaminsk: Removed.

src/catch2/catch_tostring.hpp
rkaminsk: Done. It is actually the nicer solution.

src/catch2/catch_tostring.hpp
rkaminsk: I removed the header.

src/catch2/catch_tostring.hpp
rkaminsk: Done.

rkaminsk:
The build errors on mac don't look like they have anything to do with this PR. Looks more like a gcc problem on MacOS.

horenmar:
Yep, the Mac issue seems to be a known issue with new toolchain, and I guess the MacOS image got recently updated.

Issue #2519, opened 2022-09-11
Update documentation code examples to include required headers


ton:
It would be good to update the code examples in the documentation to include the headers that are (minimally) required to use a certain feature. For example, I am reading up on the usage of generators. The only thing preventing me from using them right away in my unit test is that I have to check my local Catch2 include directories and guess which header I would need to include to use the feature.

LukasP:
Migrating a project to v3 right now and running into the same issue, there is no documentation specifying which feature is in which header, not even a comment in the header files themselves.

nandanvasudevan:
Let consider this example: https://github.com/catchorg/Catch2/blob/97313f9033cffde2002e4a377229559d360d7f73/docs/benchmarks.md?plain=1#L47-L52

Would something like this be more helpful? This ones a little bit on the nose, but just to get the idea.

## Required includes
```cpp

#include <catch2/benchmark/catch_benchmark.hpp>

```

> ## Benchmark specification
> 
> Benchmarks can be specified anywhere inside a Catch test case.
> There is a simple and a slightly more advanced version of the `BENCHMARK` macro.
> 
> Let's have a look how a naive Fibonacci implementation could be benchmarked:

ton:
Personally, I would include the `#include` directive directly inside the code examples, that is a bit cleaner I would say, but what you propose is also a big improvement.

Issue #2513, opened 2022-09-06
Update docs to include how to use lambdas within regular test cases?


robamu:
Hello,

The feature of nesting multiple sections into each other is really useful. Sometimes, I'd prefer something like functions though, where I can share the same code (which might include CHECKs and REQUIREs) across multiple sections. Maybe I missed something in the documentation, but I have found C++ lambdas to be extremely useful for this, and I can't use regular function definitions inside the `TEST_CASE` scope.

For example consider something like this

```cpp
#include <catch2/catch_test_macros.hpp>

TEST_CASE("Test", "[test]") {
    uint32_t answer = 42;
    uint32_t foo = 10;

    auto sharedTestCode = [&](bool changeAnswer) {
        if(changeAnswer) {
            answer = foo;
        }
        CHECK(foo == 10);
    };

    SECTION("Section 0") {
        sharedTestCode(false);
        CHECK(answer == 42);
    }

    SECTION("Section 1") {
        sharedTestCode(true);
        CHECK(answer == 10);
    }

}
```

Do you think this is something worth mentioning in the doucmentation? Maybe this is also a case where a test fixture would be a better option in that case, but using lambdas has worked well enough for me to use the recommended way to use Catch2.

Kind Regards
Robin Müller

nandanvasudevan:
I do this all the time, isn't it just C++?

robamu:
One could argue it's just C++, but it is usually possible to do work in C++ without lambdas entirely, especially when coming from a C background. In the context of Catch2 I have not found a way to have something similar to functions inside a Catch2 test case without lambdas. That's why I though it might be a nice idea to mention/include it in the docs.

PR #2541, opened 2022-10-05
Just some documentation.

nandanvasudevan:
## Description

Adding includes to the documentation and a basic test case with a lambda.
## GitHub Issues

#2519 - Added headers for parts of Catch2 that I am aware of.
#2513 - Basic test case with a lambda.


nandanvasudevan:
Is there anything I should do here?

horenmar:
Hi, sorry for taking so long. Your PR happens to be in the goldilocks zone of not being good enough to approve outright, not being bad enough to reject and the subject matter not being interesting enough to prioritize (small doc update versus `SKIP` or adding support for comparisons including `std::*_ordering`).

Anyway, I think the PR should be split into two, so they can progress separately. I recommend moving the lambda part into its own PR (I'll leave some notes on that separately), and leaving the include docs in this one.

docs/assertions.md
```diff
@@ -177,6 +186,20 @@ TEST_CASE_METHOD((Fixture<int, int>), "foo", "[bar]") {
...
+#include <catch2/catch_test_macros.hpp>
+
+TEST_CASE("Dumb") {
+    auto fnTest = [](const bool bPass) { CHECK(bPass); };
+
+    fnTest(true);
+}
+```
+[godbolt](https://catch2.godbolt.org/z/ebdr9vKcj)
+
```
horenmar, line 202: This shows, but doesn't explain to non-experts, an important feature: the fact that assertions can happen outside the `TEST_CASE`.

There is also something interesting to be said about lambdas specifically: how do their capture lists behave across multiple invocations (due to `SECTION`s).

So I would like to see a bit of text first explaining assertions in separate functions, then another piece about lambdas.

docs/assertions.md
```diff
@@ -16,6 +16,9 @@ Catch is different. Because it decomposes natural C-style conditional expression
 Most of these macros come in two forms:
 
 ## Natural Expressions
+```cpp
+#include <catch2/catch_test_macros.hpp>
+```
```
horenmar, line 21: I like this format for larger sections (ideally with their own headings) 👍

docs/generators.md
```diff
@@ -145,6 +148,16 @@ type, making their usage much nicer. These are
 
 > `range()` for floating point numbers has been introduced in Catch2 2.11.0
 
+### Random generators
+```cpp
+#include <catch2/generators/catch_generators_random.hpp>
+```
+
```
horenmar, line 155: But I don't like it for lists, and I don't like adding new sections just to have the include paths in them.

docs/generators.md
```diff
@@ -145,6 +148,16 @@ type, making their usage much nicer. These are
...
+### Range generators
+```cpp
+#include <catch2/generators/catch_generators_range.hpp>
+```
```
horenmar, line 159: Because the text below now reads as part of the `range` generators.

docs/generators.md
horenmar: I think there is enough extra information to be mentioned about the random generators (e.g. their repeatability guarantees) that this subheading could be filled out to be meaningful.

However, I don't see this also being true for the `range` generator, so we will need a different format for inline markings of things in lists.

PR #2543, opened 2022-10-06
silence clang's `-Wcomma`

timblechmann:


horenmar:
Can you provide an example of when this warning triggers?

timblechmann:
my code looks like this:

```
#include <catch2/catch_template_test_macros.hpp>

namespace a::b::c {

struct foo {};

using tests = std::tuple< foo >;

TEMPLATE_LIST_TEST_CASE( "test", "", tests )
{}

} // namespace a::b::c
```

(it seems to be triggered by putting the `TEMPLATE_LIST_TEST_CASE` macro into the namespace)

horenmar:
I tried to quickly [reproduce this on godbolt](https://godbolt.org/z/rEW5da7Pf), but can't.

Given that this PR would silence `-Wcomma` for the entire test case code, including user's code, I do not want to merge it without a reproducer to see what the issue actually is, and that it is part of our code (as opposed to user code, whose warnings we should not touch.).

timblechmann:
i've tried to reproduce it again. it seems to be triggered by using a clang icecc toolchain with local proprocessing. so probably the namespace was a red herring.

fwiw, the offending comma is this one
```
Catch::NameAndTags{ "test" " - " + std::string("tests") + " - " + std::to_string(index), "" } ), index++)... }; } }
                                                                                               ^
```

horenmar:
Hi, sorry for not getting to this earlier. I thought the changes would silence `-Wcomma` in the test code as well, so I put it in my backlog to return to later.

Today I did, and turns out that suppression is already limited to the registration only.

PR #2544, opened 2022-10-07
Introduce CATCH_CONFIG_PREFIX_MESSAGES to only prefix a few logging related macros.

Rijom:
## Description
The macro CATCH_CONFIG_PREFIX_ALL allows to prefix *all* of catch2's macros with CATCH_. This commit adds some granularity by introducing a new macro: CATCH_CONFIG_PREFIX_MESSAGES. It only prefixes a few debug message macros.

Why? It's rather usual that logging frameworks already use INFO/WARN macros themselves. This allows a user to avoid clashes without having to use the prefix everywhere else.


Vertexwahn:
Would be nice if you can add it also here: [expand_template( expand_template(](https://github.com/catchorg/Catch2/blob/066cc51ce657a3d787ec228e8af22830a92723fc/BUILD.bazel#L3) to ensure the Bazel build does not break.

Rijom:
> Would be nice if you can add it also here: [expand_template( expand_template(](https://github.com/catchorg/Catch2/blob/066cc51ce657a3d787ec228e8af22830a92723fc/BUILD.bazel#L3) to ensure the Bazel build does not break.

Like so?

Rijom:
@horenmar Any comments regarding this PR?

src/catch2/catch_message.hpp
```diff
@@ -110,28 +110,32 @@ namespace Catch {
     Catch::getResultCapture().emplaceUnscopedMessage( Catch::MessageBuilder( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, Catch::ResultWas::Info ) << log )
 
 
-#if defined(CATCH_CONFIG_PREFIX_ALL) && !defined(CATCH_CONFIG_DISABLE)
+#if defined(CATCH_CONFIG_PREFIX_ALL) && !defined(CATCH_CONFIG_PREFIX_MESSAGES)
+#define CATCH_CONFIG_PREFIX_MESSAGES
+#endif
```
horenmar: Like other new `CATCH_CONFIG_*` options, this should go into its own header, `src/catch2/internal/catch_config_prefix_messages.hpp`.

horenmar (changes requested):
Sorry for taking so long to get around to this PR, but it happens to be in the sweet spot where there is no actual reason to reject it beyond me not seeing the value (the maintenance overhead and extra complexity is small enough), but also me not having a motivation to merge it (because I do not see the value & even if tiny, there are still costs).

In the end I decided to merge it, if one small change happens.

src/catch2/catch_message.hpp
Rijom: Done.

PR #2546, opened 2022-10-08
Mute the sign conversion warning with explicit cast

### LINKED ISSUE #2545: sign-conversion warning in catch_tostring.hpp
**Describe the bug**

Hello, the latest devel of Catch2 raises a warning on g++ 12.

`conversion to ‘std::size_t’ {aka ‘long unsigned int’} from ‘long int’ may change the sign of the result`

This warning was introduced a few days ago by PR #2540 

File `catch_tostring.hpp` line 46:

```cpp
inline std::size_t catch_strnlen(const char *str, std::size_t n) {
    auto ret = std::char_traits<char>::find(str, n, '\0');
    if (ret != nullptr) {
        return ret - str;  // <-- Here !
    }
    return n;
}
```

**Expected behavior**

Catch2 builds without warnings.

**Reproduction steps**

Build Catch2 with g++ .

**Platform information:**
g++ version 12.1.0 (Ubuntu 12.1.0-2ubuntu1~22.04) 

**Possible fixes**

- Change the return type of `catch_strnlen` to `std::ptrdiff_t`. This would prevent to return `n` without a cast to `std::ptrdiff_t`. 
- Explicitly cast the result to `std::size_t`.
```cpp
return static_cast<std::size_t>(ret - str);
```
This should be relatively safe as `str` should never be greater than `ret`. In my opinion this is the way to go.

@rkaminsk what do you think ?

Teskann:
## Description

Muted the sign conversion warning caused by implicit cast of returned result of `catch_strnlen()`.
Used `static_cast` to cast the type.

## GitHub Issues

Closes #2545

### DIFF
--- a/src/catch2/catch_tostring.hpp
+++ b/src/catch2/catch_tostring.hpp
@@ -43,7 +43,7 @@ namespace Catch {
         inline std::size_t catch_strnlen(const char *str, std::size_t n) {
             auto ret = std::char_traits<char>::find(str, n, '\0');
             if (ret != nullptr) {
-                return ret - str;
+                return static_cast<std::size_t>(ret - str);
             }
             return n;
         }


horenmar:
Thanks.

PR #2549, opened 2022-10-12
Link to SECTIONS doc at top of test fixtures page.

samcunliffe:
## Description
I came here looking for a way to use a fixture. But what I really wanted was better done in the SECTION macro. Feels like a link right at the top would've made it clearer faster.


horenmar:
Makes sense, thanks.

PR #2553, opened 2022-10-23
CatchMiscFunctions.cmake: Use PRIVATE for target_compile_options

autoantwort:
If PUBLIC is used INTERFACE_COMPILE_OPTIONS is populated. Definition of INTERFACE_COMPILE_OPTIONS: 'Targets may populate this property to publish the compile options required to compile against the headers for the target.' But this is not be true for -ffile-prefix-map.

### DIFF
--- a/CMake/CatchMiscFunctions.cmake
+++ b/CMake/CatchMiscFunctions.cmake
@@ -7,13 +7,13 @@
 # SPDX-License-Identifier: BSL-1.0
 
 include(CheckCXXCompilerFlag)
-function(add_cxx_flag_if_supported_to_targets scope flagname targets)
+function(add_cxx_flag_if_supported_to_targets flagname targets)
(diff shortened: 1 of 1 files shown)


CMake/CatchMiscFunctions.cmake
```diff
@@ -98,7 +98,7 @@ function(add_warnings_to_targets targets)
           # "-Wunused-member-function"
         )
         foreach(warning ${CHECKED_WARNING_FLAGS})
-            add_cxx_flag_if_supported_to_targets(${warning} "${targets}")
+            add_cxx_flag_if_supported_to_targets(PUBLIC ${warning} "${targets}")
```
horenmar: Actually these can be `PRIVATE` too.

horenmar:
Thanks

PR #2554, opened 2022-10-24
Use console reporter's totals summary for compact reporter

### LINKED ISSUE #878: Compact reporter does not handle [!shouldfail] properly

psalz:
This changes the compact reporter's summary of test run totals to use the same format as the console reporter. This means that while output is no longer on a single line (two instead), it now includes totals for `failedButOk` test cases and assertions, which were previously missing.

Fixes #878.

### DIFF
--- a/src/catch2/reporters/catch_reporter_helpers.cpp
+++ b/src/catch2/reporters/catch_reporter_helpers.cpp
@@ -235,4 +235,102 @@ namespace Catch {
         out << "\n\n" << std::flush;
     }
 
+    namespace {
+        class SummaryColumn {
+        public:
+            SummaryColumn( std::string suffix, Colour::Code colour ):
+                m_suffix( CATCH_MOVE( suffix ) ), m_colour( colour ) {}
+
+            SummaryColumn&& addRow( std::uint64_t count ) && {
+                std::string row = std::to_string(count);
+                auto const new_width = std::max( m_width, row.size() );
+                if ( new_width > m_width ) {
+                    for ( auto& oldRow : m_rows ) {
+                        oldRow.insert( 0, new_width - m_width, ' ' );
+                    }
+                } else {
+                    row.insert( 0, m_width - row.size(), ' ' );
+                }
+                m_width = new_width;
+                m_rows.push_back( row );
+                return std::move( *this );
+            }
+
+            std::string const& getSuffix() const { return m_suffix; }
+            Colour::Code getColour() const { return m_colour; }
+            std::string const& getRow( std::size_t index ) const {
+                return m_rows[index];
+            }
+
+        private:
+            std::string m_suffix;
+            Colour::Code m_colour;
+            std::size_t m_width = 0;
+            std::vector<std::string> m_rows;
+        };
+
+        void printSummaryRow( std::ostream& stream,
+                              ColourImpl& colour,
+                              StringRef label,
+                              std::vector<SummaryColumn> const& cols,
+                              std::size_t row ) {
+            for ( auto const& col : cols ) {
+                auto const& value = col.getRow( row );
+                auto const& suffix = col.getSuffix();
+                if ( suffix.empty() ) {
+                    stream << label << ": ";
+                    if ( value != "0" ) {
+                        stream << value;
+                    } else {
+                        stream << colour.guardColour( Colour::Warning )
+                               << "- none -";
+                    }
+                } else if ( value != "0" ) {
+                    stream << colour.guardColour( Colour::LightGrey ) << " | "
+                           << colour.guardColour( col.getColour() ) << value
+                           << ' ' << suffix;
+                }
+            }
+            stream << '\n';
+        }
+    } // namespace
+
+    void printTestRunTotals( std::ostream& stream,
+                             ColourImpl& streamColour,
+                             Totals const& totals ) {
+        if ( totals.testCases.total() == 0 ) {
+            stream << streamColour.guardColour( Colour::Warning )
+                   << "No tests ran\n";
+            return;
+        }
+
+        if ( totals.assertions.total() > 0 && totals.testCases.allPassed() ) {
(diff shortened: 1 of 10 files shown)


src/catch2/reporters/catch_reporter_helpers.cpp
```diff
@@ -235,4 +235,88 @@ namespace Catch {
...
+            Colour::Code colour;
+            std::vector<std::string> rows;
+        };
+
+        static void printSummaryRow( StringRef label,
+                                     std::vector<SummaryColumn> const& cols,
+                                     std::size_t row,
+                                     std::ostream& stream,
+                                     ColourImpl& colour ) {
+            for ( auto col : cols ) {
```
horenmar: Does this need to be a copy? From a quick glance I would say no, the code does not modify the `SummaryColumn` here.

(I know it is like this in the original code, but I am looking through it again here and hoo boy, it is a mess)

src/catch2/reporters/catch_reporter_helpers.cpp
```diff
@@ -235,4 +235,88 @@ namespace Catch {
         out << "\n\n" << std::flush;
     }
 
+    namespace {
+        struct SummaryColumn {
+            SummaryColumn( std::string _label, Colour::Code _colour ):
+                label( CATCH_MOVE( _label ) ), colour( _colour ) {}
+            SummaryColumn addRow( std::uint64_t count ) {
```
horenmar: This is another big mess. I don't feel like saddling you with fixing the implementation, but from looking at it, the signature does not make sense either.

It returns a copy, but not before modifying the current state of the instance, so something like

`SummaryColumn&& addRow(std::uint64_t count) &&`

makes more sense IMO, so we do not copy all the rows every time (which makes repeated calls to `addRow` have quadratic complexity), and the `&&` qualifier means that this won't be modifying any lvalues.

horenmar:
Looks good, with two drive-by fixes since you are already touching that code. Honestly you can ignore them if you want 🤷

psalz:
Sure, no problem. I did a refactoring pass as a fixup commit, will squash once approved.

---

Speaking of messes, I originally planned to do a larger refactoring of how `[!shouldfail]` and `[!mayfail]` are being handled, as there are a few things that have bothered me for a while (e.g. the way a *passing* `[!shouldfail]` test case is "reported"), but I gave up after a couple of hours because I found the whole logic to be very confusing. For example that a failed assertion in a `[!shouldfail]` is not `isOk` because it doesn't have `ResultDisposition::SuppressFail` (which only seems to be used for `CHECKED_IF/ELSE`), or [whatever is happening here](https://github.com/catchorg/Catch2/blob/8ce92d2c7288b6b3261caf1c016f8a779b6a8efc/src/catch2/internal/catch_run_context.cpp#L250-L253). If I ever find some time I would like to revisit this again, though.

src/catch2/reporters/catch_reporter_helpers.cpp
```diff
@@ -236,36 +236,50 @@ namespace Catch {
...
-                    while ( oldRow.size() > row.size() )
-                        row = ' ' + row;
+        class SummaryColumn {
+        public:
+            SummaryColumn( std::string suffix, Colour::Code colour ):
+                m_suffix( CATCH_MOVE( suffix ) ), m_colour( colour ) {}
+
+            SummaryColumn&& addRow( std::uint64_t count ) && {
+                std::string row = std::to_string(count);
+                auto const new_width = std::max( m_width, row.size() );
```
horenmar, line 246: you want `m_width` to be `size_t` as well.

psalz: Right, fixed.

src/catch2/reporters/catch_reporter_helpers.cpp
```diff
@@ -236,36 +236,50 @@ namespace Catch {
...
+                if ( new_width > m_width ) {
+                    for ( auto& oldRow : m_rows ) {
+                        oldRow = std::string( new_width - m_width, ' ' )
+                                     .append( oldRow );
```
horenmar: Tiny nitpick:
```
oldRow.insert(0, new_width - m_width, ' ');
```

is never worse and in _some_ (rather specific) circumstances actually better.

horenmar: Similar change applies to the other branch as well.

horenmar (approved):
lgtm

horenmar:
Yeah, shouldfail/mayfail is also a "fun" mess. It is what happens when the code is evolved over time for 10+ years 😃

src/catch2/reporters/catch_reporter_helpers.cpp
psalz: Fixed and squashed

PR #2557, opened 2022-10-28
Fix references to license file

ltoenning:
## Description

The license file was renamed with 6a502cc2f5fedd59b3495b58708f0d6d987ed9e1. This change was not reflected in the header license section. I excluded ``third_party/clara.hpp`` from this change (because it is third party).


horenmar:
Thanks.

PR #2563, opened 2022-11-10
Fix Bazel build

Vertexwahn:
Bazel build is currently broken, since new cmakedefines where introduced. This PR fixes the Bazel build. The PR only affects the Bazel build experience.


horenmar:
Thanks, adding a CI job should keep the build unbroken in the future.

PR #2564, opened 2022-11-10
Make ScopedMessage in INFO logging macro const

Rjblaauboer:
## Description

What: Made Catch::ScopedMessage in the INFO logging macro const because it is not modified. 

Why: If users have const correctness checks enabled then the current non-const message would be flagged forcing users to wrap the INFO macro themselves or forego using it altogether.

We encountered this while using [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/misc/const-correctness.html) with the `misc-const-correctness` check.


horenmar:
Sure, thanks.

PR #2565, opened 2022-11-12
meson: update minimum version

neheb:
include_directories of type string began being supported in meson 0.50.0. New warning with meson 0.64


horenmar:
Thanks

PR #2566, opened 2022-11-12
Make MatcherGenericBase copy constructor take const parameter.

mry792:
## Description
Fix the copy constructor signature of `MatcherGenericBase` to take a `const` parameter. While C++ allows copy constructors taking non-const parameters, this is very atypical and blocks making copies of const objects.


horenmar:
Thanks. AFAIK there is no reason for the use of plain ref there, just an oversight when the code was written.

PR #2568, opened 2022-11-14
Update Fibonacci example in README.md

ShawnZhong:
I'm new to Catch2 and found myself confused when I pasted the example on `README.md`. 

- The example in `README.md` won't compile because of name mismatches. 

- The implementation also does not match the test case (please let me know if that's intentional). 

- I read from [here](https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#special-tags) that `[!benchmark]` denotes a benchmark to hide the test case by default, but it's unclear at a glance how to run them in the resulting executable until reading [command-line.md](https://github.com/catchorg/Catch2/blob/devel/docs/command-line.md#specifying-which-tests-to-run)

This PR attempts to fix the issues above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,20 +47,20 @@ TEST_CASE( "Factorials are computed", "[factorial]" ) {
 #include <cstdint>
 
 uint64_t fibonacci(uint64_t number) {
-    return number < 2 ? 1 : fibonacci(number - 1) + fibonacci(number - 2);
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks. I went with different fix, to make the implementation the same as the one in the benchmarking tests.

PR #2570, opened 2022-11-18
add MessageMatches matcher

lbckmnn:
## Description

If you want to test whether a function throws an exception with a message, it is currently not possible to check the exception type and apply a matcher on the exception message in one run. You can check the **exact** message with `REQUIRE_THROWS_MATCHES` and `Message("test")` but you can not apply a matcher on the message.

If you want to apply a matcher on the exception message and check the type, the current options are:
1) check type and message in two runs:
```cpp
//REQUIRE_THROWS_WITH allows to apply a matcher on the message but doesn't allow checking the type
REQUIRE_THROWS_WITH(throwsException(),  StartsWith("Argument is invalid")); 
//REQUIRE_THROWS_AS allows checking the type, but not the message
REQUIRE_THROWS_AS(throwsException(), std::invalid_argument); 
```

2) use `REQUIRE_THROWS_MATCHES`:
This options only allows applying matchers on the exception itself, and not on the message so you have to write a custom matcher.

This PR adds a matcher that makes it possible to use `std::string` matchers like `StartsWith` or `ContainsSubstring`  on exception messages together with `REQUIRE_THROWS_MATCHES`.

Example:
```cpp
REQUIRE_THROWS_MATCHES(throwsException(),  std::runtime_error,  MessageMatches(StartsWith("Argument is invalid")));
```


horenmar:
FYI: I like the addition, but won't get to this until late next week.

extras/catch_amalgamated.cpp
```diff
@@ -6,7 +6,7 @@
 // SPDX-License-Identifier: BSL-1.0
 
 //  Catch v3.2.0
-//  Generated: 2022-11-16 19:30:16.114602
```
horenmar: Same as the other PR

src/catch2/matchers/catch_matchers_exception.hpp
```diff
@@ -29,6 +29,32 @@ class ExceptionMessageMatcher final : public MatcherBase<std::exception> {
 //! Creates a matcher that checks whether a std derived exception has the provided message
 ExceptionMessageMatcher Message(std::string const& message);
 
+template <typename StringMatcherType>
```
horenmar, line 32: Do you intend for this to take arbitrary matcher, rather than just the ones derived from `StringMatcherBase`?

lbckmnn: The regex matcher might be useful and it seems not to inherit from `StringMatcherBase`
Do we gain anything if we limit this to a subclass of `StringMatcherBase`?

horenmar: Ok, makes sense.

PR #2572, opened 2022-11-20
fix compiler warning in parseUint and catch only relevant exceptions

lbckmnn:
compiling catch2 with gcc 12.2.0 throws the warning:

`catch_parse_numbers.cpp:47:5: warning: control reaches end of non-void function [-Wreturn-type]`

This PR fixes the warning. 

I also noticed the wildcard `std::exception` catch. 
std::stoull throws only `std::invalid_argument` and `std::out_of_range` so i changed that too.
Otherwise exceptions like `std::bad_alloc` would be catched here.


extras/catch_amalgamated.cpp
```diff
@@ -6,7 +6,7 @@
 // SPDX-License-Identifier: BSL-1.0
 
 //  Catch v3.2.0
-//  Generated: 2022-11-16 19:30:16.114602
```
horenmar: [No touching amalgamated files in PRs](https://github.com/catchorg/Catch2/blob/devel/docs/contributing.md#using-github)

horenmar (changes requested):
I will merge the change, but I really don't understand how GCC thinks the control flow could reach end of the function without returning.

lbckmnn:
> I will merge the change, but I really don't understand how GCC thinks the control flow could reach end of the function without returning.

Yes the only way to reach the end of a function without return is when somehow something not inheriting `std::exception` gets thrown.

PR #2585, opened 2022-11-30
adds clang-tidy warning suppressions for `CHECK` and `TEST_CASE`

cjdb:
Both `CHECK` and `TEST_CASE` were giving warnings at the point of use. This commit makes it so that they do not (for now).

Checks disabled:

* `cppcoreguidelines-avoid-do-while` (for `CHECK`)
* `cppcoreguidelines-avoid-non-const-global-variables` (for `TEST_CASE`)

### DIFF
--- a/src/catch2/catch_test_macros.hpp
+++ b/src/catch2/catch_test_macros.hpp
@@ -137,7 +137,7 @@
   #define CHECK_THROWS_AS( expr, exceptionType ) INTERNAL_CATCH_THROWS_AS( "CHECK_THROWS_AS", exceptionType, Catch::ResultDisposition::ContinueOnFailure, expr )
   #define CHECK_NOTHROW( ... ) INTERNAL_CATCH_NO_THROW( "CHECK_NOTHROW", Catch::ResultDisposition::ContinueOnFailure, __VA_ARGS__ )
 
-  #define TEST_CASE( ... ) INTERNAL_CATCH_TESTCASE( __VA_ARGS__ )
+  #define TEST_CASE( ... ) INTERNAL_CATCH_TESTCASE( __VA_ARGS__ ) /* NOLINT(cppcoreguidelines-avoid-non-const-global-variables) */
   #define TEST_CASE_METHOD( className, ... ) INTERNAL_CATCH_TEST_CASE_METHOD( className, __VA_ARGS__ )
   #define METHOD_AS_TEST_CASE( method, ... ) INTERNAL_CATCH_METHOD_AS_TEST_CASE( method, __VA_ARGS__ )
   #define REGISTER_TEST_CASE( Function, ... ) INTERNAL_CATCH_REGISTER_TESTCASE( Function, __VA_ARGS__ )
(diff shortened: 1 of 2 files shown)


src/catch2/catch_test_macros.hpp
```diff
@@ -137,7 +137,7 @@
   #define CHECK_THROWS_AS( expr, exceptionType ) INTERNAL_CATCH_THROWS_AS( "CHECK_THROWS_AS", exceptionType, Catch::ResultDisposition::ContinueOnFailure, expr )
   #define CHECK_NOTHROW( ... ) INTERNAL_CATCH_NO_THROW( "CHECK_NOTHROW", Catch::ResultDisposition::ContinueOnFailure, __VA_ARGS__ )
 
-  #define TEST_CASE( ... ) INTERNAL_CATCH_TESTCASE( __VA_ARGS__ )
+  #define TEST_CASE( ... ) INTERNAL_CATCH_TESTCASE( __VA_ARGS__ ) /* NOLINT(cppcoreguidelines-avoid-non-const-global-variables) */
```
horenmar, line 140: 1) This is not enough, you either need to add this to `CATCH_TEST_CASE` as well, or place it into `INTERNAL_CATCH_TESTCASE`.
2) Why not actually make the variable const? https://github.com/catchorg/Catch2/commit/986c78016e696d29e09abe9b4b7c855bb4f871b5

horenmar: I merged a change that makes all possible registrars `const`. See if this fixes your issue.

src/catch2/internal/catch_test_macro_impl.hpp
```diff
@@ -47,7 +47,7 @@
 
 ///////////////////////////////////////////////////////////////////////////////
 #define INTERNAL_CATCH_TEST( macroName, resultDisposition, ... ) \
-    do { /* NOLINT(bugprone-infinite-loop) */ \
+    do { /* NOLINT(bugprone-infinite-loop, cppcoreguidelines-avoid-do-while) */ \
```
horenmar, line 50: I'll merge this, but I am honestly getting tired of clang-tidy complaining about perfectly fine code.

I am kinda-ok with it being unable to reason about the loop enough to not see that it is not infinite, but complaining about macro idiom that is older than I am, and is used to make macros more sanitary is bleh.

ChrisThrasher: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-do-while.html#options

I wanted to mention that clang-tidy has option to disable this check for macros so users encountering this problem can fix it in their own projects if they wish. I can't think of any reason someone would want to keep this warning for 1st party macros but would be upset that Catch2 uses this pattern. As you said it's a very well-established pattern.

killerbot242:
I was planning on creating the exact same pull request. 
I did notice this one was not applied (yet), could this please be done. I understand the argument to rather not do it, but please understand when user would like to follow this guideline on their own code, it is annoying that Catch2 gives rise to this warning, and as such either we need to patch it locally or we need to suppress it in the project/unit tests, which is then also suppressing in our own code.

horenmar:
I've given this some more thought:

1) I am willing to merge code changes to keep clang-tidy happy. I prefer code changes to suppressions, see e.g. the change to `const`ify the registration vars.
2) I am not willing to entertain `cppcoreguidelines-avoid-do-while`. Set your checker to ignore `do {} while` in macros.

PR #2586, opened 2022-11-30
Avoid usage of master when possible

sarroutbi:
Signed-off-by: Sergio Arroutbi <sarroutb@redhat.com>

## Description
References to "master" should be avoided when possible

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1319,7 +1319,7 @@ CHM_FILE               =
 HHC_LOCATION           =
 
 # The GENERATE_CHI flag controls if a separate .chi index file is generated
-# (YES) or that it should be included in the master .chm file (NO).
+# (YES) or that it should be included in the main .chm file (NO).
 # The default value is: NO.
 # This tag requires that the tag GENERATE_HTMLHELP is set to YES.
 
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -155,7 +155,7 @@ with expansion:
 
 
 ### Clang/G++ -- skipping leaf sections after an exception
-Some versions of `libc++` and `libstdc++` (or their runtimes) have a bug with `std::uncaught_exception()` getting stuck returning `true` after rethrow, even if there are no active exceptions. One such case is this snippet, which skipped the sections "a" and "b", when compiled against `libcxxrt` from master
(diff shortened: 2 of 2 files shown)


docs/limitations.md
```diff
@@ -155,7 +155,7 @@ with expansion:
 
 
 ### Clang/G++ -- skipping leaf sections after an exception
-Some versions of `libc++` and `libstdc++` (or their runtimes) have a bug with `std::uncaught_exception()` getting stuck returning `true` after rethrow, even if there are no active exceptions. One such case is this snippet, which skipped the sections "a" and "b", when compiled against `libcxxrt` from master
+Some versions of `libc++` and `libstdc++` (or their runtimes) have a bug with `std::uncaught_exception()` getting stuck returning `true` after rethrow, even if there are no active exceptions. One such case is this snippet, which skipped the sections "a" and "b", when compiled against `libcxxrt` from devel branch
```
horenmar: Assuming I found [the right repo](https://github.com/libcxxrt/libcxxrt), they still use `master` for their main branch. Changing the name is wrong here.

Doxyfile
```diff
@@ -1319,7 +1319,7 @@ CHM_FILE               =
 HHC_LOCATION           =
 
 # The GENERATE_CHI flag controls if a separate .chi index file is generated
-# (YES) or that it should be included in the master .chm file (NO).
+# (YES) or that it should be included in the primary .chm file (NO).
```
horenmar: These comments are autogenerated by doxygen. If the upstream has changed what they generate here, I am fine with an update, but otherwise I don't want to add unnecessary diff here.

sarroutbi: Indeed, if you generate a template file with latest version of doxygen, it generates here a reference with "in the main .chm file (NO)". I will change it

docs/limitations.md
sarroutbi: Sorry, I misunderstood it. I will revert it. Thanks for clarifying

horenmar: Feel free to use "from the master branch" if you think it makes it clearer though.

sarroutbi: > Feel free to use "from the master branch" if you think it makes it clearer though.

Done

horenmar:
Thanks

PR #2592, opened 2022-12-10
Fix missing include causing compiler error

bitter:
## Description
Fix missing include for configurations where exceptions are disabled.
```
catch_registry_hub.cpp:65:17: error: use of undeclared identifier 'CATCH_INTERNAL_ERROR'
                CATCH_INTERNAL_ERROR("Attempted to register active exception under CATCH_CONFIG_DISABLE_EXCEPTIONS!");
```


horenmar:
Thanks

PR #2593, opened 2022-12-12
Remove -H. argument from the 'Configure the full test build' command

ScottHutchinson:
The -H. argument seems to be a typo. To run the command, I have to remove that argument.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -87,7 +87,7 @@ $ ./scripts/generateSingleHeader.py
 ```
 2. Configure the full test build
 ```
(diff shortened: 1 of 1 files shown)


horenmar:
Are you using powershell?

ScottHutchinson:
> Are you using powershell?

Yes. Does that argument have a purpose?

horenmar:
`-H<folder-with-CML>` tells cmake to look for the CML.txt to run in specific folder. The reason it works without it is that by default, cmake looks in current working dir. And for some reason, powershell makes an absolute hash of that flag.

PR #2602, opened 2022-12-20
.clang-format: `SortIncludes: True` is incorrect

zhrkvl:
Also `'false'` instead of `false` gives warnings like "Required: boolean. Actual: string. "

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -4,7 +4,7 @@ Standard: c++14
 
 # Note that we cannot use IncludeIsMainRegex functionality, because it
 # does not support includes in angle brackets (<>)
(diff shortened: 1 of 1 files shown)


horenmar:
Hmmm, this was changed between clang-format 12 and 13. I am not sure which version to target, so I'll have to think about it for a bit.

horenmar:
Well I am using 12 now and in the foreseeable future, so thanks for the PR but I am keeping the current configuration file.

(Ideally clang-format would have a `ClangFormatVersion` field and either have backwards compatibility, or at least good error messages, but 🤷 )

PR #2604, opened 2022-12-27
Keep generated headers within project binary directory

ChrisThrasher:
## Description
This stops the `generated-includes/` directory from appearing in the build directories of projects that consume Catch2 via `add_subdirectory`.


horenmar:
Sure, that makes sense.

PR #2620, opened 2023-01-12
Export CMake package

fredericouellet-eaton:
## Description
Add package export to CMake


fredericouellet-eaton:
Cancel it; I found a bug

PR #2622, opened 2023-01-13
Document template macros are in spearate header

arashbm:
## Description

Type parametrised and signature parametrised test macros are now in their own header. The documentation should reflect that.


horenmar:
Thanks

PR #2627, opened 2023-01-17
Fix error: ISO C++ forbids in-class initialization of non-const static member 'm_started'

bblanchon:
This is a PR for Catch 1.x.
It fixes the following compilation error on GCC 4.4 and GCC 4.6:

```
error: ISO C++ forbids initialization of member ‘m_started’
error: making ‘m_started’ static
error: ISO C++ forbids in-class initialization of non-const static member ‘m_started’
```

### DIFF
--- a/include/internal/catch_fatal_condition.hpp
+++ b/include/internal/catch_fatal_condition.hpp
@@ -30,7 +30,7 @@ namespace Catch {
 
 namespace Catch {
     class FatalConditionHandler {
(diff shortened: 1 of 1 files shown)


horenmar:
Sure, I'll merge it. I am surprised anyone still uses v1 though.

bblanchon:
Thanks, @horenmar.
I'm still using v1 because it's compatible with C++98.

PR #2630, opened 2023-01-20
Export symbols for all compilers on Windows

ChrisThrasher:
## Description
Linking Catch2 v3.2.1 as a shared library when using Clang on Windows fails because symbols only get exported for MSVC (and MSVC-like compilers) but not for Clang. Removing this conditional means that Clang on Windows now gets this too. This should have no effect on compilers on non-Windows platforms. I believe we could move this out of the `if(BUILD_SHARED_LIBS)` block as well since it will just do nothing if you're building static libraries.

Here are two CI pipeline runs from SFML illustrating before and after applying this patch.

Before:
https://github.com/ChrisThrasher/SFML/actions/runs/3946193148

After
https://github.com/ChrisThrasher/SFML/actions/runs/3972002714

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -487,12 +487,10 @@ set(CATCH_WARNING_TARGETS ${CATCH_WARNING_TARGETS} PARENT_SCOPE)
 # so we want to check & warn users if they do this. However, we won't abort
 # the configuration step so that we don't have to also provide an override.
 if (BUILD_SHARED_LIBS)
-    if (MSVC)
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks.

> I believe we could move this out of the `if(BUILD_SHARED_LIBS)` block as well since it will just do nothing if you're building static libraries.

After reading the docs, I agree. But since the conditional still has to be there (the behaviour of static libraries compiled with default visibility and hidden visibility is different when they are linked into dynamic library later), I think it is fine to leave it the way it is now.

PR #2639, opened 2023-02-08
ObjectStorage: fix UB (strict aliasing violation)

### LINKED ISSUE #2638: Missing std::launder() calls
Catch2 and Catch3 contain `std::optional`-like code that uses a `char` array as storage for object created there with placement new. In C++17, this pattern may be undefined, because accessing the `T` through a cast pointer from `char[]` breaks strict aliasing rules.

Catch2's `Option` is not affected, since it stores the return value of placement new, circumventing the issue. But `ObjectStorage` has the issue.

I didn't look at Catch3 code, yet, but it seems the affected code is still present there, too.

marcmutz:
ObjectStorage is acting like a std::optional, manually managing the lifetime of an object placed into a char array buffer. C++17 changes the lifetime rules around such patterns, with the effect that this code now exhibits undefined behaviour, because &data is a pointer to a struct containing a char array, not a pointer to T, and it remains that way even after the static_cast.

The fix is to use std::launder(), which, for C++14 compilers, we hide behind a macro.

The similar Option type is not affected, since it has an additional T* member that stores the result of placement new, and therefore doesn't alias anything.

Fixes #2638.

### DIFF
--- a/include/internal/benchmark/catch_constructor.hpp
+++ b/include/internal/benchmark/catch_constructor.hpp
@@ -54,11 +54,11 @@ namespace Catch {
                 void destruct_on_exit(typename std::enable_if<!Destruct, U>::type* = 0) { }
 
                 T& stored_object() {
-                    return *static_cast<T*>(static_cast<void*>(&data));
+                    return *CATCH_LAUNDER(static_cast<T*>(static_cast<void*>(&data)));
                 }
 
                 T const& stored_object() const {
-                    return *static_cast<T*>(static_cast<void*>(&data));
+                    return *CATCH_LAUNDER(static_cast<T*>(static_cast<void*>(&data)));
                 }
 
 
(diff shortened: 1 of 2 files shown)


include/internal/benchmark/catch_constructor.hpp
```diff
@@ -54,11 +54,11 @@ namespace Catch {
                 void destruct_on_exit(typename std::enable_if<!Destruct, U>::type* = 0) { }
 
                 T& stored_object() {
-                    return *static_cast<T*>(static_cast<void*>(&data));
+                    return *CATCH_LAUNDER(static_cast<T*>(static_cast<void*>(&data)));
```
horenmar, line 57: Same as with the other PR.

marcmutz:
My resident committee member says that the only remaining use-case for launder() is as a de-virtualization barrier. C++20 made types with const and reference members transparently replaceable, and the only compiler that was optimizing on this, IBM xlc++, non-conformant. Even then, xlc++ only applied the optimization for static-storage-duration objects. So I guess we can do without.

PR #2640, opened 2023-02-08
ObjectStorage: fix UB (strict aliasing violation)

### LINKED ISSUE #2638: Missing std::launder() calls

marcmutz:
ObjectStorage is acting like a std::optional, manually managing the lifetime of an object placed into a char array buffer. C++17 changes the lifetime rules around such patterns, with the effect that this code now exhibits undefined behaviour, because &data is a pointer to a struct containing a char array, not a pointer to T, and it remains that way even after the static_cast.

The fix is to use std::launder(), which, for C++14 compilers, we hide behind a macro.

Incidentally, this should fix the -Wstrict-aliasing we've been suppressing there for real.

As a drive-by, replace the char array with a struct containing a char array (code from Catch2 2.x). This is safer with std::launder(), which is a bit weird with arrays (see reachability discussion in the std::launder() docs).

The similar Optional type is not affected, since it has an additional T* member that stores the result of placement new, and therefore doesn't alias anything, but in the interest of keeping the code similar, replace the char array with a struct containing the char array here, too.

Fixes #2638.

### DIFF
--- a/src/catch2/benchmark/catch_constructor.hpp
+++ b/src/catch2/benchmark/catch_constructor.hpp
@@ -24,12 +24,12 @@ namespace Catch {
 
                 ObjectStorage(const ObjectStorage& other)
                 {
-                    new(&data) T(other.stored_object());
+                    new(&storage) T(other.stored_object());
                 }
 
                 ObjectStorage(ObjectStorage&& other)
                 {
-                    new(data) T(CATCH_MOVE(other.stored_object()));
+                    new(&storage) T(CATCH_MOVE(other.stored_object()));
                 }
 
                 ~ObjectStorage() { destruct_on_exit<T>(); }
@@ -37,7 +37,7 @@ namespace Catch {
                 template <typename... Args>
                 void construct(Args&&... args)
                 {
-                    new (data) T(CATCH_FORWARD(args)...);
+                    new (&storage) T(CATCH_FORWARD(args)...);
                 }
 
                 template <bool AllowManualDestruction = !Destruct>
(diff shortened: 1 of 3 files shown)


src/catch2/benchmark/catch_constructor.hpp
```diff
@@ -54,20 +54,16 @@ namespace Catch {
                 template <typename U>
                 void destruct_on_exit(std::enable_if_t<!Destruct, U>* = nullptr) { }
 
-#if defined( __GNUC__ ) && __GNUC__ <= 6
-#    pragma GCC diagnostic push
-#    pragma GCC diagnostic ignored "-Wstrict-aliasing"
-#endif
-                T& stored_object() { return *reinterpret_cast<T*>( data ); }
+                T& stored_object() {
+                    return *CATCH_LAUNDER(reinterpret_cast<T*>( &storage ));
```
horenmar, line 58: This is currently missing include to use `CATCH_LAUNDER`.

dimztimz:
This class, `ObjectStorage` can be refactored to be union with just a single member. One should start by replacing `class ObjectStorage` with `union ObjectStorage` and add the needed members functions. Everything gets simpler then, no casts, no launder, no alignas. Only placement new should be used. Probably same goes for the class `Optional`.

IIRC the point of launder is if you construct object of one type with placement new, destruct it and then construct object of different type within the same buffer. This is not such case.

marcmutz:
> IIRC the point of launder is if you construct object of one type with placement new, destruct it and then construct object of different type within the same buffer. This is not such case.

This is correct in C++20. In C++17, you still need the launder() even in a union, in case the type held contains const or reference members.

marcmutz:
My resident committee member says that the only remaining use-case for launder() is as a de-virtualization barrier. C++20 made types with const and reference members transparently replaceable, and the only compiler that was optimizing on this, IBM xlc++, non-conformant. Even then, xlc++ only applied the optimization for static-storage-duration objects. So I guess we can do without.

PR #2643, opened 2023-02-10
cmake-integration.md: Use "tests" as test target name in all examples.

kisielk:
## Description
A small change just to make copying the examples a little easier / quicker when getting started with a project.


horenmar:
Sure, thanks.

PR #2651, opened 2023-03-08
Add support for the IAR compiler

jushar:
## Description

This pull request fixes compilation with the (proprietary) IAR compiler (tested with IAR EW for ARM 9.10).

Without significant changes to the build environment, it is not possible to run the C-SPY emulator against the self tests though.
That's why I did a couple of (succeeding) manual tests for some basic examples that ran in the emulator.


horenmar:
That looks reasonable, thanks

Issue #2656, opened 2023-03-11
Sharded tests CMake integration not properly installed and detected


viravera:
**Describe the bug**
In trying to upgrade to sharded tests, I found two problems:
   1. CatchShardTests.cmake and CatchShardTestsImpl.cmake are not installed with the other extras during make install.
   2. The CatchShardTestsImpl.cmake file location is incorrectly calculated during user project build step.

**Expected behavior**
   1. make install should place files in correct location.
   2. CatchShardTestsImpl.cmake should be findable when building projects using Catch's sharding support. 

**Reproduction steps**
   1. Run make install, note that CatchShardTests.cmake and CatchShardTestsImpl.cmake do not get installed like the other extras.
   2. Build any project using the installed CatchShardTests.cmake. Currently, unless CatchShardTestsImpl.cmake is in the same location as the calling script, they do not work.

**Platform information:**

 - OS: **Ubuntu 20.04**
 - Compiler+version: **GCC v12.2**
 - Catch version: **v3.3.2**
 - CMake version: **v3.22.2**

**Additional context**
   1. The problem with the catch's install can be fixed by simply adding extras/CatchShardTests.cmake and extras/CatchShardTestsImpl.cmake to the list of installed extra files in /CMakeLists.txt at line 144.
   2. The problem with the incorrect calculation of CatchShardTestsImpl,cmake appears to be due to the fact that CMAKE_CURRENT_LIST_DIR is not the same at configuration time as at the time of the function call, similar to the issue noted here: https://stackoverflow.com/questions/12802377/in-cmake-how-can-i-find-the-directory-of-an-included-file.
   To fix it, I altered the CatchShardTests.cmake file to match the pattern used in Catch.cmake to find its dependency CatchAddTests.cmake.
   
   
   I will submit a pull request with the changes that fixed this issue for me momentarily.

PR #2657, opened 2023-03-11
Fixed extras installation and shard impl script detection

viravera:
## Description
What: Fixed CMake sharding installation and associated test script location calculation

Why: I need sharding support to work from a clean catch build for my CI.

## GitHub Issues
https://github.com/catchorg/Catch2/issues/2656


horenmar:
Thanks for the fix.

I originally didn't add these to be installed because they were just an experiment, not meant as replacement for the other scripts (and I don't like maintaining CMake scripts 😃 ). Good to hear that they are useful to you though.

PR #2658, opened 2023-03-11
Add more open source libraries using Catch

vadz:
I was looking at this document in the v2 branch and saw this part of it:

> This effort largely relies on the maintainers of the projects themselves updating this page and submitting a PR 

and so decided to add two more of the projects I maintain to it (another one was already there), but while making this PR I realized that this part is not there any more, so I don't know if this still applies. If it doesn't, please feel free to just ignore and close this one.

### DIFF
--- a/docs/opensource-users.md
+++ b/docs/opensource-users.md
@@ -110,6 +110,9 @@ A header-only TOML parser and serializer for modern C++.
 ### [Trompeloeil](https://github.com/rollbear/trompeloeil)
 A thread-safe header-only mocking framework for C++14.
 
+### [wxWidgets](https://www.wxwidgets.org/)
+Cross-Platform C++ GUI Library.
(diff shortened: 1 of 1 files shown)


horenmar:
The docs currently say 

> _Please only add projects here if you are their maintainer, or have the
maintainer's explicit consent._

which is close enough I think.

The underlying reason for this is to list projects that are happy-enough with using Catch2 that they want to put in a bit of effort to say that they use it.

------------

Also thanks.

vadz:
> The underlying reason for this is to list projects that are happy-enough with using Catch2

Just to make it completely explicit, I'm very happy with Catch and use it in all my projects (not only open source ones) since I first discovered it many years ago (2016?).

> Also thanks.

And thank you very much for maintaining it!

PR #2661, opened 2023-03-22
Fix template syntax

mateuszchudyk:
This PR fixes template syntax of Clara::Parser function.


horenmar:
Two issue here
1) `catch_amalgamated.*` are not supposed to be changed outside of releases
2) The change is not valid. https://godbolt.org/z/hf3ToGEKr

PR #2669, opened 2023-04-04
Fix reporter detection in catch_discover_tests

### LINKED ISSUE #2668: catch_discover_tests: xml reporter vs. XML reporter (lower/upper case mismatch)
**Describe the bug**
According to the documentation, `--reporter xml` with lower case "xml" should work, but using `catch_discover_tests` with `REPORTER xml` leads to an error  `"xml" is not a valid reporter!` in CatchAddTests.cmake:85 (as of Catch 3.3.2). On the other hand, `REPORTER XML` with upper case "XML" works.

Listing the reporters produces 
```
   Available reporters:
    Automake:   Reports test results in the format of Automake .trs files
    compact:    Reports test results on a single line, suitable for IDEs
    console:    Reports test results as plain lines of text
    JUnit:      Reports test results in an XML format that looks like Ant's
                  junitreport target
    SonarQube:  Reports test results in the Generic Test Data SonarQube XML
                  format
    TAP:        Reports test results in TAP format, suitable for test harnesses
    TeamCity:   Reports test results as TeamCity service messages
    XML:        Reports test results as an XML document
```
(Note the upper case XML)

**Expected behavior**
Using the reporter feature in `catch_discover_tests` should be consistent with the documented command line option or case-insensitive.
- See https://github.com/catchorg/Catch2/blob/devel/docs/command-line.md#choosing-a-reporter-to-use where the example uses lower case "xml"
- See https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md#automatic-test-registration where it explicitly mentioned that the argument is passed to the test runner as `--reporter reporter`

This broke some time between 2.13.9 and 3.3.2 (see https://github.com/cpp-best-practices/cmake_conan_boilerplate_template/blob/main/test/CMakeLists.txt where it still works)

**Reproduction steps**
CMake section:
```
find_package(Catch2 CONFIG REQUIRED)

include(Catch)

include(test_sources.cmake)
add_executable(tests tests.cpp)
target_link_libraries(tests PUBLIC Catch2::Catch2WithMain)

# automatically discover tests that are defined in catch based test files you can modify the unittests. Set TEST_PREFIX
# to whatever you want, or use different for different binaries
catch_discover_tests(
  tests
  REPORTER  xml
  OUTPUT_SUFFIX  .xml)
```

fails at compilation time.

Also note that the method used in CatchAddTests.cmake:85 is very crude, e.g. I could choose a reporter "suitable" without a failing build since that word happens to appear in the --list-reporters output.

arnemertz:
Closes #2668
Instead of grepping for the reporter, use it in the process and check for the error code

## Description
Inside the `catch_discover_tests` CMake script, instead of searching for the supposed reporter in the verbose (!) output of --list-reporters, actually use it as the actual reporter and check for failures when doing so. This will avoid failures when a correct reporter is not spelled in the correct case (reporters are case insensitive), and it avoids success when the supposed reporter is only a fill word in one of the reporter descriptions.

## GitHub Issues
- #2668 fixed

### DIFF
--- a/extras/Catch.cmake
+++ b/extras/Catch.cmake
@@ -35,7 +35,7 @@ same as the Catch name; see also ``TEST_PREFIX`` and ``TEST_SUFFIX``.
                          [TEST_LIST var]
                          [REPORTER reporter]
                          [OUTPUT_DIR dir]
-                         [OUTPUT_PREFIX prefix}
+                         [OUTPUT_PREFIX prefix]
                          [OUTPUT_SUFFIX suffix]
     )
 
--- a/extras/CatchAddTests.cmake
+++ b/extras/CatchAddTests.cmake
@@ -66,30 +66,30 @@ endif()
 
 string(REPLACE "\n" ";" output "${output}")
 
-# Run test executable to get list of available reporters
-execute_process(
-  COMMAND ${TEST_EXECUTOR} "${TEST_EXECUTABLE}" ${spec} --list-reporters
-  OUTPUT_VARIABLE reporters_output
-  RESULT_VARIABLE reporters_result
-  WORKING_DIRECTORY "${TEST_WORKING_DIR}"
-)
-if(NOT ${reporters_result} EQUAL 0)
-  message(FATAL_ERROR
-    "Error running test executable '${TEST_EXECUTABLE}':\n"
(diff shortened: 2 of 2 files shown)


horenmar:
Thanks.

PR #2670, opened 2023-04-11
cmake: Add new DISCOVERY_MODE option to catch_discover_tests

### LINKED ISSUE #2493: Add DISCOVERY_MODE to catch_discover_tests
catch_discover_tests is modelled after CMake's gtest_discover_tests, but it's not up to date.
Since CMake 3.18 gtest_discover_tests has supported a DISCOVERY_MODE option, which when set to PRE_TEST allows for the test discovery to be delayed until just prior to test execution. This is useful when cross-compiling, since the build machine may be unable to run the test executable.
Adding DISCOVERY_MODE to catch_discover_tests would be equally useful.

hkaelber:
## Description

Provide `DISCOVERY_MODE` option to `catch_discover_tests` to allow for `PRE_TEST` discovery mode.

Cf. details in commit msg-es.

This change follows closely the corresponding functionality / changes in the GoogleTest counterpart to maintain the convenient similarity to `gtest_discover_tests()` and friends.

## GitHub Issues

Closes #2493

### DIFF
--- a/docs/cmake-integration.md
+++ b/docs/cmake-integration.md
@@ -126,6 +126,7 @@ catch_discover_tests(target
                      [OUTPUT_DIR dir]
                      [OUTPUT_PREFIX prefix]
                      [OUTPUT_SUFFIX suffix]
+                     [DISCOVERY_MODE <POST_BUILD|PRE_TEST>]
 )
 ```
 
@@ -198,6 +199,16 @@ If specified, `suffix` is added to each output file name, like so
 `--out dir/<test_name>suffix`. This can be used to add a file extension to
 the output file name e.g. ".xml".
 
+* `DISCOVERY_MODE mode`
+
+If specified allows control over when test discovery is performed.
+For a value of `POST_BUILD` (default) test discovery is performed at build time.
+For a a value of `PRE_TEST` test discovery is delayed until just prior to test
+execution (useful e.g. in cross-compilation environments).
+``DISCOVERY_MODE`` defaults to the value of the
+``CMAKE_CATCH_DISCOVER_TESTS_DISCOVERY_MODE`` variable if it is not passed when
+calling ``catch_discover_tests``. This provides a mechanism for globally
+selecting a preferred test discovery behavior.
 
 ### `ParseAndAddCatchTests.cmake`
 
--- a/extras/Catch.cmake
+++ b/extras/Catch.cmake
@@ -37,6 +37,7 @@ same as the Catch name; see also ``TEST_PREFIX`` and ``TEST_SUFFIX``.
                          [OUTPUT_DIR dir]
                          [OUTPUT_PREFIX prefix]
                          [OUTPUT_SUFFIX suffix]
+                         [DISCOVERY_MODE <POST_BUILD|PRE_TEST>]
     )
 
   ``catch_discover_tests`` sets up a post-build command on the test executable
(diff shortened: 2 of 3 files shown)


docs/cmake-integration.md
```diff
@@ -126,6 +126,7 @@ catch_discover_tests(target
                      [OUTPUT_DIR dir]
                      [OUTPUT_PREFIX prefix]
                      [OUTPUT_SUFFIX suffix]
+                     [DISCOVERY_MODE mode]
```
horenmar: I think explicitly specifying the two options is better (and matches the CMake docs for the GTest analog).
```suggestion
                     [DISCOVERY_MODE <POST_BUILD|PRE_TEST>]
```

extras/Catch.cmake
```diff
@@ -159,39 +178,82 @@ function(catch_discover_tests TARGET)
...
+      "  add_test(${TARGET}_NOT_BUILT ${TARGET}_NOT_BUILT)"                        "\n"
+      "endif()"                                                                    "\n"
+    )
+
+    file(GENERATE OUTPUT "${ctest_include_file}" CONTENT "${ctest_include_content}")
+    message(STATUS "XXX writing to ${ctest_include_file}")
+    file(WRITE "${ctest_include_file}" "XXX")
+
+  else()
+    message(FATAL_ERROR "Unknown DISCOVERY_MODE: ${_DISCOVERY_MODE}")
```
horenmar: I think it would read better to have the check right after the default/fallback logic.

extras/Catch.cmake
```diff
@@ -159,39 +178,82 @@ function(catch_discover_tests TARGET)
...
+      "    )"                                                                      "\n"
+      "  endif()"                                                                  "\n"
+      "  include(\"${ctest_tests_file}\")"                                         "\n"
+      "else()"                                                                     "\n"
+      "  add_test(${TARGET}_NOT_BUILT ${TARGET}_NOT_BUILT)"                        "\n"
+      "endif()"                                                                    "\n"
+    )
+
+    file(GENERATE OUTPUT "${ctest_include_file}" CONTENT "${ctest_include_content}")
+    message(STATUS "XXX writing to ${ctest_include_file}")
```
horenmar: This seems to be a leftover from debugging? (And the line below this)

docs/cmake-integration.md
hkaelber: Agree, changed here and in the overview in `Catch.cmake`

extras/Catch.cmake
hkaelber: Oups, indeed, removed

extras/Catch.cmake
hkaelber: Agree, changed

hkaelber:
Sorry, forgot that you prefer fixup commits in your review process. But shoudl not be a big deal in this PR, pls cf. incremental diff [here](https://github.com/catchorg/Catch2/compare/3211facad2ac33d02482b6cf8fb46af688b44530..d03fafead825d15da1e0f52ead0e1343e8438a88)

horenmar:
LGTM, thanks.

riandrake:
Hi, sorry if this is the wrong place. When trying this, I get an error in CMake:

```
CMake Error in CMakeLists.txt:
  Evaluation file to be written multiple times with different content.  This
  is generally caused by the content evaluating the configuration type,
  language, or location of object files:
```

Attached is my usage:

2023 08 22 @ 04 51 48PM@2x

I'm using CLion with Ninja Multi-Config.

Are there limitations I'm not aware of?

hkaelber:
> Hi, sorry if this is the wrong place. When trying this, I get an error in CMake:
> 
> ```
> CMake Error in CMakeLists.txt:
>   Evaluation file to be written multiple times with different content.  This
>   is generally caused by the content evaluating the configuration type,
>   language, or location of object files:
> ```
> 
> Attached is my usage:
> 
> 2023 08 22 @ 04 51 48PM@2x
> I'm using CLion with Ninja Multi-Config.
> 
> Are there limitations I'm not aware of?

Hi @riandrake, thx for spotting this.

Indeed currently multi-config generators are not supported with `PRE_TEST` discovery because I ignored them when porting this because -- of lazyness, as we don't have support for Ninja Multi-Config in our primay build-env and I could not easily test it. :-)

Could you maybe test https://github.com/catchorg/Catch2/pull/2739? Works for me so far, though I still miss testing all combinations. Thx

riandrake:
@hkaelber this worked, thanks! I have a question though.

I noted that it didn't collect the tests until `PRE_TEST` time as expected, but this happened only once. My hope was that I could change the `TEST_SPEC` variable between runs and have it collect the tests again for targeted testing.

Am I misunderstanding how `PRE_TEST` works? If so, could I feature request this additional functionality somewhere?

hkaelber:
> @hkaelber this worked, thanks! I have a question though.
> 
> I noted that it didn't collect the tests until `PRE_TEST` time as expected, but this happened only once. My hope was that I could change the `TEST_SPEC` variable between runs and have it collect the tests again for targeted testing.
> 
> Am I misunderstanding how `PRE_TEST` works? If so, could I feature request this additional functionality somewhere?

Not sure I understood your point, @riandrake.

I've never used multi-config so far, but to my understanding

1. at configure/cmake time the config-specifc test includes are generated (containing the postponed discovery command) together with the main include file, that simply pulls in the config-specific test-include
2. at every ctest invocation (per concrete config, Debug, Release, etc) the actual config-specific test file should be generated and executed

Correct? (Though I'm 100% sure how ctest is supposed to be executed in a multi-config env.)
That means that the `TEST_SPEC` is already fixed in 1. for all configs.

riandrake:
Understood, thank you.

PR #2675, opened 2023-04-19
Improve unreachable-code warnings

hkaelber:
## Description

Enable CI to report -Wunreachable-code-aggressive warnings in clang builds which covers all,  -Wunreachable-code, -code-break, -code-return.

Adapt tests.
Fix one instance of unreachable-code-return, regression from d0f70fdfd.


horenmar:
Thanks

Issue #2674, opened 2023-04-19
CMake Integration does not support test names containing semicolons and backslashes


robinchrist:
**Describe the bug**
One of our working students has created several test cases for the parser of a custom DSL which often contains `;` (semicolon) and Windows-style file paths in expressions and has named them after the expression that is being tested if short enough, e.g.
```
TEST_CASE("@Script[C:\\EPM1A]=x;\"SCALA_ZERO:\"", "[script regressions]")
```

This leads to the creation of a lot of wrong test cases, potentially leading to false positives:
```
The following tests FAILED:
         13 - @Script[C:\EPM1A]=x (Failed)
         14 - "SCALA_ZERO:" (Failed)
```

This can also be seen in the corresponding `tests-<hash>-cmake`:
```
add_test( [==[@Script[C:\EPM1A]=x]==] /censored/path [==[@Script\[C:\EPM1A\]=x]==]  )
set_tests_properties( [==[@Script[C:\EPM1A]=x]==] PROPERTIES WORKING_DIRECTORY /censored/path)
add_test( [==["SCALA_ZERO:"]==] /censored/path [==["SCALA_ZERO:"]==]  )
set_tests_properties( [==["SCALA_ZERO:"]==] PROPERTIES WORKING_DIRECTORY /censored/path)
```

By adding some debug logging to `CatchAddTests.cmake`, we can see that it's a bug in the `foreach` logic:

Debug log statements:
```
message("Line: ${test}")
message("Escaped Line: ${test_name}")
```

```
Line: @Script[C:\EPM1A]=x
Escaped Line: @Script\[C:\EPM1A\]=x
Line: "SCALA_ZERO:"
Escaped Line: "SCALA_ZERO:"
```

I believe this is due to https://github.com/catchorg/Catch2/blob/9a2a4eadc0e50e0534278010624eac15991413a7/extras/CatchAddTests.cmake#L67
```
string(REPLACE "\n" ";" output "${output}")
```
Obviously, if the test case name already contains `;`, the test case will be split and it won't work.

```
string(REPLACE ";" "\;" output "${output}")
string(REPLACE "\n" ";" output "${output}")
```

seems to fix this.

One thing we can also see in the resulting output file (one we have applied the fix above):
```
add_test( [==[@Script[C:\EPM1A]=x;"SCALA_ZERO:"]==] /censored/path [==[@Script\[C:\EPM1A\]=x;"SCALA_ZERO:"]==]  )
```
[...]

PR #2676, opened 2023-04-19
WIP: Fix handling of semicolon and backslash characters in CMake test discovery

robinchrist:
## Description
This PR fixes the handling of semicolon and backslash characters in test names in the CMake test discovery 

## GitHub Issues
https://github.com/catchorg/Catch2/issues/2674

## TODO:
How can we test this? I believe introducing a new "category" of tests that actually makes use of `catch_discover_tests` would be the best way?

### DIFF
--- a/extras/CatchAddTests.cmake
+++ b/extras/CatchAddTests.cmake
@@ -74,6 +74,10 @@ function(catch_discover_tests_impl)
     )
   endif()
 
+  # Make sure to escape ; (semicolons) in test names first, because
+  # that'd break the foreach loop for "Parse output" later and create
+  # wrongly splitted and thus failing test cases (false positives)
+  string(REPLACE ";" "\;" output "${output}")
   string(REPLACE "\n" ";" output "${output}")
 
   # Prepare reporter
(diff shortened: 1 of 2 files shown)


horenmar:
> How can we test this? I believe introducing a new "category" of tests that actually makes use of catch_discover_tests would be the best way

Yeah, so far these CMake scripts are untested, because it was always more work than someone was willing to do to get them tested. Especially if the various options are to be covered as well. Even limiting ourselves to just the test parsing will be kinda annoying.

One of the issues is that it will need to be checked in a separately compiled CMake project (to avoid the new tests spamming the current CTest tests), and the check will have to be done by a separate (Python) script. We should check that the various problematic test names are both registered & invoked properly -- first thing I can think of is to have each `TEST_CASE` write out a specific string and then checking that the output from `ctest --verbose` contains all the outputs it should.

robinchrist:
> first thing I can think of is to have each `TEST_CASE` write out a specific string and then checking that the output from `ctest --verbose` contains all the outputs it should.

Hmm... Is that really necessary?
Given that test cases won't run to begin with if something about the name handling is broken, wouldn't it be easier to just write a couple of test cases with `SUCCEED()`, run `ctest` and check that none of them fail?

As I've mentioned in the issue, there were a lot of false positives (failing test cases) before I fixed the bug.
This worked due to the way `ctest` runs the test: The binary to be tested is invoked for every test case with the `<test name>` argument - So if the test name wasn't handled properly, the test binary will just not run any test (`No tests ran`) and `ctest` will count the test as failed.

horenmar:
> Hmm... Is that really necessary?

Technically no, but that is where my testing paranoia has gotten to over time 😃 See e.g. the various reporter tests that fail if the test-specific reporter does not report as used, so that there is a loud error if the test reporter is not selected for whatever reason.

One example:
```
set_tests_properties(
    Reporters::CapturedStdOutInEvents
  PROPERTIES
    PASS_REGULAR_EXPRESSION "X27 - TestReporter constructed"
    FAIL_REGULAR_EXPRESSION "X27 ERROR"
)
```

-----------------------


My issue with plain `SUCCEED` & check that all tests fail is simple: what if a test fails to register what so ever?

This issue can be solved by counting the passed tests and checking that number against an expected number of tests.

Can this fail? Yes, if the script breaks in a way that causes test spec to match multiple tests, e.g. bad escaping for `*` we would not know, because CTest would report the right number of tests.

...

I am honestly willing to merge the changes as they are right now, so I will obviously also merge the PR if it contains a weaker level of checking than I would like, but I would prefer more strict checks, unless they eventually prove to be too annoying to be maintained.

robinchrist:
Maybe we can split the work?
I'm not familiar enough with the project structure to create the infrastructure for tests as you described them, but I could create test cases once the infrastructure is in place?

horenmar:
Sorry for taking so long, but I finally had enough time to work on this. There is now a test scaffolding for the `catch_discover_tests` functionality, and tomorrow I'll rebase your PR on current devel and merge it.

PR #2679, opened 2023-04-25
Fix spelling

Vertexwahn:
Fix spelling


horenmar (changes requested):
Thanks, but not changes to `catch_amalgamated.*` should happen outside of releases.

Vertexwahn:
@horenmar Removed changes from `catch_amalgamated.*` files

horenmar (approved):

PR #2683, opened 2023-05-09
Update CatchConfigOptions.cmake

YarikTH:
## Description

Fix CMake warning

```
CMake Warning (dev) at catch2-src/CMake/CatchConfigOptions.cmake:71 (set):
  uninitialized variable 'BUILD_SHARED_LIBS'
```

CMake's --warn-uninitialized is very useful feature to prevent some silly problems like hard to detect typos in cmake variables. Unfortunately cmake doesn't have any way to locally disable this warning for 3rd-party code, so only way to deal with it - ignore or actually fix it.

### DIFF
--- a/CMake/CatchConfigOptions.cmake
+++ b/CMake/CatchConfigOptions.cmake
@@ -68,7 +68,11 @@ set(_OtherConfigOptions
 foreach(OptionName ${_OtherConfigOptions})
   AddConfigOption(${OptionName})
 endforeach()
-set(CATCH_CONFIG_SHARED_LIBRARY ${BUILD_SHARED_LIBS})
+if(DEFINED BUILD_SHARED_LIBS)
+    set(CATCH_CONFIG_SHARED_LIBRARY ${BUILD_SHARED_LIBS})
+else()
+    set(CATCH_CONFIG_SHARED_LIBRARY "")
+endif()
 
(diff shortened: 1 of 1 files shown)


horenmar:
Honestly I don't like the change, but it seems to be a reasonable way of handling CMake oddities.

ChrisThrasher:
Looks like `CATCH_CONFIG_SHARED_LIBRARY` ultimately exists so that shared libraries work on Windows. If that's the goal I might recommend using CMake's [`GenerateExportHeader`](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) module. The way it works lets you remove some preprocessor code and it happens to avoid this particular issue with using an uninitialized CMake variable. 

https://github.com/ChrisThrasher/argon/blob/0571868a5709d12259b098f485356ac625de56d5/CMakeLists.txt#L25-L27

```cmake
if(NOT BUILD_SHARED_LIBS)
    target_compile_definitions(argon PUBLIC ARGON_STATIC_DEFINE)
endif()
```

If `BUILD_SHARED_LIBS` is true then you add a preprocessor symbol but otherwise you omit that symbol. At no point do you have to read something that is potentially uninitialized or guard against that possibility.

YarikTH:
It looks silly, but yes, actually CMake don't warn you when uninitialized variables participating in conditions, only if they are explicitly dereferenced 😕

YarikTH:
Other project faced problems with incorrect behavior because of typo in option name. There is a discussion about ways to prevent it and cmake's behavior.
https://github.com/open-quantum-safe/liboqs/issues/865

PR #2698, opened 2023-06-01
Bazel support: Update skylib

Vertexwahn:
This PR only affects the Bazel build experience:

- Update of skylib


horenmar:
Ok, sure.

PR #2707, opened 2023-06-22
Add -Wdouble-promotion to common warnings

teksturi:
This has not huge benefits in Catch2 codebase, but it is still easy to take in use and it will benefit those which build catch2 with project default compiler options.


horenmar:
I am probably willing to merge it, but you need to 

* update approval tests
* fix all instances of the warnings

PR #2714, opened 2023-07-01
Fix source comment typos in `extras/catch_amalgamated.{cpp|hpp}`

luzpaz:
## Description

Fix source comment typos in `extras/catch_amalgamated.{cpp|hpp}`  
Found via `codespell -q 3 -L dependees,hel,nd,seh`

## GitHub Issues
n/a

### DIFF
--- a/extras/catch_amalgamated.cpp
+++ b/extras/catch_amalgamated.cpp
@@ -596,7 +596,7 @@ namespace Catch {
             elem = trim(elem);
         }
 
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks, but no edits to the amalgamated files should happen outside releases. If the typos are fixed in the original source files, they will be fixed then.

luzpaz:
Sorry, my bad.

PR #2717, opened 2023-07-17
Hard wrap tutorial doc for easier raw Markdown reading

david-alvarez-rosa:
Hard wrap text on tutorial doc to 72 column limit, to make it easier to be read the raw Mardown from text editor.

Also, make consistent the usage of two spaces after period.


SagarJadhav30 (commented):
intresting

PR #2726, opened 2023-07-31
Don't define tryTranslators function if exception are disabled

VitaliiTrubchaninovUbisoft:
"-Wunused-function -Wall" produces if this function is defined when exceptions are disabled

### DIFF
--- a/src/catch2/internal/catch_exception_translator_registry.cpp
+++ b/src/catch2/internal/catch_exception_translator_registry.cpp
@@ -15,6 +15,7 @@
 
 namespace Catch {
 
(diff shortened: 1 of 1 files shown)


VitaliiTrubchaninovUbisoft:
appveyor build break is not related to my changes. 
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find PythonInterp: Found unsuitable version "2.7.17", but
  required is at least "3" (found C:/Python27/python.exe)

horenmar:
Thanks.

PR #2730, opened 2023-08-11
Edited amalgamated file generator, to block REUSE from getting confused

rpavlik:
## Description
I use the [REUSE tool](https://reuse.software) to generate a software bill-of-materials
for projects I work on. Some of them include a copy of Catch2. However, the REUSE
tool gets confused by code or docs that include things that look like copyright/license statements.

(Without this patch, the following error happens:

```
reuse._util - ERROR - Could not parse 'BSL-1.0\n','
reuse.project - ERROR - 'src/external/catch2/tools/scripts/generateAmalgamatedFiles.py' holds an SPDX expression that cannot be parsed, skipping the file
```
)

They do support comments to disable recognition, so I have added those in the script in question

(It would be really cool if all of Catch2 passed REUSE, but at least not breaking the tool is a start.)

### DIFF
--- a/tools/scripts/generateAmalgamatedFiles.py
+++ b/tools/scripts/generateAmalgamatedFiles.py
@@ -1,4 +1,9 @@
 #!/usr/bin/env python3
+#              Copyright Catch2 Authors
+# Distributed under the Boost Software License, Version 1.0.
+#   (See accompanying file LICENSE.txt or copy at
+#        https://www.boost.org/LICENSE_1_0.txt)
+# SPDX-License-Identifier: BSL-1.0
 
 import os
 import re
@@ -11,6 +16,8 @@ root_path = os.path.join(catchPath, 'src')
 starting_header = os.path.join(root_path, 'catch2', 'catch_all.hpp')
 output_header = os.path.join(catchPath, 'extras', 'catch_amalgamated.hpp')
 output_cpp = os.path.join(catchPath, 'extras', 'catch_amalgamated.cpp')
+
+# REUSE-IgnoreStart
 
 # These are the copyright comments in each file, we want to ignore them
 copyright_lines = [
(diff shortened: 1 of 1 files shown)


horenmar:
> It would be really cool if all of Catch2 passed REUSE

What doesn't?

rpavlik:
Here's the result of "reuse lint": 
[reuse.txt](https://github.com/catchorg/Catch2/files/12325568/reuse.txt)

Mostly missing the license by name in the LICENSES folder (easy fix: `reuse download --all`), then a handful of accessory (non-core) files are missing copyright notices and SPDX-License-Identifier lines, it looks like.

All the "code" files and markdown docs should be pretty easy, just add the copyright and `SPDX-License-Identifier:` line. The input and output test data, plus images and json files, presumably would need the license and copyright data in an adjacent `.license` file - see https://reuse.software/tutorial/#binary-and-uncommentable-files

Ideally you'd be able to get all the stuff marked up without having to put data in a `.reuse/dep5` file, since those don't "nest" well when vendoring projects.

horenmar:
Thanks, I'll merge the change to the generator.

I also looked at the report, and have some thoughts:
* A tool should understand well defined spdx identifiers without me having to provide it the license manually.
* Files in `examples/`, `extras/`, and `fuzzing/` should have BSL license info as well, in case the users want to reuse them.
* I do not understand why various other files that are not part of the install tree need copyright info. As an user you should not need to care what the license of say `tools/scripts/releaseCommon.py` is.

rpavlik:
Thanks for the merge!

The REUSE tool is not a general purpose license extractor, it's explicitly for folks who are trying to follow the REUSE specification that describes specific formats of easy to parse machine readable copyright and license data in every file. It's definitely "overkill" by some standards but it is easy to check and enforce so it can be a nice habit.

I think I will just end up subsetting this repo in our own to exclude files we don't need, which should include all the ones that REUSE is complaining about, which will further solve our project's problem, so no worries there.

PR #2731, opened 2023-08-14
Add cstdint to relevant headers

CoolingJam:
## Description
Simple fix to get catch2 to compile on GCC 13.1. GCC 13 [requires explicit inclusion](https://gcc.gnu.org/gcc-13/porting_to.html) of some headers, such as `<cstdint>`


CoolingJam:
Oops, was already fixed but I was accidentally using outdated version of catch2

PR #2732, opened 2023-08-17
fix some bugprone-macro-parentheses warnings

ChristianTackeGSI:
## Description
When using the public headers of catch2 in another project that uses clang-tidy with some checks enabled, then some warnings in catch2's headers are also reported. Those can distract from the real problem in our project. So I would like to reduce the warnings from catch2.

This fixes a bunch of bugprone-macro-parentheses warnings.

See: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/macro-parentheses.html

## GitHub Issues
None

### DIFF
--- a/src/catch2/internal/catch_test_macro_impl.hpp
+++ b/src/catch2/internal/catch_test_macro_impl.hpp
@@ -34,7 +34,7 @@
 #else // CATCH_CONFIG_FAST_COMPILE
 
 #define INTERNAL_CATCH_TRY try
-#define INTERNAL_CATCH_CATCH( handler ) catch(...) { handler.handleUnexpectedInflightException(); }
+#define INTERNAL_CATCH_CATCH( handler ) catch(...) { (handler).handleUnexpectedInflightException(); }
 
 #endif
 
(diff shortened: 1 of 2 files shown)


horenmar:
This does not even compile.

ChristianTackeGSI:
Okay, in my first attempt this used to compile. Now it also fails.

Anyway: I have stripped this down to the part that does compile and also gives a successful `make test`. Hope this is better now?

horenmar:
Thanks, I'll take a look.

> Okay, in my first attempt this used to compile. Now it also fails.

I suspect that you didn't have test build enabled, and because the main library doesn't exercise those macros, breaking them won't stop it from compiling.

ChristianTackeGSI:
Is there anything left, that I forgot about?
What's the next step?

horenmar:
You haven't forgotten about anything, I did forget about the PR though.

src/catch2/catch_message.hpp
```diff
@@ -110,7 +110,7 @@ namespace Catch {
 
 ///////////////////////////////////////////////////////////////////////////////
 #define INTERNAL_CATCH_UNSCOPED_INFO( macroName, log ) \
-    Catch::getResultCapture().emplaceUnscopedMessage( Catch::MessageBuilder( macroName##_catch_sr, CATCH_INTERNAL_LINEINFO, Catch::ResultWas::Info ) << log )
```
horenmar: This breaks streaming usage like

`UNSCOPED_INFO("Status is " << status.ok());`

I see that we do not have this test for `UNSCOPED_INFO`, only for `INFO` -> I assume that's why this change isn't on the `INFO` internals as well.

I added the test in 9c541ca7

horenmar (commented):
I am willing to merge the other changes, as they should not break anything.

src/catch2/catch_message.hpp
ChristianTackeGSI: Okay, removed that part.

PR #2739, opened 2023-08-27
Catch.cmake: Support CMake multi-config with PRE_TEST discovery mode

hkaelber:
## Description

Support CMake multi-config generators with PRE_TEST discovery mode.
Cf. https://github.com/catchorg/Catch2/pull/2670#issuecomment-1687418035

### DIFF
--- a/extras/Catch.cmake
+++ b/extras/Catch.cmake
@@ -176,8 +176,10 @@ function(catch_discover_tests TARGET)
   string(SUBSTRING ${args_hash} 0 7 args_hash)
 
   # Define rule to generate test list for aforementioned test executable
-  set(ctest_include_file "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}_include-${args_hash}.cmake")
-  set(ctest_tests_file "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}_tests-${args_hash}.cmake")
+  set(ctest_file_base "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}-${args_hash}")
+  set(ctest_include_file "${ctest_file_base}_include.cmake")
+  set(ctest_tests_file "${ctest_file_base}_tests.cmake")
+
   get_property(crosscompiling_emulator
     TARGET ${TARGET}
     PROPERTY CROSSCOMPILING_EMULATOR
@@ -218,6 +220,14 @@ function(catch_discover_tests TARGET)
 
   elseif(_DISCOVERY_MODE STREQUAL "PRE_TEST")
 
+    get_property(GENERATOR_IS_MULTI_CONFIG GLOBAL
+        PROPERTY GENERATOR_IS_MULTI_CONFIG
+    )
+
+    if(GENERATOR_IS_MULTI_CONFIG)
+      set(ctest_tests_file "${ctest_file_base}_tests-$<CONFIG>.cmake")
+    endif()
+
     string(CONCAT ctest_include_content
       "if(EXISTS \"$<TARGET_FILE:${TARGET}>\")"                                    "\n"
       "  if(NOT EXISTS \"${ctest_tests_file}\" OR"                                 "\n"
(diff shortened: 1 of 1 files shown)


horenmar:
How far along is this?

hkaelber:
> How far along is this?

@horenmar, IMHO it's ready for review. I was hesitating, because of missing experience with the usage of multi-config targets. After re-reading the docs, made it finally work correctly. Could you pls review?

extras/Catch.cmake
```diff
@@ -249,7 +259,15 @@ function(catch_discover_tests TARGET)
       "endif()"                                                                    "\n"
     )
 
-    file(GENERATE OUTPUT "${ctest_include_file}" CONTENT "${ctest_include_content}")
+    if(GENERATOR_IS_MULTI_CONFIG)
+      foreach(_config ${CMAKE_CONFIGURATION_TYPES})
+        file(GENERATE OUTPUT "${ctest_file_base}_include-${_config}.cmake" CONTENT "${ctest_include_content}" CONDITION $<CONFIG:${_config}>)
+      endforeach()
+      file(WRITE "${ctest_include_file}" "include(\"${ctest_file_base}_include-\${CTEST_CONFIGURATION_TYPE}.cmake\")")
```
horenmar: I would love to see this dummy file check that `CTEST_CONFIGURATION_TYPE` is set, and if it is not, providing a nice error message instead of this beauty

```
ctest
Test project C:/ubuntu/temp/discover-tests-tests/msvc-sln-1
CMake Error at C:/ubuntu/temp/discover-tests-tests/msvc-sln-1/tests-b12d07c_include.cmake:1 (include):
  include could not find requested file:

    C:/ubuntu/temp/discover-tests-tests/msvc-sln-1/tests-b12d07c_include-.cmake
  CTestTestfile.cmake:7 (include)


Errors while running CTest
```

(This is not needed for merging, but it would be nice 😃 )

horenmar (commented):
I tested this with MSVC and it seems to work.

extras/Catch.cmake
hkaelber: Pls recheck

horenmar: Great, thanks.

extras/Catch.cmake
```diff
@@ -249,7 +259,22 @@ function(catch_discover_tests TARGET)
...
+    if(GENERATOR_IS_MULTI_CONFIG)
+      foreach(_config ${CMAKE_CONFIGURATION_TYPES})
+        file(GENERATE OUTPUT "${ctest_file_base}_include-${_config}.cmake" CONTENT "${ctest_include_content}" CONDITION $<CONFIG:${_config}>)
+      endforeach()
+      string(CONCAT ctest_include_multi_content
+        "if(NOT CTEST_CONFIGURATION_TYPE)"                                              "\n"
+        "  message(\"No configuration for testing specified, use '-C <cfg>'.\")"        "\n"
+        "else()"                                                                        "\n"
+        "  include(\"${ctest_file_base}_include-\${CTEST_CONFIGURATION_TYPE}.cmake\")"  "\n"
+        "endif()"                                                                       "\n"
```
horenmar, line 271: This is fine to merge already, but I still have a question: is there a reason to do this like this (outside of consistency with the surrounding code), rather than either making the newlines part of the line-strings, or using multiline strings?

IIRC something like this would work as well and be simpler to read

```cmake
set(ctest_include_multi_content [[
if(NOT CTEST_CONFIGURATION_TYPE)
  message("No configuration for testing specified, use '-C <cfg>'.")
else()
  include("C:/ubuntu/temp/discover-tests-tests/msvc-sln-1/tests-b12d07c_include-${CTEST_CONFIGURATION_TYPE}.cmake")
endif()
]])
```

horenmar: Right, the issue is that we want to interpret _some_ of the variables, while `[[]]` does not do any interpretation. nevermind then

PR #2740, opened 2023-08-30
implement stringify for std::nullopt_t

mjerabek:
## Description

With `CHECK(some_optional == std::nullopt)`, `nullopt` was stringified differently than `std::optional<T>()` (that is, `{?}` as opposed to `{ }`). That's because `std::nullopt` is not an optional but a special type.

This MR adds stringification for `std::nullopt_t`.

### DIFF
--- a/src/catch2/catch_tostring.hpp
+++ b/src/catch2/catch_tostring.hpp
@@ -398,6 +398,12 @@ namespace Catch {
             }
         }
     };
+    template <>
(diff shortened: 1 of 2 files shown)


mjerabek:
Btw, is it expected that I didn't have to modify the approval tests? That is, is it expected that `ToStringOptional.tests.cpp` is not included in them?

horenmar:
That's what the approvals tag does.

PR #2741, opened 2023-09-09
Set IWYU pragma export in catch_all.hpp

vient:
## Description

Fix IWYU warnings when using catch_all.hpp

### DIFF
--- a/src/catch2/catch_all.hpp
+++ b/src/catch2/catch_all.hpp
@@ -22,109 +22,109 @@
 #ifndef CATCH_ALL_HPP_INCLUDED
 #define CATCH_ALL_HPP_INCLUDED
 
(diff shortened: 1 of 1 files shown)


horenmar:
1) You want [pragma `begin_exports`/`end_exports`](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-begin_exportsend_exports)
2) Won't this cause issues where IWYU will try to replace includes of other headers with `catch_all.hpp`?

vient:
1. I'm actually using clangd, they only support most basic pragmas :( https://clangd.llvm.org/guides/include-cleaner#iwyu-pragmas
2. I checked that clangd does not do this, not sure about original IWYU tool

shahsb (approved):
LGTM!!

PR #2749, opened 2023-09-25
Update tutorial.md

simhonchourasia:
Make it more clear that using SECTIONs does the entire test case from the beginning


horenmar:
Sure, why not.

PR #2750, opened 2023-09-26
why-catch.md: Add JetBrains survey link

perlun:
I found this statement interesting, so I looked at the JetBrains Developer Survey results to see where it could be found. Perhaps linking to this directly would make sense, so people can validate the claim for themselves.


horenmar:
Sure, why not.

PR #2754, opened 2023-10-23
Improve formatting of test specification docs

amerry:
## Description

The existing formatting created one-element lists separated by paragraphs, when it would make more sense to have the paragraphs that are providing more information about one of those list entries be part of the list entry itself.

I think this makes the documentation easier to read in both markdown and html form, and should also improve the structure for assistive technologies.


horenmar:
That probably does look a bit better, thanks.

PR #2756, opened 2023-10-31
Mark `CATCH_CONFIG_` options as advanced

ChrisThrasher:
## Description
These options are rather low-level and don't need to be seen in the CMake cache unless you opt into seeing all other advanced options.

This removes a lot of cache entries from the screen when using a GUI or TUI to view the cache thus making it easier for developers of Catch2 and users of it using FetchContent to focus on the cache variables they're most likely to change.

Before:
Screenshot 2023-10-31 at 10 20 16 AM

After:
Screenshot 2023-10-31 at 10 18 58 AM


horenmar (approved):
Yeah, sure.

PR #2757, opened 2023-11-01
Update cmake-integration.md

joshheyse:
Using version 3.0.1 has build errors using gcc 13.2.1 and set(CMAKE_CXX_STANDARD 20)

Moving up to 3.4.0 resolves the issues.


horenmar:
Identical change was merged in #2762.

PR #2759, opened 2023-11-07
Increase build parallelism on macOS

kkarbowiak:
According to [this information](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), all the macOS runners have at least 3 cores, so it makes sense to increase the build parallelism to this number.

### DIFF
--- a/.github/workflows/mac-builds.yml
+++ b/.github/workflows/mac-builds.yml
@@ -42,11 +42,10 @@ jobs:
 
     - name: Build tests + lib
       working-directory: ${{runner.workspace}}/build
-      run: make -j 3
+      run: make -j `sysctl -n hw.ncpu`
 
     - name: Run tests
       env:
           CTEST_OUTPUT_ON_FAILURE: 1
       working-directory: ${{runner.workspace}}/build
-      # Hardcode 3 cores we know are there
-      run: ctest -C ${{matrix.build_type}} -j 3
+      run: ctest -C ${{matrix.build_type}} -j `sysctl -n hw.ncpu`
--- a/.github/workflows/mac-builds.yml
+++ b/.github/workflows/mac-builds.yml
@@ -42,11 +42,11 @@ jobs:
 
     - name: Build tests + lib
       working-directory: ${{runner.workspace}}/build
-      run: make -j 2
+      run: make -j 3
 
     - name: Run tests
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks, but since the time I wrote these I learned a better way to do this (see review).

.github/workflows/mac-builds.yml
```diff
@@ -42,11 +42,11 @@ jobs:
 
     - name: Build tests + lib
       working-directory: ${{runner.workspace}}/build
-      run: make -j 2
+      run: make -j 3
```
horenmar:
```suggestion
      run: make -j `sysctl -n hw.ncpu`
```

.github/workflows/mac-builds.yml
```diff
@@ -42,11 +42,11 @@ jobs:
...
 
     - name: Run tests
       env:
           CTEST_OUTPUT_ON_FAILURE: 1
       working-directory: ${{runner.workspace}}/build
-      # Hardcode 2 cores we know are there
-      run: ctest -C ${{matrix.build_type}} -j 2
+      # Hardcode 3 cores we know are there
+      run: ctest -C ${{matrix.build_type}} -j 3
```
horenmar:
```suggestion
      run: ctest -C ${{matrix.build_type}} -j `sysctl -n hw.ncpu`
```

janisozaur:
While make is not as slow as some other build systems (**cough** msbuild **cough**), ninja is still faster and autodetects core count.

.github/workflows/mac-builds.yml
kkarbowiak: Cool! In my projects I just use `-j`, which executes **all** in parallel, but those are pretty small (less than 20 source files), so the risk of exhausting available memory is negligible.

horenmar:
I usually don't see ninja being faster in from-zero full builds (it is awesome for incremental dev builds though), and last time I tried it wasn't installed on the VMs by default so the build had to spend time installing it every time.

janisozaur:
It seems the GitHub team is indeed opposed to installing ninja by default: https://github.com/actions/runner-images/issues/8348

PR #2760, opened 2023-11-08
Bump up actions/checkout version to v4

kkarbowiak:
This gets rid of the warning displayed in actions run annotation:
`The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2.`


horenmar:
Thanks

PR #2762, opened 2023-11-10
lifted suggested version in documentation from 3.0.1 to 3.4.0 to avoid compilation issues

senarclens:
Did this, because with 3.0.1, the files
- catch_xmlwriter.cpp
- catch_string_manip.hpp
- catch_test_case_info.hpp

were missing

This led to some obvious and some obscure compile errors when compiling with g++ 13.2.0 (std c++17 or c++20)
One of these errors being "Elaborated-type-specifier for a scoped enum must not use the ‘class’ keyword"

Since this is a rather bad experience and debugging it is time-consuming, I suggest to simply boost the recommended version in the snippet which is likely copied by new Catch2 users

### DIFF
--- a/docs/cmake-integration.md
+++ b/docs/cmake-integration.md
@@ -51,7 +51,7 @@ Include(FetchContent)
 FetchContent_Declare(
   Catch2
   GIT_REPOSITORY https://github.com/catchorg/Catch2.git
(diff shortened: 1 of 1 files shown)


horenmar:
Sure, why not.

Issue #1879, opened 2020-03-06
TeamCity reporter should report SECTIONs


heidemn:
**Description**
The JUnit reporter reports SECTIONs.
The TeamCity reporter does not.
Not sure if reporting SECTIONs should always be enabled, but could there be at least an option?

Except for this, I would very much like to use the TeamCity reporter, since it seems to use a lot less memory compared to the JUnit reporter (10 MB vs. 4 GB in my case).

**Additional context**
Tried with latest version, Catch2 2.11.1.

**Compare JUnit report:**
![image](https://user-images.githubusercontent.com/5494429/76091847-a3844100-5fbe-11ea-8e23-09e2ab540b19.png)

**...with TeamCity report:**
![image](https://user-images.githubusercontent.com/5494429/76091665-4f795c80-5fbe-11ea-8d22-9ce20dd5e64e.png)

PhilipSoDisguise:
Hi, I have the exact issue going on,
any update on this as of 2023?
Thank you

horenmar:
We now support a per-reporter options, so I would accept a PR that makes this change.

PR #2765, opened 2023-11-14
Add sections support to TeamCity reporter

swichu91:
## Description
This MR allows for printing SECTIONs when using TeamCity reporter.

Two new TeamCity reporter options were added: 
`Xsections=true/false` - enable printing sections
`Xseparator=` - optional, override the default separator(".") when composing unique test cases from sections

As TeamCity only operates on ordinary test cases(no extra sections tags as far as I know), it is required to flatten the section tree. 
Test case names are built as follows:`name='<TEST_CASE>.<SECTION>.*']`
Additionally, any spaces within test/section name are replaced with underscores.

An example of entry:
`##teamcity[testStarted name='basic_test.check_if_x.should_do_y']`

TBH, I would get rid of section option in favour of supporting it permanently. Such change will make code much simpler and more readable, but the decision is up to you.

## GitHub Issues
Mentioned here: https://github.com/catchorg/Catch2/issues/1879

### DIFF
--- a/src/catch2/reporters/catch_reporter_teamcity.cpp
+++ b/src/catch2/reporters/catch_reporter_teamcity.cpp
@@ -5,68 +5,66 @@
 //        https://www.boost.org/LICENSE_1_0.txt)
 
 // SPDX-License-Identifier: BSL-1.0
+#include <catch2/catch_test_case_info.hpp>
+#include <catch2/internal/catch_enforce.hpp>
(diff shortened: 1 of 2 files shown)


horenmar:
Why are spaces replaced with underscores? I was under the impression that TeamCity can handle spaces in names.

krcook:
What needs to happen to move this pull request forward?

PR #2772, opened 2023-12-01
Fix compilation on Xbox platforms

SupSuper:
Xbox does not support getenv.

`_GAMING_XBOX` is defined by the [Microsoft GDK](https://learn.microsoft.com/en-us/gaming/gdk/) on Xbox platforms.


horenmar:
Thanks

Issue #2583, opened 2022-11-27
-Wsign-conversion error in catch hpp


omnibench:
Catch2 is causing my build to fail with -Wsign-conversion on:
```
[build] build/_deps/catch2-src/src/catch2/..\catch2/internal/catch_clara.hpp:126:59: error: implicit conversion changes signedness: 'std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<std::basic_string<char>>>>::difference_type' (aka 'long long') to 'unsigned long long' [-Werror,-Wsign-conversion]
[build]                     return m_tokenBuffer.size() + ( itEnd - it );
[build]                            
```

ChrisThrasher:
I tried fixing the sign conversion warnings in Catch2 and it was not easy. You're better off marking the Catch2 include directories as `SYSTEM` include directories (if you're using CMake) to disable warnings in Catch headers.

topazus:
```cmake
add_compile_options(-Wno-error=conversion)
add_compile_options(-Wno-error=sign-conversion)
```
Does this method is recommended?

ChrisThrasher:
That will get the job done but what I said before is the recommended answer. You should not be applying extra compiler warnings to Catch2 itself. All compilers have ways of saying that headers in a given directory belong to other projects and thus don't emit warnings about them. The CMake abstraction for this is something called "`SYSTEM` include directories".

topazus:
Thanks. After digging into this, I got your points. I also found your efforts of suppressing Catch2 warnings in SFML project.

https://github.com/SFML/SFML/blob/e45628e2ebc5843baa3739781276fa85a54d4653/test/CMakeLists.txt#L18-L22

louwers:
The CMake guide should probably be updated to reflect that this is needed: https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md

I tried

```
FetchContent_Declare(
  Catch2
  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
  GIT_TAG        v3.7.1
  SYSTEM
)
```

but I still get the warning.

For me the warning comes from a source file:

```
/Users/bart/src/geometry.hpp/build/_deps/catch2-src/src/catch2/reporters/catch_reporter_console.cpp:354:44: error: implicit conversion changes signedness: 'int' to 'size_type' (aka 'unsigned long') [-Werror,-Wsign-conversion]
[...]

PR #2775, opened 2023-12-06
Add -Wsign-conversion to the list of common warning flags

Wmbat:
## Description

Fixes all implicit sign conversions and turns on -Wsign-conversion when building catch2. I usually tried to find a good solution to the warning instead of casting. 

Ran all the tests on my machine and it seemed good to go

## GitHub Issues
Related to #2583


src/catch2/reporters/catch_reporter_console.cpp
```diff
@@ -297,7 +297,7 @@ class TablePrinter {
     std::ostream& m_os;
     std::vector<ColumnInfo> m_columnInfos;
     ReusableStringStream m_oss;
-    int m_currentColumn = -1;
+    std::size_t m_currentColumn = 0;
```
oboukli: Would it be better if this was an iterator?

shahsb: +1

shahsb (approved):
This is a good improvement and will make a great impact to the project. Thanks @Wmbat for making these changes.. Good Job!!

src/catch2/reporters/catch_reporter_console.cpp
Wmbat: Changed

Wmbat:
Fixed the merge conflicts and switch to using an iterator instead of an index

not fully sure why it says I modified so many files though

PR #2785, opened 2023-12-27
Make compiler flags for reproducible builds optional

jpalus:
no change in default behavior: -ffile-prefix-map is still added if supported by compiler but add option for disabling it as it breaks generic tools for extracting debuginfo/debugsource from build artifacts like: https://sourceware.org/git/?p=debugedit.git;a=blob;f=scripts/find-debuginfo.in


horenmar:
Sure, why not.

PR #2793, opened 2024-01-10
Improve tutorial.md with link to two-file setup instructions.

iglesias:
## Description
Add link in tutorial because otherwise it is not clear how to follow those steps (for example, it is not mentioned where those two files can be found) and it requires more effort to start.


iglesias:
Hi, @horenmar, may I ask you whether this change can be merged? Excuses the insistence.

horenmar:
Makes sense, thanks.

PR #2794, opened 2024-01-10
reporters: silence -Wsubobject-linkage

timblechmann:
gcc emits `Wsubobject-linkage`, because the the publicly visible `TablePrinter` contains `ColumnInfo`, which is part of an anonymous namespace

## Description
```
path/to/src/catch2/reporters/catch_reporter_console.cpp:296:7: warning: ‘Catch::TablePrinter’ has a field ‘Catch::TablePrinter::m_columnInfos’ whose type uses the anonymous namespace [-Wsubobject-linkage]
  296 | class TablePrinter {
      |       ^~~~~~~~~~~~
```

### DIFF
--- a/src/catch2/reporters/catch_reporter_console.cpp
+++ b/src/catch2/reporters/catch_reporter_console.cpp
@@ -209,13 +209,6 @@ findMax( std::size_t& i, std::size_t& j, std::size_t& k, std::size_t& l ) {
         return l;
 }
 
-enum class Justification { Left, Right };
-
-struct ColumnInfo {
-    std::string name;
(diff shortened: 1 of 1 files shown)


horenmar:
What GCC versions do you see this with? I tried adding the flag locally and 11.4.0 does not warn about the current code.

timblechmann:
i'm seeing this with gcc-13 (i could assume that unity builds / lto may also affect this warning being emitted).

as a rule of thumb a struct that's not in an anonymous namespace should not contain any members that are defined in an anonymous namespace

PR #2801, opened 2024-01-23
Fix clang-tidy `bugprone-chained-comparison` warnings

ChrisThrasher:
## Description
This triggers when running clang-tidy's bugprone-* family of checks in code which uses Catch2 even if Catch2 headers are marked as SYSTEM headers due to how code expanded from macros is treated as first party even if the macro comes from a 3rd party library. The fix is luckily pretty straightforward.

This check is added in clang-tidy-18 due for release later this year.

https://github.com/llvm/llvm-project/pull/76365


ChrisThrasher:
CI has exploded so perhaps this is not the right fix

ChrisThrasher:
Looks like we can't change the C++ to fix this and instead have to add a new suppression to deal with this. I don't love having to do this but it's worth it for providing a better user experience.

Issue #2700, opened 2023-06-06
update conan to v2


nrbnlulu:
**Describe the bug**
conan warns on deprecated API's usage
```console
WARN: deprecated: Usage of deprecated Conan 1.X features that will be removed in Conan 2.X:
WARN: deprecated:     'cpp_info.filenames' used in: catch2/3.1.0
WARN: deprecated:     'cpp_info.names' used in:  catch2/3.1.0
WARN: deprecated:     'cpp_info.build_modules' used in:  catch2/3.1.0
```

**Expected behavior**
No deprecation warnings.

**Reproduction steps**
try to build a conan project with `catch2/3.1.0` as a requirement.

**Platform information:**

 - OS: **Ubuntu 22.0.4**
 - Compiler+version: 
 ```console
Ubuntu clang version 14.0.0-1ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
 ```
 - Catch version: **v3.1.0**

horenmar:
I don't use Conan, so the port is community effort -> if someone goes and fixes it up I will merge it.

LinuxDevon:
Was looking into fixing this. So two notes I know from playing around with this some:
1. The build script we have `.conan/build.py` will need to be updated / changed maybe. See this message for why: https://github.com/conan-io/conan-package-tools#conan-20-support-warning
2. If we update to conan 2.0 all conan 1.X will break. I am not aware of how to maintain compatibility and think it is just the price we pay for moving up to it.

If I proceed, is there anyone with a little more knowledge on Conan 2? Want to bounce some questions off them. Also not sure the test path here beside ensuring that the test_package works. Is that enough along with the removal of the warnings?

ChrisThrasher:
> If we update to conan 2.0 all conan 1.X will break. I am not aware of how to maintain compatibility and think it is just the price we pay for moving up to it.

I know it's possible to write Conan recipes compatible with v1 and v2. I don't know the details of how you achieve that though.

LinuxDevon:
I pushed a PR yesterday. It does in fact maintain backwards compatibility. Wasn't too bad but the documentation for conan is not great...

PR #2805, opened 2024-02-01
Conan v2 

LinuxDevon:
## Description

This is to address the warnings and upcoming Conan 1 deprecation. 

I compared the changes of the package before and after. I took a snapshot of the directory using conan 1 before changes. Then one with conan 2 after. 

See the text files:
- [filesystem_conan1.txt](https://github.com/catchorg/Catch2/files/14120593/filesystem_conan1.txt)
- [filesystem_conan2.txt](https://github.com/catchorg/Catch2/files/14120594/filesystem_conan2.txt)

I added a build yaml for conan versions i tested against. I didn't do any c++ versioning or gcc testing since that is done with the other builds. I simply wanted to ensure that conan ran and could build against the test package. I kept backwards compatibility with v1.

There is a `.conan/build.py` file i was debating about removing. I don't think it is used but if it is how? The [conan-package-tools](https://github.com/conan-io/conan-package-tools/tree/develop) don't work with conan 2.0. There aren't plans to support so not sure if that changes things as well.

## GitHub Issues

#2700


conanfile.py
```diff
@@ -8,53 +14,86 @@ class CatchConan(ConanFile):
...
+            self.cpp_info.set_property("cmake_file_name", "Catch2")
+
+            # Catch2
+            self.cpp_info.components["catch2base"].set_property("cmake_file_name", "Catch2")
+            self.cpp_info.components["catch2base"].set_property("pkg_config_name", "Catch2")
+            self.cpp_info.components["catch2base"].libs = ["Catch2" + lib_suffix]
+            self.cpp_info.components["catch2base"].builddirs.append("lib/cmake/Catch2")
+
+            # Catch2WithMain
+            self.cpp_info.components["catch2main"].set_property("cmake_file_name", "Catch2WithMain")
```
Minimonium: Please do specify `cmake_target_name` similar to
https://github.com/conan-io/conan-center-index/blob/master/recipes/catch2/3.x.x/conanfile.py#L168

That way you would be able to correctly use `Catch2::Catch2WithMain` target instead of the lowercase.

LinuxDevon: Should be updated. Does this not fix the issue then with the linked issue? I didn't realize that the conan file was seperate in the conan.io repo.

LinuxDevon: I think i did this. I kept the `cmake_file_name` but not sure if that matters. 

I also didn't realize that the conan file for conan.io is separate from this repo. So will this close the issue in reference? Maybe if it is maintained over there we just drop the conan files here

Minimonium: The Conan Center policy is to maintain recipes for the default repo so in-source conanfiles are only needed for project-specific development. I'm not sure how many people require it so it's up to the maintainer in here to decide if it's still needed. 

Regarding this specific PR - `cmake_target_name` is used for the target name. It needs to be specified with namespace because of the different filename to namespace to target to component conventions different projects have. `cmake_file_name` is needed only for the `find_package` syntax, it's not useful for components and it doesn't make sense to specify the target name in there. The global level `cmake_target_name` I think refers to the target which is wired on Conan-level global component dependencies, it's not actually used in CMake. 

Additionally, prefer `os.path.join` over the `/` directory separator for cross-environment compatibility.

horenmar:
Thanks

PR #2807, opened 2024-02-02
Clarify unit of test duration in docs

juliapaluch:
## Description

Test duration with `-d yes` or with `-D` is shown in seconds with millisecond precision, not in milliseconds.


horenmar:
Thanks

PR #2817, opened 2024-02-19
Evaluate argument of (DYNAMIC_)SECTION, TEST_CASE in static-analysis mode

### LINKED ISSUE #2816: Wunused-variable in (DYNAMIC_)SECTION when CATCH_CONFIG_EXPERIMENTAL_STATIC_ANALYSIS_SUPPORT is enabled
**Describe the bug**

In static analysis mode, the argument of `SECTION`/`DYNAMIC_SECTION` is not evaluated. If it contains
a variable that is used only there, it'll lead to `Wunused-variable` warning.

**Expected behavior**
No warning

**Reproduction steps**

https://compiler-explorer.com/z/K9fxKGaz4

**Platform information:**
 - Compiler+version: clang (clang-tidy)
 - Catch version: devel

mjerabek:
## Description

As the title says, it's fairly straightforward.

And also add configuration for `clang-tidy`, run it in CI, and fix/suppress the resulting warnings.

## GitHub Issues

Closes #2816

### DIFF
--- a/.github/workflows/linux-other-builds.yml
+++ b/.github/workflows/linux-other-builds.yml
@@ -103,3 +103,52 @@ jobs:
           CTEST_OUTPUT_ON_FAILURE: 1
       working-directory: ${{runner.workspace}}/build
       run: ctest -C ${{matrix.build_type}} -j `nproc` ${{matrix.other_ctest_args}}
+  clang-tidy:
+    name: clang-tidy ${{matrix.version}}, ${{matrix.build_description}}, C++${{matrix.std}} ${{matrix.build_type}}
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        include:
+          - version: "15"
+            build_description: all
+            build_type: Debug
+            std: 17
+            other_pkgs: ''
+            cmake_configurations: -DCATCH_BUILD_EXAMPLES=ON -DCATCH_ENABLE_CMAKE_HELPER_TESTS=ON
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Prepare environment
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y ninja-build clang-${{matrix.version}} clang-tidy-${{matrix.version}} ${{matrix.other_pkgs}}
+
+    - name: Configure build
+      working-directory: ${{runner.workspace}}
+      env:
+        CXX: clang++-${{matrix.version}}
+        CXXFLAGS: ${{matrix.cxxflags}}
+      # Note: $GITHUB_WORKSPACE is distinct from ${{runner.workspace}}.
+      #       This is important
+      run: |
+        clangtidy="clang-tidy-${{matrix.version}};-use-color"
+        # Use a dummy compiler/linker/ar/ranlib to effectively disable the
+        # compilation and only run clang-tidy.
+        cmake -Bbuild -H$GITHUB_WORKSPACE \
+              -DCMAKE_BUILD_TYPE=${{matrix.build_type}} \
+              -DCMAKE_CXX_STANDARD=${{matrix.std}} \
+              -DCMAKE_CXX_STANDARD_REQUIRED=ON \
+              -DCMAKE_CXX_EXTENSIONS=OFF \
+              -DCATCH_DEVELOPMENT_BUILD=ON \
+              -DCMAKE_CXX_CLANG_TIDY="$clangtidy" \
+              -DCMAKE_CXX_COMPILER_LAUNCHER=/usr/bin/true \
+              -DCMAKE_AR=/usr/bin/true \
+              -DCMAKE_CXX_COMPILER_AR=/usr/bin/true \
+              -DCMAKE_RANLIB=/usr/bin/true \
+              -DCMAKE_CXX_LINK_EXECUTABLE=/usr/bin/true \
+              ${{matrix.cmake_configurations}} \
+              -G Ninja
+
+    - name: Run clang-tidy
+      working-directory: ${{runner.workspace}}/build
+      run: ninja
--- a/src/catch2/internal/catch_reporter_spec_parser.cpp
+++ b/src/catch2/internal/catch_reporter_spec_parser.cpp
@@ -117,7 +117,7 @@ namespace Catch {
             auto kv = splitKVPair( parts[i] );
             auto key = kv.key, value = kv.value;
 
-            if ( key.empty() || value.empty() ) {
+            if ( key.empty() || value.empty() ) { // NOLINT(bugprone-branch-clone)
                 return {};
             } else if ( key[0] == 'X' ) {
                 // This is a reporter-specific option, we don't check these
--- a/tests/ExtraTests/X91-AmalgamatedCatch.cpp
+++ b/tests/ExtraTests/X91-AmalgamatedCatch.cpp
@@ -16,10 +16,10 @@
 TEST_CASE("Just a dummy test") {
     auto i = GENERATE(1, 2, 3);
     SECTION("a") {
-        REQUIRE(1 != 4);
+        REQUIRE(i != 4);
     }
     SECTION("b") {
-        CHECK(1 != 5);
+        CHECK(i != 5);
     }
     REQUIRE_THAT(1,
                  Catch::Matchers::Predicate<int>([](int i) {
--- a/tests/SelfTest/UsageTests/Benchmark.tests.cpp
+++ b/tests/SelfTest/UsageTests/Benchmark.tests.cpp
@@ -90,14 +90,14 @@ TEST_CASE("Benchmark containers", "[!benchmark]") {
         };
         REQUIRE(v.size() == size);
 
-        int array[size];
+        int array[size] {};
         BENCHMARK("A fixed size array that should require no allocations") {
             for (int i = 0; i < size; ++i)
                 array[i] = i;
         };
         int sum = 0;
-        for (int i = 0; i < size; ++i)
-            sum += array[i];
+        for (int val : array)
+            sum += val;
         REQUIRE(sum > size);
 
         SECTION("XYZ") {
@@ -121,8 +121,8 @@ TEST_CASE("Benchmark containers", "[!benchmark]") {
                 runs = benchmarkIndex;
             };
 
-            for (size_t i = 0; i < v.size(); ++i) {
-                REQUIRE(v[i] == runs);
+            for (int val : v) {
+                REQUIRE(val == runs);
             }
         }
     }
@@ -135,8 +135,8 @@ TEST_CASE("Benchmark containers", "[!benchmark]") {
             for (int i = 0; i < size; ++i)
                 v[i] = generated;
         };
-        for (size_t i = 0; i < v.size(); ++i) {
-            REQUIRE(v[i] == generated);
+        for (int val : v) {
+            REQUIRE(val == generated);
         }
     }
 
--- a/tests/SelfTest/UsageTests/Matchers.tests.cpp
+++ b/tests/SelfTest/UsageTests/Matchers.tests.cpp
@@ -1027,7 +1027,6 @@ TEST_CASE( "Combining MatchNotOfGeneric does not nest",
 }
 
 struct EvilAddressOfOperatorUsed : std::exception {
-    EvilAddressOfOperatorUsed() {}
     const char* what() const noexcept override {
         return "overloaded address-of operator of matcher was used instead of "
                "std::addressof";
@@ -1035,7 +1034,6 @@ struct EvilAddressOfOperatorUsed : std::exception {
 };
 
 struct EvilCommaOperatorUsed : std::exception {
-    EvilCommaOperatorUsed() {}
     const char* what() const noexcept override {
         return "overloaded comma operator of matcher was used";
     }
@@ -1073,7 +1071,6 @@ struct ImmovableMatcher : Catch::Matchers::MatcherGenericBase {
 };
 
 struct MatcherWasMovedOrCopied : std::exception {
-    MatcherWasMovedOrCopied() {}
     const char* what() const noexcept override {
         return "attempted to copy or move a matcher";
     }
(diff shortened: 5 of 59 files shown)


mjerabek:
Not sure if/where should I add tests for this.

horenmar:
I am _reasonably sure_ that this breaks `SECTION` macros in some cases, so a test would be nice 😛 .

This is a valid `SECTION` start: `SECTION("short", "long description with many many words")`, and unless I am very mistaken, your change would break it.

mjerabek:
Huh, interesting. I had no idea it can be used like that :smile: 

Any tips where the tests for this should be?

mjerabek:
- fixed the implementation (yes, the tests uncovered other bugs)
- added basic clang-tidy run to CI

Some limitations:
- Some files are checked multiple times, because they are present in `compile_commands.json` multiple times with different compile options. Also I'm quite certain that `run-clang-tidy` doesn't handle this well and we run it multiple times with the same configuration.
- `run-clang-tidy` doesn't stop on error, so it always runs in full. If we used cmake's clang-tidy integration, we could rely on ninja. But cmake doesn't support running clang-tidy without also running the compiler, so it would either take longer, or we would have to fake a "null compiler" (I've done that in work, it works well, but it's not very nice). Or we could put together a short python script that generates a ninja build file and run that (I would probably slightly prefer that to the current solution, as we could also better deal with duplicate files in `compile_commands.json`).
- Only the default checks are enabled at the moment. We should probably explicitly define what we want to check and what we want to consider an error.
- If there are errors, the job fails (good). But we ignore any warnings. :shrug: 

Also, I might add some clang-tidy-specific tests later to test that `clang-tidy` understands that `REQUIRE(expr)` implies `expr` (and also implement it, because now it most probably cannot understand that). But that's for a follow-up PR.

tests/SelfTest/UsageTests/Benchmark.tests.cpp
```diff
@@ -97,7 +97,7 @@ TEST_CASE("Benchmark containers", "[!benchmark]") {
         };
         int sum = 0;
         for (int i = 0; i < size; ++i)
-            sum += array[i];
+            sum += array[i]; // NOLINT(clang-analyzer-core.uninitialized.Assign)
```
horenmar: Wouldn't a better fix be changing `int array[size]` into `int array[size]{}` above? The test below would still require the change in `BENCHMARK` to happen to pass, but it should shut up the analyzer.

tests/ExtraTests/X91-AmalgamatedCatch.cpp
```diff
@@ -14,7 +14,7 @@
 #include "catch_amalgamated.hpp"
 
 TEST_CASE("Just a dummy test") {
-    auto i = GENERATE(1, 2, 3);
+    [[maybe_unused]] auto i = GENERATE(1, 2, 3);
     SECTION("a") {
         REQUIRE(1 != 4);
```
horenmar: I _think_ this might've been supposed to be `REQUIRE(i != 4)`, and similarly in the next section, so the `maybe_unused` wouldn't be needed.

.github/workflows/linux-other-builds.yml
```diff
@@ -103,3 +103,48 @@ jobs:
...
+  clang-tidy:
+    name: clang-tidy-${{matrix.version}}, ${{matrix.build_description}}, C++${{matrix.std}} ${{matrix.build_type}}
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          # Extras and examples with clang-tidy-10
+          - version: "10"
+            tidy_options: ''
+            build_description: Extras + Examples
```
horenmar:
```suggestion
            build_description: clang-tidy
```
instead?

horenmar:
> run-clang-tidy doesn't stop on error, so it always runs in full. If we used cmake's clang-tidy integration, we could rely on ninja. But cmake doesn't support running clang-tidy without also running the compiler, so it would either take longer, or we would have to fake a "null compiler" (I've done that in work, it works well, but it's not very nice). Or we could put together a short python script that generates a ninja build file and run that (I would probably slightly prefer that to the current solution, as we could also better deal with duplicate files in compile_commands.json).

That's not great. Maybe we could merge the clang-tidy check into an already existing build? This would create a build with outsized build time, but we can't build all the builds in parallel either way, so as long as it gets scheduled reasonably early it shouldn't have much impact.

> Some files are checked multiple times, because they are present in compile_commands.json multiple times with different compile options. Also I'm quite certain that run-clang-tidy doesn't handle this well and we run it multiple times with the same configuration.

Yes, they are indeed compiled multiple times in the full test build. Some tests require building against custom configuration of the implementation binary, so they have to rebuild it. You can find them quickly by looking for targets that link against `Catch2_buildall_interface`.

> Only the default checks are enabled at the moment. We should probably explicitly define what we want to check and what we want to consider an error.

Definitely. The checks in clang-tidy can be a mess, with overlapping checks with different implementation, this came up in the original discussion for the static-analysis mode. Having a reasonable set that we can point to as being supported would be nice.

> Also, I might add some clang-tidy-specific tests later to test that clang-tidy understands that REQUIRE(expr) implies expr (and also implement it, because now it most probably cannot understand that).

It almost definitely cannot understand it. The reasoning there is only slightly simpler than figuring out `SECTION`s, and the analyzer helper mode has to help with those a lot. However, from cursory check, there is a relatively reasonable way of [telling clang-tidy about custom assertions](https://clang-analyzer.llvm.org/faq.html#custom_assert). It would require making the `REQUIRE` macro also change its definition based on the defines though.

tests/SelfTest/UsageTests/Benchmark.tests.cpp
mjerabek: Done.

tests/ExtraTests/X91-AmalgamatedCatch.cpp
mjerabek: Done.

.github/workflows/linux-other-builds.yml
mjerabek: "clang-tidy" is already included in the job description. I've replaced this by "all", but it's not really _completely_ all. I guess it's hard to put variant-specific description when there is just one variant.

mjerabek:
In the end, I have rewritten the CI job to use `cmake` to run `clang-tidy`, with a fake compiler (turned out it's not messy when you're not adding PCHs and ccache into the mix, which I now realize are both useless for just clang-tidy run; oh well, I'm putting that on the endless list of pending CI refactorings).

The `clang-tidy` checks are now explicitly configured. The rationale is following: allow-list whole categories, deny-list particular checks from the categories. This way, when a new clang version adds new checks, they are easily discovered (if they print a warning) and may be acted upon - either fix the findings or disable the check if it is deemed not useful. YMMV

Werror is enabled for some concrete checks.

Unfortunately, we cannot enable `bugprone-use-after-move`. First, it cannot be made to understand our custom `unique_ptr` (https://github.com/llvm/llvm-project/issues/36901). Second, it [doesn't understand](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/use-after-move.html) that `if (i == 1)` and `if (i == 2)` are exclusive, so it doesn't work with the `SECTION` hinting.

We also have to disable `bugprone-macro-repeated-side-effects` because if wrongly triggers for `CATCH_MOVE` and `CATCH_FORWARD`, although the second expansion of the macro argument is in unevaluated context (no issue for that yet AFAICT).

.github/workflows/linux-other-builds.yml
horenmar: Right, then "extras + examples" is fine.

mjerabek:
So, I have split each check to its separate commit as you wanted. It's up to you now to select which ones do you want to keep and which ones you want to suppress instead.

tests/SelfTest/UsageTests/Misc.tests.cpp
```diff
@@ -396,7 +396,7 @@ TEMPLATE_PRODUCT_TEST_CASE("Product with differing arities", "[template][product
 using MyTypes = std::tuple<int, char, float>;
 TEMPLATE_LIST_TEST_CASE("Template test case with test types specified inside std::tuple", "[template][list]", MyTypes)
 {
-    REQUIRE(sizeof(TestType) > 0);
+    REQUIRE(sizeof(TestType) > 0); // NOLINT(bugprone-sizeof-expression)
```
horenmar: Is this due to the expression desugaring into `sizeof(sizeof(TestType) > 0)`?

The point here is to just have some assertion in the test case, so it can be replaced with something else instead... For this one `STATIC_REQUIRE(std::is_arithmetic<TestType>::value)` should be fine.

tests/SelfTest/UsageTests/Misc.tests.cpp
```diff
@@ -406,7 +406,7 @@ struct NonDefaultConstructibleType {
 using MyNonDefaultConstructibleTypes = std::tuple<NonDefaultConstructibleType, float>;
 TEMPLATE_LIST_TEST_CASE("Template test case with test types specified inside non-default-constructible std::tuple", "[template][list]", MyNonDefaultConstructibleTypes)
 {
-    REQUIRE(sizeof(TestType) > 0);
+    REQUIRE(sizeof(TestType) > 0); // NOLINT(bugprone-sizeof-expression)
```
horenmar: Same here, but maybe `is_trivially_copyable`.

tests/SelfTest/UsageTests/Misc.tests.cpp
```diff
@@ -421,7 +421,7 @@ struct NonCopyableAndNonMovableType {
 using NonCopyableAndNonMovableTypes = std::tuple<NonCopyableAndNonMovableType, float>;
 TEMPLATE_LIST_TEST_CASE("Template test case with test types specified inside non-copyable and non-movable std::tuple", "[template][list]", NonCopyableAndNonMovableTypes)
 {
-    REQUIRE(sizeof(TestType) > 0);
+    REQUIRE(sizeof(TestType) > 0); // NOLINT(bugprone-sizeof-expression)
```
horenmar: `default_constructible` or something.

tests/SelfTest/UsageTests/ToStringGeneral.tests.cpp
```diff
@@ -27,9 +27,8 @@ TEST_CASE( "Character pretty printing" ){
         char space = ' ';
         CHECK(space == ' ');
         char chars[] = {'a', 'z', 'A', 'Z'};
-        for (int i = 0; i < 4; ++i){
-            char c = chars[i];
-            REQUIRE(c == chars[i]);
+        for (char c : chars){
+            REQUIRE(c == c);
         }
     }
```
horenmar: These tests are all wrong, I'll fix them in main and you can drop the commit.

tests/SelfTest/UsageTests/Message.tests.cpp
```diff
@@ -80,28 +80,28 @@ TEST_CASE( "Output from all sections is reported", "[failing][messages][.]" ) {
...
+        std::clog << "Write to std::clog\n";
     }
     SECTION( "Interleaved writes to cerr and clog" ) {
         std::cerr << "Inter";
         std::clog << "leaved";
         std::cerr << ' ';
         std::clog << "writes";
         std::cerr << " to error";
-        std::clog << " streams" << std::endl;
+        std::clog << " streams\n";
```
horenmar: I suspect these should all explicitly flush for 100% correctness.

tests/SelfTest/UsageTests/Misc.tests.cpp
```diff
@@ -158,9 +158,9 @@ TEST_CASE( "looped tests", "[.][failing]" ) {
 }
 
 TEST_CASE( "Sends stuff to stdout and stderr", "[.]" ) {
-    std::cout << "A string sent directly to stdout" << std::endl;
-    std::cerr << "A string sent directly to stderr" << std::endl;
-    std::clog << "A string sent to stderr via clog" << std::endl;
+    std::cout << "A string sent directly to stdout\n";
+    std::cerr << "A string sent directly to stderr\n";
+    std::clog << "A string sent to stderr via clog\n";
```
horenmar: ditto

tests/SelfTest/UsageTests/Message.tests.cpp
mjerabek: I also wondered about that, but the tests passed, so I thought it should be ok. We can either use `<< '\n' << std::flush` or disable the check.

tests/SelfTest/UsageTests/Misc.tests.cpp
mjerabek: I'm not entirely certain. I think it has to do something with the tested type being `int`.

> `suspicious comparison of 'sizeof(expr)' to a constant

Agree that it's better to replace.

tests/SelfTest/UsageTests/Message.tests.cpp
horenmar: Add the `std::flush`, that's what Catch2 implementation does when it relies on flush happening.

tests/SelfTest/UsageTests/Matchers.tests.cpp
```diff
@@ -1081,6 +1081,7 @@ struct MatcherWasMovedOrCopied : std::exception {
 
 struct ThrowOnCopyOrMoveMatcher : Catch::Matchers::MatcherGenericBase {
     ThrowOnCopyOrMoveMatcher() = default;
+    // NOLINTNEXTLINE(bugprone-copy-constructor-init)
```
horenmar: I am looking at the docs, and apparently changing the code below to `Catch::Matcher::MatcherGenericBase(other)` would work.

Or maybe deleting the initialization of the base class completely might work, as the argument is already marked as unused via deleted name.

tests/SelfTest/UsageTests/Matchers.tests.cpp
```diff
@@ -1079,6 +1079,7 @@ struct MatcherWasMovedOrCopied : std::exception {
 struct ThrowOnCopyOrMoveMatcher : Catch::Matchers::MatcherGenericBase {
     ThrowOnCopyOrMoveMatcher() = default;
 
+    // NOLINTBEGIN(readability-redundant-member-init)
```
horenmar: Same as the other lint here, is it still needed after a code cleanup?

src/catch2/internal/catch_reporter_spec_parser.cpp
```diff
@@ -117,7 +117,7 @@ namespace Catch {
             auto kv = splitKVPair( parts[i] );
             auto key = kv.key, value = kv.value;
 
-            if ( key.empty() || value.empty() ) {
+            if ( key.empty() || value.empty() ) { // NOLINT(bugprone-branch-clone)
```
horenmar, line 120: What's supposed to be duplicated here?

tests/SelfTest/UsageTests/Matchers.tests.cpp
mjerabek: Deleting the call would trigger a different warning :smile: Calling the copy constructor _might_ work, though. I just find it all mightily stupid.

src/catch2/internal/catch_reporter_spec_parser.cpp
mjerabek: The `else` branch waaay down.

tests/SelfTest/UsageTests/Misc.tests.cpp
mjerabek: Done.

tests/SelfTest/UsageTests/Misc.tests.cpp
mjerabek: Done.

tests/SelfTest/UsageTests/Message.tests.cpp
mjerabek: Added `<< std::flush`.

tests/SelfTest/UsageTests/Misc.tests.cpp
mjerabek: Added `<< std::flush`.

tests/SelfTest/UsageTests/Matchers.tests.cpp
mjerabek: Calling the copy constructor worked. I have removed the suppressions.

tests/SelfTest/UsageTests/Matchers.tests.cpp
mjerabek: Same here.

tests/SelfTest/UsageTests/ToStringGeneral.tests.cpp
mjerabek: Dropped.

src/catch2/internal/catch_reporter_spec_parser.cpp
horenmar: ugh, that's dumb

mjerabek: Very :smile:

tests/SelfTest/UsageTests/Misc.tests.cpp
mjerabek: Done.

PR #2821, opened 2024-02-21
fix double promotion in catch_approx.cpp

### LINKED ISSUE #2820: unnecessary double promotion generate warning.
**Describe the bug**
Small MPUs have limited floating point units, for this reason we avoid double promotion.
If we try to compile Catch2 in our build pipeline with -Werror and -Wdouble-promotion, the build will fail.

**Expected behavior**

Catch2 should compile without warnings, helpful in the embedded world. 

**Reproduction steps**

Compile Catch with GCC 12.3.0:

```text
/home/sven/work/Universal-LED-Driver/build-presets/EmulationDev/_deps/catch2-src/src/catch2/catch_approx.cpp: In constructor ‘Catch::Approx::Approx(double)’:
/home/sven/work/Universal-LED-Driver/build-presets/EmulationDev/_deps/catch2-src/src/catch2/catch_approx.cpp:28:57: warning: implicit conversion from ‘float’ to ‘double’ to match other operand of binary expression [-Wdouble-promotion]
   28 |     :   m_epsilon( std::numeric_limits<float>::epsilon()*100. ),
```

**Platform information:**

 - OS: **Linux**
 - Compiler+version: **GCC v12.3.0**
 - Catch version: **v3.4.0**

SvenJo:
## Description
Fix double promotion in catch_approx.cpp

## GitHub Issues

Closes #2820

### DIFF
--- a/src/catch2/catch_approx.cpp
+++ b/src/catch2/catch_approx.cpp
@@ -25,7 +25,7 @@ bool marginComparison(double lhs, double rhs, double margin) {
 namespace Catch {
 
     Approx::Approx ( double value )
-    :   m_epsilon( std::numeric_limits<float>::epsilon()*100. ),
+    :   m_epsilon( static_cast<double>(std::numeric_limits<float>::epsilon())*100. ),
         m_margin( 0.0 ),
         m_scale( 0.0 ),
         m_value( value )


horenmar:
Sure, why not.

Issue #2792, opened 2024-01-09
Compiler problem with operator and include catch_all.hpp


rsc-vitronic:
**Description**
There is a compiler error, when `#include <catch2/catch_all.hpp>` is included. 

Error msg is: ` C2039: "value": is not an element of "Catch::Detail::is_eq_comparable<LhsT,std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,void>"`

Simple example when the error occurs:
```cpp
#include <catch2/catch_all.hpp>

TEST_CASE("first") {
  REQUIRE(true);
}
````
Using the include `catch2/catch_test_macros.hpp`  instead of `catch2/catch_all.hpp` all is fine

**Reproduction steps**
Including `catch2/catch_all.hpp` in one of the catch internal tests (e.g. String.tests.cpp) reproduce this error too.

**Platform information:**
 - OS: **Windows 10**
 - Compiler+version: **Visual Studio 2017 15.9.16**
 - SDK-Version: *10.0.17763.0*
 - Catch version: **v3.5.0** and **v3.5.1**

Using **v3.4.0** all is fine.

rsc-vitronic:
The error can be reproduced, 
````cpp

#include <catch2/catch_test_macros.hpp>



#include <catch2/benchmark/catch_benchmark_all.hpp>
#include <catch2/catch_approx.hpp>
#include <catch2/catch_assertion_info.hpp>
#include <catch2/catch_assertion_result.hpp>
#include <catch2/catch_config.hpp>
#include <catch2/catch_get_random_seed.hpp>
#include <catch2/catch_message.hpp>
#include <catch2/catch_section_info.hpp>
#include <catch2/catch_session.hpp>
#include <catch2/catch_tag_alias.hpp>
#include <catch2/catch_tag_alias_autoregistrar.hpp>
#include <catch2/catch_template_test_macros.hpp>
#include <catch2/catch_test_case_info.hpp>
#include <catch2/catch_test_macros.hpp>
#include <catch2/catch_test_spec.hpp>
#include <catch2/catch_timer.hpp>
#include <catch2/catch_tostring.hpp>
#include <catch2/catch_totals.hpp>
#include <catch2/catch_translate_exception.hpp>
#include <catch2/catch_version.hpp>
#include <catch2/catch_version_macros.hpp>


// #include <catch2/generators/catch_generators_all.hpp>
#include <catch2/generators/catch_generator_exception.hpp>
[...]

PR #2822, opened 2024-02-23
Compiler compatibility for Visual Studio 2017 (v141) 

rsc-vitronic:
## Description
The problem with (and only with)  Visual Studio 2017 was that the `friend bool operator==` could not sufficient resolved / matched for a template class. In this case it was caused by the  `ExtendedMultResult`. 
An detailed description of the porblem and the boiled down problem can be found in the issue #2792.

As Martin mentioned in the issue #2792: Make the operator not-friend is a sufficient solution.

## GitHub Issues
Compiler problem with operator and include catch_all.hpp #2792

### DIFF
--- a/src/catch2/internal/catch_random_integer_helpers.hpp
+++ b/src/catch2/internal/catch_random_integer_helpers.hpp
@@ -41,9 +41,8 @@ namespace Catch {
         struct ExtendedMultResult {
             T upper;
             T lower;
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks I guess, I would happily leave VS 2017 unsupported. 😃

dayfixer:
Thanks a lot!!

PR #2823, opened 2024-02-25
avoid useless case (FDuration pointer deref to FDuration=

SvenJo:
## Description

Useless cast from FDuration to FDuration in `catch_analyse.cpp`

Found with the GCC warning: -Wuseless-cast

* Parameter: `FDuration* first`
* `auto it = first;` // it is also `FDuration*`
* `std::vector<FDuration> samples;`
* `samples.push_back(FDuration(*it));` push_back expects `FDuration`, so no cast needed.

same for `mean += FDuration(*it);`: `mean` is defined as `FDuration mean = FDuration(0);`, so `mean += *it;` should be fine.


horenmar:
Thanks

PR #2836, opened 2024-03-11
Fix build with gcc 5.4

cmorve-te:
The Core Guidelines state "A base class destructor should be either public and virtual, or protected and non-virtual" so, hopefully, no static analyser will complain.


horenmar:
Sure, let's try this.

PR #2845, opened 2024-03-13
Fix float reproducibility tests on Apple Silicon chips

ChrisThrasher:
## Description

These only need to be disabled on select x86 chips but are fine to run on all Apple CPUs.

### DIFF
--- a/tests/SelfTest/IntrospectiveTests/RandomNumberGeneration.tests.cpp
+++ b/tests/SelfTest/IntrospectiveTests/RandomNumberGeneration.tests.cpp
@@ -558,7 +558,7 @@ namespace {
 // into smaller type when the type is saved into memory. This obviously
 // leads to a different answer, than doing the math in the correct precision.
 #if ( defined( _MSC_VER ) && _M_IX86_FP < 2 ) || \
-    ( defined( __GNUC__ ) && !defined( __SSE2_MATH__ ) )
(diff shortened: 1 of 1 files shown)


tests/SelfTest/IntrospectiveTests/RandomNumberGeneration.tests.cpp
```diff
@@ -558,7 +558,7 @@ namespace {
 // into smaller type when the type is saved into memory. This obviously
 // leads to a different answer, than doing the math in the correct precision.
 #if ( defined( _MSC_VER ) && _M_IX86_FP < 2 ) || \
-    ( defined( __GNUC__ ) && !defined( __SSE2_MATH__ ) )
+    ( defined( __GNUC__ ) && (defined( __i386__ ) || defined( __x86_64__ )) && !defined( __SSE2_MATH__ ) )
```
horenmar, line 561: Note that this does not actually _fix_ the tests, this just enables them on M1 machines.

Secondary issue is that the original warning will still be present on platforms that correctly disable the tests.

Full fix is to keep this updated define, but also to 
 1) move the `ifdef` above the block defining all the `uniform_fp_test_params`
 2) Change the change the tests "uniform_floating_point_distribution can handle unitary ranges" not to rely on `uniform_fp_test_params`.

PR #2850, opened 2024-03-29
Mention SUCCEED along with FAIL in logging.md

jeremy-rifkin:
## Description

I found it confusing that logging.md mentions `FAIL` but not `SUCCEED`, this probably wasn't intentional.

## GitHub Issues
N/A


horenmar:
Definitely unintentional -> `SUCCEED` just comes up less often than `FAIL`.

PR #2855, opened 2024-04-15
Handle active Sections for fatal errors

### LINKED ISSUE #1210: virtual void Catch::CumulativeReporterBase::testCaseEnded(const Catch::TestCaseStats&):  Assertion `m_sectionStack.size() == 0' failed
## Description
I am getting this error sometimes in CI.
```
testd: /build/agent-2/FAL-BS-JOB1/third-party/catch-1.9.3/catch.hpp:9611: 
virtual void Catch::CumulativeReporterBase::testCaseEnded(const Catch::TestCaseStats&): 
Assertion `m_sectionStack.size() == 0' failed.
```
Rerunning the job fixes the problem.  

This was mentioned in a comment to #663.

### Extra information
* Catch command line options: **-r junit -o testresults.xml**
* Catch version: **v1.9.3**
* Operating System: **RHEL 4.9**
* Compiler+version: **g++ 5.3**

hbirler:
Closes #1210

When a signal is caught, the destructors of Sections will not be called. Thus, we must call `sectionEndedEarly` manually for those Sections.

Example test case:
```
TEST_CASE("broken") {
   SECTION("section") {
      /// Use illegal cpu instruction
      __asm__ __volatile__("ud2" : : : "memory");
   }
}
```

### DIFF
--- a/src/catch2/internal/catch_run_context.cpp
+++ b/src/catch2/internal/catch_run_context.cpp
@@ -450,6 +450,13 @@ namespace Catch {
         assertionEnded(CATCH_MOVE(result) );
         resetAssertionInfo();
 
+        // Best effort cleanup for sections that have not been destructed yet
+        // Since this is a fatal error, we have not had and won't have the opportunity to destruct them properly
+        while (!m_activeSections.empty()) {
+            auto nl = m_activeSections.back()->nameAndLocation();
+            SectionEndInfo endInfo{ SectionInfo(CATCH_MOVE(nl.location), CATCH_MOVE(nl.name)), {}, 0.0 };
+            sectionEndedEarly(CATCH_MOVE(endInfo));
+        }
         handleUnfinishedSections();
 
         // Recreate section for test case (as we will lose the one that was in scope)
--- a/tests/ExtraTests/CMakeLists.txt
+++ b/tests/ExtraTests/CMakeLists.txt
@@ -467,6 +467,18 @@ set_tests_properties(
     PASS_REGULAR_EXPRESSION "Errors occurred during startup!"
 )
 
+add_executable(ReportingCrashWithJunitReporter ${TESTS_DIR}/X36-ReportingCrashWithJunitReporter.cpp)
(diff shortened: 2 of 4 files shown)


hbirler:
I am not familiar with the internals of Catch2, so there might be a nicer way of doing this. For example, I am not sure whether it is possible to properly construct `SectionEndInfo endInfo` with correct `prevAssertions` and `durationInSeconds`

hbirler:
Additionally, it might be safe to do:
```
auto nl = CATCH_MOVE(m_activeSections.back()->nameAndLocation());
```
instead of a copy, but I took the "safe" route (allocating memory might also not be the best idea within a signal handler).

hbirler:
The function `RunContext::handleFatalErrorCondition` seems to not be covered at all by the existing tests, and I wouldn't know how to test signal handlers reliably within the existing framework.

src/catch2/internal/catch_run_context.cpp
```diff
@@ -450,6 +450,13 @@ namespace Catch {
         assertionEnded(CATCH_MOVE(result) );
         resetAssertionInfo();
 
+        // Best effort cleanup for sections that have not been destructed yet
+        // Since this is a fatal error, we have not had and won't have the opportunity to destruct them properly
+        while (!m_activeSections.empty()) {
+            auto nl = m_activeSections.back()->nameAndLocation();
+            SectionEndInfo endInfo{ SectionInfo(CATCH_MOVE(nl.location), CATCH_MOVE(nl.name)), {}, 0.0 };
+            sectionEndedEarly(CATCH_MOVE(endInfo));
+        }
```
horenmar, line 459: I added a test that confirms that with this change JUnit does not hard crash when handling a signal.

I'll have to think about the change a bit more before merging them though.

horenmar: Yeah, this seems fine. Not moving the nameAndLocation is probably better. It _should_ be fine, but also something that could easily cause issues in the future, after some refactoring.

horenmar:
Thanks, this has been around for a while.

PR #2857, opened 2024-04-16
test: add bazel build rule for SelfTest

c8ef:
## Description

This PR primarily accomplishes two tasks:

1. It adds `MODULE.bazel.lock` to the `.gitignore` file.
2. It includes a Bazel build rule for SelfTest, which can be utilized with the command `bazel test //tests:catch2_self_test`.


horenmar:
What's the advantage of having SelfTest build rule in repo, when the main development is done using CMake?

c8ef:
There may not be much advantage, but I noticed the meson build file for the test case. I believe it would be beneficial to have a corresponding Bazel build rule for those using Bazel.

horenmar:
Fair enough.

c8ef:
Thanks!

PR #2858, opened 2024-04-16
Fix build on ARM64EC

AgostonSzepessy:
Fixes build break due to `#pragma intrinsic( _umul128 )` on ARM64EC. This intrinsic only exists on x64. However, ARM64EC builds also define `_M_X64` so the `#pragma intrinsic` needs to be excluded on those.


horenmar:
I played around with the pragma a bit on compiler explorer, and I think that the best solution is to just remove it completely.

AgostonSzepessy:
I removed the pragma bit.

Issue #2605, opened 2022-12-27
unresolved external symbol when using a string_view inside a REQUIRE


famastefano:
**Describe the bug**
If I try to link my tests that contain a REQUIRE clause with a string_view, I got this linker error:
```
LNK2019	unresolved external symbol "public: static class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl Catch::StringMaker<class std::basic_string_view<char,struct std::char_traits<char> >,void>::convert(class std::basic_string_view<char,struct std::char_traits<char> >)" (?convert@?$StringMaker@V?$basic_string_view@DU?$char_traits@D@std@@@std@@X@Catch@@SA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@V?$basic_string_view@DU?$char_traits@D@std@@@4@@Z) referenced in function "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl Catch::Detail::stringify<class std::basic_string_view<char,struct std::char_traits<char> > >(class std::basic_string_view<char,struct std::char_traits<char> > const &)" (??$stringify@V?$basic_string_view@DU?$char_traits@D@std@@@std@@@Detail@Catch@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEBV?$basic_string_view@DU?$char_traits@D@std@@@3@@Z)
```

**Expected behavior**
I expect to link without problems.

**Reproduction steps**
```cpp
std::string_view view = "test";
std::string string = "test";
REQUIRE(view == string);
```

**Platform information:**

 - OS: **Windows 10 Pro 19045.2364**
 - Compiler+version:
**Microsoft Visual Studio Professional 2022 (64-bit) - LTSC 17.2 Version 17.2.10**
**MSCV 19.32.31335**
 - Catch version: **v3.2.1** (via Conan Center)

**Additional context**
I don't know if it's Conan or something that happened to Catch, but my tests didn't have this problem months ago.
Now that I've migrated from vcpkg to Conan, this problem showed up.
At the moment I don't have time to provide a minimum example nor to check a previous version of Catch.
I'll try in my free time after work, but it means a lot of hours :P

famastefano:
Added a minimum reproducible example:
[catch-issue-2605.zip](https://github.com/catchorg/Catch2/files/10319712/catch-issue-2605.zip)

To compile:
```bash
> conan install . --build=missing
[...]

PR #2860, opened 2024-04-17
conanfile: set compatibility_cppstd = False.

vsaulue:
## Description

The Catch libraries have different API/ABI depending on the C++ standard they are compiled with. For example, [StringMaker::convert(std::string_view)](https://github.com/catchorg/Catch2/blob/53ddf37af4488cac7724761858ae3cca9d2d65e7/src/catch2/catch_tostring.cpp#L129) isn't in the binary when compiled with C++14, only with C++17 or later.

By default, Conan is allowed to serve Catch libraries compiled in C++14 to a project using C++17/20, potentially causing linker errors because of missing symbols (see https://github.com/catchorg/Catch2/issues/2605). This PR overrides this default using [conan extension properties](https://docs.conan.io/2/reference/conanfile/attributes.html#extension-properties): the C++ standard used to build the Catch library will exactly match the one of the requiring project (building Catch from source if necessary).

This is probably much more constraining than necessary (I assume it will prevent conan from using a `c++17` catch2 binary into a `gnu17` project). But that's the only simple solution I have. And as a user, locally rebuilding Catch "just in case" is much better than getting linker errors.

## Note:

Right now for Windows, the [conan-center catch2 package](https://conan.io/center/recipes/catch2) has only prebuilt binaries for C++14. So many Windows users are at risk of getting served C++14 Catch2 binaries in all their non-C++14 projects.

This can also happen on other platforms: if the first Catch2 binaries that ends up in your Conan cache were build in C++14, Conan will currently serve them in all your future C++17/20 projects.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,6 +19,7 @@ class CatchConan(ConanFile):
     license = "BSL-1.0"
     version = "latest"
     settings = "os", "compiler", "build_type", "arch"
+    extension_properties = {"compatibility_cppstd": False}
 
     options = {
         "shared": [True, False],


horenmar:
Thanks. This is an interesting feature, vcpkg does not have similar one -- their workaround is to always build the impl lib with C++17.

PR #2861, opened 2024-04-18
conanfile: fix cmake_target_name of Catch2::Catch2.

vsaulue:
## Description

The "Catch2 without default main" target doesn't have a specific name in Conan, and defaults to catch2::catch2base. This commit switches it back to Catch2::Catch2, as described in the docs.


horenmar:
Thanks

PR #2862, opened 2024-04-23
Fix clang-tidy `bugprone-chained-comparison` warnings on v2.x

rikyoz:
This PR is just a backport of the https://github.com/catchorg/Catch2/pull/2801 PR on the v2.x branch.
It suppresses the `bugprone-chained-comparison` warning issued by recent versions of clang-tidy.


horenmar:
Out of curiosity, why are you still using v2?

rikyoz:
Thank you for the merge!

> Out of curiosity, why are you still using v2?

Mainly because I still need to support old compilers like GCC 4.9 or MSVC 2015 in some of the projects where I use Catch2.
In one of these projects, I actually upgraded to v3 at some point, but then had to go back to v2 for this reason.
I'll definitely upgrade to v3 as soon as I can raise the minimum requirements for them.

PR #2867, opened 2024-05-01
Move header from internal

oboukli:
Move `catch_case_sensitive.hpp` from directory
`src/catch2/internal/internal` to be directly under `src/catch2/`.

Header `catch_case_sensitive.hpp` defines `enum CaseSensitive` which is part of the public API, since it is exposed by the string matcher contracts.

Fixes issue caught by Clang Tidy misc-include-cleaner check.


horenmar:
Thanks, makes sense.

PR #2875, opened 2024-05-23
Explicitly cast values of different types

svenihoney:
## Description
In case the warning -Werror=conversion is active with GCC, the warnings about "conversion from A to B may change value" lead to a compilation error. This explicitly convert the values to address these warnings.


horenmar:
Eh, sure.

PR #2880, opened 2024-06-28
Add optional argument to catch_discover_tests to set DYLD_FRAMEWORK_PATH

andy-phillips:
If a 3rd party library used in tests depends on the DYLD_FRAMEWORK_PATH environment variable to be set on OSX (in order to find its dependent libraries packaged as Apple frameworks during test discovery and run them), then catch_discover_tests will fail. It's similar to the need and usage of the DL_PATHS argument.

Two solutions I initially thought of. One, set the DYLD_FRAMEWORK_PATH implicitly to the same value as DL_PATHS. Second, adding an explicit optional argument to set the paths specifically used for Apple frameworks.

In this PR I have opted for the second solution to allow the two variables to hold different values, even though DYLD_FRAMEWORK_PATH only needs to be set on OSX. There is some repetition in the code to process the paths and set the variables which could be improved in a separate commit or PR.


horenmar:
It's fine to keep the paths duplicated.

extras/Catch.cmake
```diff
@@ -124,6 +124,13 @@ same as the Catch name; see also ``TEST_PREFIX`` and ``TEST_SUFFIX``.
     test executable and when the tests are executed themselves. This requires
     cmake/ctest >= 3.22.
 
+  ``DL_FRAMEWORK_PATHS path...``
+    Specifies paths that need to be set for the dynamic linker to find libraries
+    packaged as frameworks on Apple platforms when running the test executable
+    (DYLD_FRAMEWORK_PATH). These paths will both be set when retrieving the list
+    of test cases from the test executable and when the tests are executed themselves.
+    This requires cmake/ctest >= 3.22.
```
horenmar, line 132: On line 166, the script explicitly errors out if `_DL_PATHS` are used with CMake version before 3.22. `_DL_FRAMEWORK_PATHS` should be checked there as well.

andy-phillips: Thanks for pointing that out, check added.

horenmar:
Thanks

PR #2884, opened 2024-07-07
Add parens around std::min and std::max to fix compilation on Windows.

wilx:
## Description
Compilation on Windows fails do to unparenthesized `std::min` and `std::max`.  This change adds the parentheses.


horenmar:
No, don't leak `min` and `max` macros from `windows.h` into libraries.

wilx:
> No, don't leak `min` and `max` macros from `windows.h` into libraries.

That is impractical. Also, this is not new, this has been done before for Catch2: ea44e73
Please reconsider.

horenmar:
I have since then decided that I am not willing to spend the effort on working around people using `windows.h` without `NOMINMAX`.

Issue #1602, opened 2019-04-17
Run a piece of setup once only


bna080:
We are writing unit tests where for some test cases we need to run a specific part of the setup and tear-down (say one line of code) once before first section is run and once after all the sections in the test case are done. We would like to do it this way because this specific part of setup and tear-down takes a long time to execute which slows down the total test time. 
Is there a way how to achieve this with Catch? I know it sort of breaks the Catch logic, but I have not found a clean way how to do it looking at the Catch Reference and running the "slow" command before and after every section execution just takes too much time.

horenmar:
You should be able to use Event Listeners, specifically the `testCaseStarting` and `testCaseEnded` hooks.

This has the disadvantage of putting the init and uninit code outside of the main test case, but the advantage of letting you do exactly what you want.

bna080:
Thank you, this almost works for me. However, in the testCaseStarting/testCaseEnded handlers I need to perform operations on an object which is created in the test case itself and which is also accessed and used in the sections of the test case. Is there a way to access such an object, or pass a reference to it to the handlers?

horenmar:
There is not, the events are called before the test case is first run (and thus before such object exists) and after the test case is last run (and thus the object is dead at that point). This means that you will also need to introduce some global state.

Alternatively, if the destruction part is not overly important (e.g. you are fine with the destruction happening after the end of _all_ the tests) and the init is basically a constructor, you could abuse a static variable, which will be initialized exactly once and will be destructed at the end of the program.

bna080:
Ah, of course you are right, my object does not exist at that point. 
I have been using a static variable until now and it works fine, but I need to do multiple operations on it once before all the tests and once after all the tests. That I have now in a special section which always has to be the first one and another one which is the last one. Not ideal, but it works. Anyway, I will look into using the listeners with a global station as you suggested.

ficzerepeti:
I'd like to add that I'm also interested in solving the described issue in a nicer way than using static variables or writing custom code listening to events.

I thought a SetupBeforeFirstRun and TearDownAfterLastRun macros could be introduced (maybe with better names).

```
TEST_CASE("This is a test", "[NiceTag]")
{
    static std::optional<MyVeryTimeConsumingObject> objectToSetup;

    SetupBeforeFirstRun
    {
        ... // do some stuff
        objectToSetup.emplace(param1, param2);
        ... // do more stuff
    }

    .... // test things
[...]

PR #2885, opened 2024-07-07
TEST_CASE_FIXTURE: A new fixture macro for allowing persistent fixtures throughout a TEST_CASE

KStocky:
## Description

This PR introduces a new `TEST_CASE` macro called `TEST_CASE_FIXTURE`. `TEST_CASE_FIXTURE` offers the same functionality as `TEST_CASE_METHOD` except for one difference. The object on which the test method is invoked is only created once for all invocations of the test case. The object is created just after the `testCaseStarting` event is broadcast and the object is destroyed just before the `testCaseEnding` event is broadcast.

The main motivation for this new functionality is to allow `TEST_CASE`s to do expensive setup and teardown once per `TEST_CASE`, without having to resort to abusing event listeners or static function variables with manual initialization.

## GitHub Issues

This PR hopes to be an implementation of #1602


docs/cmake-integration.md
KStocky, line 1: This was done when I was generating ToC after I had added my documentation to test-fixtures.md. I can remove it if this should be done in another PR since it isn't exactly to do with this PR at all.

docs/other-macros.md
KStocky, line 1: I have removed this section and placed it in with the test-fixtures documentation since it is a macro for creating a test fixture.

docs/test-fixtures.md
KStocky, line 1: I have reorganised this file to better document all of the fixture macros in one place and to compare and contrast them with each other.

examples/CMakeLists.txt
KStocky, line 1: I couldn't find out what the naming scheme of examples is... Happy to rename this once I know what the name should be!

src/catch2/catch_test_macros.hpp
```diff
@@ -195,6 +198,7 @@
   #define TEST_CASE( ... )  INTERNAL_CATCH_TESTCASE_NO_REGISTRATION(INTERNAL_CATCH_UNIQUE_NAME( CATCH2_INTERNAL_TEST_ ), __VA_ARGS__)
   #define TEST_CASE_METHOD( className, ... ) INTERNAL_CATCH_TESTCASE_NO_REGISTRATION(INTERNAL_CATCH_UNIQUE_NAME( CATCH2_INTERNAL_TEST_ ))
   #define METHOD_AS_TEST_CASE( method, ... )
+  #define TEST_CASE_FIXTURE( className, ... )
```
KStocky: I just copied what TEST_CASE_METHOD does in this case. It would seem that these are purposefully unimplemented

src/catch2/internal/catch_test_registry.hpp
```diff
@@ -47,6 +47,33 @@ Detail::unique_ptr<ITestInvoker> makeTestInvoker( void (C::*testAsMethod)() ) {
...
+    void testCaseStarting() override { 
+        m_fixture = Detail::make_unique<C>(); 
+    }
+
+    void testCaseEnding() override { 
+        m_fixture.reset(); 
+    }
+
+    void invoke() const override {
+        auto* f = const_cast<C*>( m_fixture.get() );
```
KStocky: Just want to draw attention to this because const_cast is usually a sin and UB. However, this is a valid use-case of it that does not invoke UB since m_fixture does not point to a const object. This const_cast is required due to how Detail::unique_ptr implements get(). It differs from std::unique_ptr since it will return a pointer to const if the unique_ptr itself is const. This is not what std::unique_ptr does. https://en.cppreference.com/w/cpp/memory/unique_ptr/get.

The other option was to manually manage a raw pointer and add special member functions to ensure that it is handled correctly, but I have erred with const_cast in this case. Happy to change this though

src/catch2/internal/catch_test_registry.hpp
```diff
@@ -47,6 +47,33 @@ Detail::unique_ptr<ITestInvoker> makeTestInvoker( void (C::*testAsMethod)() ) {
...
+        ( f->*m_testAsMethod )();
+    }
+};
+
+template<typename C>
+Detail::unique_ptr<ITestInvoker> makeTestInvokerFixture( void ( C::*testAsMethod )() ) {
```
KStocky: I could not make a `makeTestInvoker` overload since it would have the same function signature as the factory function for TEST_CASE_METHOD.

src/catch2/catch_test_macros.hpp
```diff
@@ -142,6 +144,7 @@
   #define TEST_CASE( ... ) INTERNAL_CATCH_TESTCASE( __VA_ARGS__ )
   #define TEST_CASE_METHOD( className, ... ) INTERNAL_CATCH_TEST_CASE_METHOD( className, __VA_ARGS__ )
   #define METHOD_AS_TEST_CASE( method, ... ) INTERNAL_CATCH_METHOD_AS_TEST_CASE( method, __VA_ARGS__ )
+  #define TEST_CASE_FIXTURE( className, ... ) INTERNAL_CATCH_TEST_CASE_FIXTURE( className, __VA_ARGS__ )
```
KStocky: I am not tied to this name in any way. I think it is an OK name because this is the typical behaviour of a fixture in other testing frameworks i.e. the fixture objects is the same fixture object for all test cases that use it. 

I toyed with TEST_CASE_PERSISTENT_FIXTURE and TEST_CASE_WITH_SETUP among other names but they seemed to be a bit too long... In the end, I decided to just go with this name for now and if anyone has a better suggestion, I am more than happy to go with that.

docs/test-fixtures.md
```diff
@@ -30,8 +47,59 @@ class UniqueTestsFixture {
...
+
+
+METHOD_AS_TEST_CASE( TestClass::testCase, "Use class's method as a test case", "[class]" )
+```
+
+This type of fixture is similar to [TEST_CASE_METHOD](#1-test_case_method) except in this case it will directly use the provided class to create an object rather than a derived class.
+
+### 3. `TEST_CASE_FIXTURE`
+
+> [Introduced](link-to-issue-or-PR) in Catch2 X.Y.Z
```
KStocky: I think this is the correct way to introduce a version placeholder...

docs/test-fixtures.md
```diff
@@ -30,8 +47,59 @@ class UniqueTestsFixture {
  }
 ```
 
-The two test cases here will create uniquely-named derived classes of UniqueTestsFixture and thus can access the `getID()` protected method and `conn` member variables. This ensures that both the test cases are able to create a DBConnection using the same method (DRY principle) and that any ID's created are unique such that the order that tests are executed does not matter.
+The two test cases here will create uniquely-named derived classes of UniqueTestsFixture and thus can access the `getID()` protected method and `conn` member variables. This ensures that both the test cases are able to create a DBConnection using the same method (DRY principle) and that any ID's created are unique such that the order that tests are executed does not matter. 
+
+### 2. `METHOD_AS_TEST_CASE`
```
KStocky, line 61: Cut and pasted from other-macros.md

docs/test-fixtures.md
horenmar: The suffix is correct (the release scripts will pick it up and update to the new version), but the target has to be done manually, so `https://github.com/catchorg/Catch2/pull/2885` instead of the `link-to-issue-or-PR`.

docs/test-fixtures.md
```diff
@@ -1,9 +1,26 @@
...
+**Contents**<br>
+[Non-Templated test fixtures](#non-templated-test-fixtures)<br>
+[Templated test fixtures](#templated-test-fixtures)<br>
+[Signature-based parameterised test fixtures](#signature-based-parametrised-test-fixtures)<br>
+[Template fixtures with types specified in template type lists](#template-fixtures-with-types-specified-in-template-type-lists)<br>
 
-Although Catch allows you to group tests together as [sections within a test case](test-cases-and-sections.md), it can still be convenient, sometimes, to group them using a more traditional test fixture. Catch fully supports this too. You define the test fixture as a simple structure:
+## Non-Templated test fixtures
+
+Although Catch allows you to group tests together as [sections within a test case](test-cases-and-sections.md), it can still be convenient, sometimes, to group them using a more traditional test fixture. Catch fully supports this too with 3 different macros for non-templated test fixtures. They are:
```
horenmar: Formatting nitpick first: new docs use linebreaks at 80 chars (well, more like 72, you know how it is).

New docs also use Catch2 to refers to the project. (The reason I don't mass update this in the old docs is that it serves as a good heuristic on how old the docs are, and how they will be written).

docs/test-fixtures.md
```diff
@@ -30,8 +47,59 @@ class UniqueTestsFixture {
...
+    SECTION("First partial run") {
+        REQUIRE(val == 0);
+    }
+
+    SECTION("Second partial run") {
+        REQUIRE(val == 1);
+    }
+}
+```
+This test case will be executed twice as there are two leaf sections. On the first run `val` will be `0` and on the second run `val` will be `1`. This is useful if you would like to share some expensive setup code with all runs of your test case which can't be done at static initialization time.
```
horenmar: I am not particularly happy about the example/explanation combo. My first reaction was "well, how's that different from using static variable here".

I can't give you better version right now, but I think this should be focused on the performance difference from `TEST_CASE_METHOD`, maybe have something like

```cpp
// Pretend doing hard work
std::this_thread::sleep_for(2s);
```

in the fixture constructor and explain that `TEST_CASE_FIXTURE` takes 2 seconds, while `TEST_CASE_METHOD` needs 2/4/6/8/... seconds to execute.

src/catch2/internal/catch_test_registry.hpp
horenmar: The implementation difference is intentional, the shallow const on language pointers makes it annoyingly easy to throw away logical constness. So the const cast here is fine-ish.

Here is a thought though: why not make the fixture member function const?

It would be inconsistent with the `TEST_CASE_METHOD` macro, but this is fine; inconsistency can be okay if it brings improvement. The advantage is that it would be much harder to accidentally mutate the fixture -> this is much more important here than for `TEST_CASE_METHOD`, as the fixture here is persistent between test runs. It is still possible to allow fixture mutation, by marking the intended-to-change member variable with `mutable`.

I actually took your branch and tried a quick sketch of this and I think it is better, but I am willing to hear counter arguments.

horenmar:
As an FYI, I am (rather obviously) not finished with the review, but my bus ride is over and I'll have to return to this later.

src/catch2/internal/catch_test_registry.hpp
KStocky: I have zero issues with this honestly. I agree that this would be a better course of action and would make test writers  think about what they want to be mutatable. Const as the default is never usually a bad thing. I will make this change, and also note it in the docs, and perhaps also make a example to demonstrate it. Thanks for the great improvement!

docs/test-fixtures.md
```diff
@@ -30,8 +47,59 @@ class UniqueTestsFixture {
...
+This type of fixture is similar to [TEST_CASE_METHOD](#1-test_case_method) except in this case it will directly use the provided class to create an object rather than a derived class.
+
+### 3. `TEST_CASE_FIXTURE`
+
+> [Introduced](link-to-issue-or-PR) in Catch2 X.Y.Z
+
+`TEST_CASE_FIXTURE` behaves in the same way as [TEST_CASE_METHOD](#1-test_case_method) except that there will only be one instance created throughout the entire run of a test case. To demonstrate this have a look at the following example:
+
+```cpp
+struct MyFixture{
```
horenmar: Also a formatting nitpick: space.
```suggestion
struct MyFixture {
```

src/catch2/catch_test_case_info.hpp
```diff
@@ -112,6 +112,14 @@ namespace Catch {
         TestCaseHandle(TestCaseInfo* info, ITestInvoker* invoker) :
             m_info(info), m_invoker(invoker) {}
 
+        void testCaseStarting() const {
```
horenmar: Not a big fan of this name. It is consistent with the event names, but it does not describe the reason why it exists well. How about "prepareTestCase" instead?

(And same for `testCaseEnding` below.)

src/catch2/catch_test_macros.hpp
horenmar: > because this is the typical behaviour of a fixture in other testing frameworks i.e. the fixture objects is the same fixture object for all test cases that use it.

This is not true. The most used C++ test framework is GTest, which creates separate fixture instance per test. Catch2's current support for this does the same, and thus also Doctest and other clones. If I remember CppUnit correctly, it also creates a separate instance of fixture per test.

Now you might argue that none of these is a good match, because only Catch2 and its derivatives support defining multiple partial test cases in single test case body, and for GTest and CppUnit we are talking about different test case bodies, but I think of Catch2's test case trees as a short hand for defining multiple test cases, which would make them equivalent.

I am not saying that `TEST_CASE_FIXTURE` is the wrong name, just disputing the assertion about the typical behaviour of a (C++) test framework.

----------------

If I am reading the xUnit docs correctly, it also creates a separate fixture instance per test. Same for Python's unittest.

----------------

Personally I slightly prefer `TEST_CASE_PERSISTENT_FIXTURE`.

src/catch2/catch_test_macros.hpp
horenmar: `TEST_CASE_METHOD` does this though

```cpp
  #define TEST_CASE_METHOD( className, ... ) INTERNAL_CATCH_TESTCASE_NO_REGISTRATION(INTERNAL_CATCH_UNIQUE_NAME( CATCH2_INTERNAL_TEST_ ))
```

src/catch2/interfaces/catch_interfaces_test_invoker.hpp
```diff
@@ -12,6 +12,8 @@ namespace Catch {
 
     class ITestInvoker {
     public:
+        virtual void testCaseStarting();
+        virtual void testCaseEnding();
```
horenmar: Same note about name here.

docs/cmake-integration.md
horenmar: This is fine.

docs/test-fixtures.md
horenmar: This should probably also explain that two different `TEST_CASE_FIXTURE`s have **separate** instances of fixture, i.e., the fixture is not static globally, just for the paths defined in one `TEST_CASE`.

src/catch2/catch_test_macros.hpp
horenmar: Also please add the macro to `tests/ExtraTests/X02-DisabledMacros.cpp` to check that the compilation still works when this preprocessor branch is compiled.

horenmar (changes requested):
Overall I am convinced that the functionality is useful, but I want some changes before merging.

src/catch2/catch_test_macros.hpp
KStocky: TEST_CASE_PERSISTENT_FIXTURE it is then!

KStocky:
> Overall I am convinced that the functionality is useful, but I want some changes before merging.

Awesome. One other question, What is the naming scheme of the examples? There appears to be one but I could not figure it out. Hence why I just left my example as being called "Fixture.cpp".

src/catch2/interfaces/catch_interfaces_test_invoker.hpp
KStocky: I have named these `prepareTestCase` and `tearDownTestCase` now

src/catch2/catch_test_case_info.hpp
KStocky: Yup as below, I have renamed these prepareTestCase and tearDownTestCase

docs/test-fixtures.md
KStocky: Yup ran the formatter over this file now. I have also named it more appropriately.

docs/test-fixtures.md
KStocky: I have redone the example and the documentation for this feature now. I feel like it now better highlights why a test writer would use this over TEST_CASE_METHOD.

docs/test-fixtures.md
KStocky: I think I have addressed this now. However, I was not able to follow this rule with regard to tables in markdown because taking a line break would introduce a new row in the table... If you have any suggestions on how to solve this issue, I will be more than happy to do that :)

docs/test-fixtures.md
KStocky: Yup this was something that should have been more obvious to me... My bad. And fixed now!

src/catch2/catch_test_macros.hpp
KStocky: Yup this was what was causing the majority of the CI failures. Fixed now!

examples/111-Fix-PersistentFixture.cpp
```diff
@@ -0,0 +1,74 @@
...
+//              Copyright Catch2 Authors
+// Distributed under the Boost Software License, Version 1.0.
+//   (See accompanying file LICENSE.txt or copy at
+//        https://www.boost.org/LICENSE_1_0.txt)
+
+// SPDX-License-Identifier: BSL-1.0
+
+// Fixture.cpp
+
+// Catch has three ways to express fixtures:
```
horenmar:
```suggestion
// Catch2 has three ways to express fixtures:
```

docs/test-fixtures.md
horenmar: I don't think it is possible to do with tables in github's markdown, so it's fine to keep tables without linebreaks.

horenmar:
Thanks

PR #2902, opened 2024-09-04
Add meson option for windows seh usage

apache-hb:
## Description
Adds a meson option to toggle usage of windows SEH handling.

### DIFF
--- a/src/catch2/meson.build
+++ b/src/catch2/meson.build
@@ -10,6 +10,10 @@ conf_data = configuration_data()
 conf_data.set('CATCH_CONFIG_DEFAULT_REPORTER', 'console')
 conf_data.set('CATCH_CONFIG_CONSOLE_WIDTH', '80')
 
+windows_seh = get_option('windows_seh')
+conf_data.set('CATCH_CONFIG_WINDOWS_SEH', windows_seh)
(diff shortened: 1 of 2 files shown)


src/catch2/meson.build
```diff
@@ -10,6 +10,10 @@ conf_data = configuration_data()
 conf_data.set('CATCH_CONFIG_DEFAULT_REPORTER', 'console')
 conf_data.set('CATCH_CONFIG_CONSOLE_WIDTH', '80')
 
+windows_seh = get_option('windows_seh')
+conf_data.set('CATCH_CONFIG_WINDOWS_SEH', windows_seh)
+conf_data.set('CATCH_CONFIG_NO_WINDOWS_SEH', not windows_seh)
+
```
apache-hb, line 16: mesons `cfg_data.set` undefines a symbol when passed `false`, and defines it when passed `true`. The complete rules are available on [mesons docs](https://mesonbuild.com/Configuration.html).

horenmar: This does not seem to have the correct semantics.

The `CATCH_CONFIG_FOO` semantics are
`CATCH_CONFIG_FOO` - skip Catch2's autodetection, force enable FOO
`CATCH_CONFIG_NO_FOO` - skip Catch2's autodetection, force disable FOO

autodetection only happens if neither is defined

PR #2904, opened 2024-09-05
Suppress GCC useless-cast warning from CHECK_THROWS_MATCHES

laudrup:
Suppress warning from GCC about useless cast in the CHECK_THROWS_MATCHES macro the same way it is already being done for the similar CHECK macros.


laudrup:
CHECK_THROWS_MATCHES will generate a `useless-cast` warning when building with GCC. I assume this additional suppression was simply overseen when the changes 0de60d8e7ead1ddd5ba8c46b901c122eac20bf94 were added.

horenmar:
Thanks.

PR #2905, opened 2024-09-05
Fix bug in TokenStream parser

learn-more:
When presented with just '-' it would access the string at position [1]

I discovered this while mistyping the '-?' option, which triggered an assert in my test binary.

### DIFF
--- a/tests/SelfTest/IntrospectiveTests/Clara.tests.cpp
+++ b/tests/SelfTest/IntrospectiveTests/Clara.tests.cpp
@@ -51,6 +51,21 @@ TEST_CASE("Clara::Arg supports single-arg parse the way Opt does", "[clara][arg]
     REQUIRE(name == "foo");
 }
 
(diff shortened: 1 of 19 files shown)


tests/SelfTest/IntrospectiveTests/Clara.tests.cpp
```diff
@@ -51,6 +51,16 @@ TEST_CASE("Clara::Arg supports single-arg parse the way Opt does", "[clara][arg]
     REQUIRE(name == "foo");
 }
 
+TEST_CASE("Clara::Arg does not crash on incomplete input", "[clara][arg][compilation]") {
+    std::string name;
+    auto p = Catch::Clara::Arg(name, "-");
+
+    CHECK(name.empty());
+
+    p.parse( Catch::Clara::Args{ "UnitTest", "-" } );
```
horenmar: This should probably check that the parsing has failed.

learn-more: I added some extra checks and updated the Approved stuff.

PR #2918, opened 2024-10-11
Add shared library support when building with meson

Colum31:
## Description
Building shared libraries is currently not supported, when building with meson.
This PR enables building shared libraries, by passing `-Ddefault_library=shared`.
When this option is not passed, a static library is built.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -8,6 +8,7 @@
 project(
   'catch2',
   'cpp',
+  default_options: ['default_library=static'],
   version: '3.7.1', # CML version placeholder, don't delete
   license: 'BSL-1.0',
   meson_version: '>=0.54.1',
--- a/src/catch2/meson.build
+++ b/src/catch2/meson.build
@@ -351,7 +351,7 @@ if ((host_machine.system() == 'android') or
   catch2_dependencies += log_dep
 endif
 
-catch2 = static_library(
+catch2 = library(
   'Catch2',
   sources,
   dependencies: catch2_dependencies,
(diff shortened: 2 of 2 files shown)


horenmar:
Does this handle exporting symbols?

Colum31:
Yes, when running `readelf -s`  I can see exported symbols.

Colum31:
Or which symbols do you mean?

horenmar:
The important platform for this is Windows. Linux has all symbols visible by default, but Windows uses the better option, which is that all symbols are hidden by default.

CMake has [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) for this.

shahsb (approved):
LGTM!

stephanlachnit:
> The important platform for this is Windows. Linux has all symbols visible by default, but Windows uses the better option, which is that all symbols are hidden by default.
> 
> CMake has [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) for this.

There is the [`gnu_symbol_visibility`](https://mesonbuild.com/Reference-manual_functions.html#library_gnu_symbol_visibility) property in Meson which can be used to hide all symbols on Unix by default.

However, this only works when each symbol has the appropriate symbol visibility annotation, which is not the case as far as I can see. There is also no equivalent for `WINDOWS_EXPORT_ALL_SYMBOLS`, so Windows compilation would be broken by default.

PR #2919, opened 2024-10-12
Provide overloads for {Unordered}RangeEquals taking a std::initializer_list

### LINKED ISSUE #2915: RangeEquals doesn't support literal std::initializer_list
It would often be useful to write something like this:
```c++
  const auto v = calculateSomeVectorOfInts();
  CHECK_THAT(v, RangeEquals({1, 2, 3}));
```
This doesn't compile. It can be made to compile by adding an overload to RangeEquals that takes a std::initializer_list. Would a PR adding such an overload have a chance of being accepted?

It's not a huge issue since there are workarounds, including any of these:
```c++
  const auto expected = {1, 2, 3};
  CHECK_THAT(v, RangeEquals(expected));

  CHECK_THAT(v, RangeEquals(std::vector{1, 2, 3}));

  CHECK_THAT(v, RangeEquals(std::initializer_list<int>{1, 2, 3}));
```
Still, the convenience of writing the initializer_list inline without extra qualification would be nice.

stk-ableton:
This allows writing something like

```c++
  const auto v = calculateSomeVectorOfInts();
  CHECK_THAT(v, RangeEquals({1, 2, 3}));
```

Fixes #2915.

### DIFF
--- a/src/catch2/matchers/catch_matchers_range_equals.hpp
+++ b/src/catch2/matchers/catch_matchers_range_equals.hpp
@@ -110,6 +110,22 @@ namespace Catch {
 
         /**
          * Creates a matcher that checks if all elements in a range are equal
+         * to all elements in an initializer list.
+         *
+         * Uses the provided predicate `predicate` to do the comparisons
+         * (defaulting to `std::equal_to`)
+         */
+        template <typename T,
+                  typename Equality = decltype( std::equal_to<>{} )>
+        constexpr
+        RangeEqualsMatcher<std::initializer_list<T>, Equality>
+        RangeEquals( std::initializer_list<T> range,
+                     Equality&& predicate = std::equal_to<>{} ) {
+            return { range, CATCH_FORWARD( predicate ) };
+        }
+
+        /**
+         * Creates a matcher that checks if all elements in a range are equal
          * to all elements in another range, in some permutation.
          *
          * Uses the provided predicate `predicate` to do the comparisons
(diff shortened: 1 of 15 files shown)


src/catch2/matchers/catch_matchers_range_equals.hpp
```diff
@@ -134,18 +160,44 @@ namespace Catch {
...
+        /**
+         * Creates a matcher that checks if all elements in a range are equal
+         * to all elements in an initializer list, in some permutation
+         *
+         * Uses `std::equal_to` to do the comparison
+         */
+        template <typename T>
+        constexpr
+        UnorderedRangeEqualsMatcher<std::initializer_list<T>, std::equal_to<>>
+        UnorderedRangeEquals( std::initializer_list<T> range ) {
```
horenmar: I see that this copies the old overloads, but I wonder why I went with overloads instead of defaulted comparison parameter back then.

horenmar (approved):
I am going to look into the overload vs default argument a bit later, but the PR looks good.

src/catch2/matchers/catch_matchers_range_equals.hpp
stk-ableton: I wondered the same, but was reluctant to change it. Now that you mention it, I added a commit (before mine) to clean this up: 5328f90520fc. Sorry for force-pushing, but I didn't want to put a commit on top that changes it both for the existing code and the new overloads; the diffs are much cleaner this way.

horenmar:
The default args should be fine here. The usual arguments against should not apply to this use case.

Thanks for the change.

PR #2922, opened 2024-10-18
Use isatty() when using GNU libc

pinotree:
## Description

While `isatty()` is a POSIX interface and theoretically could be used more broadly than on Linux and macOS, use a conservative approach and use it on any platform that uses GNU libc.

This fixes `#1670 regression check` on GNU/Hurd (which uses GNU libc).


horenmar:
Thanks

PR #2926, opened 2024-10-29
Fix using from_range with std::vector<>::const_iterator

burlog:
This pull request addresses a compile error occurring when attempting to instantiate `std::vector` with a constant value type.

The issue arose in `catch_generators_range.hpp`, where `ResultType` was defined as a `value_type` of `const_iterator` within `GeneratorWrapper<ResultType>` which is usually `const`. This caused a conflict with `std::vector`, which requires its `value_type` to be non-const.

### DIFF
--- a/src/catch2/generators/catch_generators_range.hpp
+++ b/src/catch2/generators/catch_generators_range.hpp
@@ -91,7 +91,7 @@ public:
 
 template <typename InputIterator,
           typename InputSentinel,
-          typename ResultType = typename std::iterator_traits<InputIterator>::value_type>
+          typename ResultType = std::remove_const_t<typename std::iterator_traits<InputIterator>::value_type>>
 GeneratorWrapper<ResultType> from_range(InputIterator from, InputSentinel to) {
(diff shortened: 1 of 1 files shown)


horenmar:
Can you provide a reproducer?

burlog:
Hm, I see that `const_iterators` of the standard containers does not have const qualified `value_type`, so I don't know if my patch is relevant. On the other hand, removing const from value type here probably does not break anything.

I encounter this when I used [yaml-cpp](https://github.com/jbeder/yaml-cpp) for parsing test cases. Simplyfied example is:

```cpp
#include <catch2/generators/catch_generators_range.hpp>

#include <yaml-cpp/yaml.h>

auto yaml = YAML::LoadFile("test.yaml");

TEST_CASE("test case") {
    auto value = GENERATE(from_range(yaml));
}
```

horenmar:
Hmmmm, I am pretty sure that's yaml-cpp being wrong, I opened up an issue: https://github.com/jbeder/yaml-cpp/issues/1326

I'll have to think about whether I want to merge a workaround for now or not.

horenmar:
Just an FYI, if yaml-cpp upstream doesn't react, I plan to merge this on Friday.

PR #2936, opened 2024-11-19
Catch.cmake: Remove redundant CTEST_FILE param

hkaelber:
The `CTEST_FILE` argument was set twice. Although this still worked correctly the 2nd time setting it to empty is not required.

Thx to @craigscott-crascit for spotting this: https://github.com/catchorg/Catch2/commit/aad926baf87bb0121afc07b100e61f2b344ce156#r149159698


horenmar:
Thanks

PR #2940, opened 2024-11-30
Add meson option to not install the library

burlog:
## Description

This pull request introduces a Meson build option to allow users to disable the installation of the Catch2 library when using it as a subproject wrap.

The added `install` Meson option lets users include Catch2 as a subproject with `default_options: ['install=false']`, effectively preventing it from being installed alongside the main project or library. For example:  

```meson
catch2_with_main = subproject('catch2-with-main', default_options: ['install=false'])
```

Catch2's primary purpose is for testing, and when included as a wrap, its installation is often unnecessary. In scenarios where Catch2 is used as a testing dependency, users typically don't want it to be installed with their project.  

Currently, the Meson build system doesn't natively support disabling the installation of wraps. While Meson offers a `--skip-subprojects` flag during installation, it must be manually specified every time `meson install` is run, which can be cumbersome and error-prone.  

This change streamlines the process, improving user experience by eliminating the need for repetitive manual configuration. It aligns with the expected use cases for Catch2 when employed as a testing utility.


horenmar:
Makes sense, thanks.

Issue #2927, opened 2024-10-31
Prevent `-Weffc++` and `-W-ctor-dtor-privacy` warnings from leaking out


Adda0:
**Description**
I updated Catch2 used in my project from some old Catch2v2 version to v3. When including Catch2 in my project tests through the CMake fetcher as a dependency, compiling the project with Clang (clang 17.0.6) does not issue any warnings, but compiling with GCC (g++ 13.3.0) issues a number of `-Weffc++` and `-W-ctod-dtor-privacy` warnings.

I presume the code in Catch2 is supposed to behave as it does. Therefore, suppressing these warnings on a per-header file level should be the correct answer (or anything similar to this) to removing these warnings, while still allowing the users of Catch2 to include specific header files with the warnings set up for their tests source code.

**Steps to reproduce**
CMake snippet: 
```cmake
    if (BUILD_TESTING)
        # Catch2 library for writing tests
        FetchContent_Declare(
        Catch2
        GIT_REPOSITORY https://github.com/catchorg/Catch2.git
        GIT_TAG        v3.7.1
        )
        FetchContent_MakeAvailable(Catch2)
        list(APPEND CMAKE_MODULE_PATH ${Catch2_SOURCE_DIR}/contrib)
        include(Catch)
    endif()

target_link_libraries(tests PRIVATE <project files> Catch2::Catch2WithMain)
```
Tests source C++ file:
```cpp
#include <catch2/catch_test_macros.hpp>
#include <catch2/matchers/catch_matchers_string.hpp>

// Tests.
```

**Additional context**
Generated warnings:
- (binary) `user-defined <some class>::operator ... always evaluates both arguments`,
- `all member functions in class <some class> are private`,
- `<some class> has pointer data members ... but does not declare ...`,
- ...
Log of the build process: [warnings.txt](https://github.com/user-attachments/files/17585330/warnings.txt).

**Suggested alternative**
There are a number of options to suppress these warnings.

Manually, each user can disable the warnings for each test source file, for example by wrapping the includes of Catch2 headers in pragmas:
```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Weffc++"
#include <catch2/catch_test_macros.hpp>
#include <catch2/matchers/catch_matchers_string.hpp>
#pragma GCC diagnostic pop
```

I also managed to suppress the warnings on a per-project basis by setting the Catch2 library as a system library in CMake: (include inside the above-mentioned CMake code after the line `include(Catch)`):
[...]

PR #2945, opened 2025-01-12
Silence -Weffc++ compiler warnings from gcc 11.4

grhoten:
## Description
These changes either address or silence the warnings that come from using -Weffc++ with gcc. This code has to be compiled in other projects, and it's easier to use this option if Catch isn't filling the compilation logs with warnings in its code, especially the headers. This compiler option has helped me in the past to find a subtle bug, and I'd like to continue doing that. Using this compiler option can also help highlight what does need to be tested.

The warnings about the operators were silenced with these changes because they seem like basic functionality. The rest were addressed as best that I can address them.

## GitHub Issues
This addresses the `-Weffc++` part of #2927.


grhoten:
This isn't ready yet.

PR #2946, opened 2025-01-12
Add test case stats functionality and test case name priority output

ligazetom:
## Description
Add 2 new flags `--output-test-case-name-first` and `--output-test-case-stats`. Former will print name of the test case before it is run (as was requested [here](https://github.com/catchorg/Catch2/issues/1579#issuecomment-983815709)) and the latter will print regular `TestCaseName: **stats**`. I've seen https://github.com/catchorg/Catch2/pull/2343 so this is a bit more lightweight and hopefully not too much against the grain with the argument namespace polution. idklmk

As is mentioned in refed issue, I was missing this feature as well and as much as I want to understand why it's not yet implemented I don't, but I understand why it's not the default.

This is more of a proposal as I'm not really confident with the naming and even worse, I'm using catch for like 3 days now, so I know very little about its internals and overall design.

With that said, I'd be glad to get some feedback. If this feature will be considered, I'll try to polish this PR more with tests and everything, but for that I'll need to look a bit deeper into docs.

I would split these "features" into two, but because of non-buffered output, there are some interdependencies because of newlines already (as with `-s` and `-d yes` would be and idk how many more flags that configure output) between these two, so I wanted to keep them together.

Also atm. I have no idea what table printer does so all I've really done is just ran it, saw it working and here we are.

Also, since everything was so easily accessible, this was pretty much low hanging fruit, I'm even a bit concerned if there isn't such functionality already, but I haven't found it. Would be pretty bad if I've implemented something that's already there. 🤞 

## GitHub Issues
https://github.com/catchorg/Catch2/issues/1579

### DIFF
--- a/src/catch2/reporters/catch_reporter_console.cpp
+++ b/src/catch2/reporters/catch_reporter_console.cpp
@@ -512,11 +512,31 @@ void ConsoleReporter::benchmarkFailed( StringRef error ) {
         << ColumnBreak() << RowBreak();
 }
 
+void ConsoleReporter::testCaseStarting( TestCaseInfo const& testInfo ) {
+    StreamingReporterBase::testCaseStarting( testInfo );
+
+    if ( m_config->outputTestCaseNameFirst() ) {
+        m_stream << testInfo.name << ":\n";
+    }
+}
+
 void ConsoleReporter::testCaseEnded(TestCaseStats const& _testCaseStats) {
     m_tablePrinter->close();
     StreamingReporterBase::testCaseEnded(_testCaseStats);
     m_headerPrinted = false;
+
+    if ( !_testCaseStats.aborting && m_config->outputTestCaseStats() ) {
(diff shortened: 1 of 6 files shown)


src/catch2/reporters/catch_reporter_console.cpp
```diff
@@ -512,11 +512,31 @@ void ConsoleReporter::benchmarkFailed( StringRef error ) {
         << ColumnBreak() << RowBreak();
 }
 
+void ConsoleReporter::testCaseStarting( TestCaseInfo const& testInfo ) {
+    StreamingReporterBase::testCaseStarting( testInfo );
+
+    if ( m_config->outputTestCaseNameFirst() ) {
+        m_stream << testInfo.name << ":\n";
```
ligazetom, line 519: Since the point of this is to print the test case name before we segfault I guess `flush` would be nice.

Issue #2641, opened 2023-02-08
ASan detecting double-free when using JUnit & console reporters simultaneously


jorgenpt:
**Describe the bug**
We recently upgraded to Catch3 to utilize the support for multiple reporters. We run all our tests with ASan on Linux, and this change has caused ASan to detect a double-free inside of MessageInfo::~MessageInfo. Here's the report:

```
=================================================================
==2304==ERROR: AddressSanitizer: attempting double-free on 0x60d0000a07c0 in thread T0:
    #0 0x520acd in operator delete(void*) (/home/ubuntu/jenkins/workspace/project/Binaries/Linux/TestsDevelopment+0x520acd)
    #1 0x556643 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string() /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/basic_string.h:672:9
    #2 0x556643 in Catch::MessageInfo::~MessageInfo() /home/ubuntu/jenkins/workspace/Server/Test/catch_amalgamated.hpp:1289:12
    #3 0x556643 in Catch::ScopedMessage::~ScopedMessage() /home/ubuntu/jenkins/workspace/Server/Test/catch_amalgamated.cpp:776:5
    #4 0x667352 in void std::_Destroy<Catch::ScopedMessage>(Catch::ScopedMessage*) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_construct.h:151:19
    #5 0x667352 in void std::_Destroy_aux<false>::__destroy<Catch::ScopedMessage*>(Catch::ScopedMessage*, Catch::ScopedMessage*) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_construct.h:163:6
    #6 0x667352 in void std::_Destroy<Catch::ScopedMessage*>(Catch::ScopedMessage*, Catch::ScopedMessage*) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_construct.h:195:7
[...]

Issue #1904, opened 2020-04-06
SEGV in the test framework


whatsthecraic:
Hi there,
I'm hitting a spurious SEGV in a multithreaded test, but, at this point, I'm not sure whether this depends by my program or by the test framework: 
```
Thread 32 "test_parallel" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff4cab700 (LWP 9574)]
std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string (this=0x58, __in_chrg=<optimized out>) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h:658
658	/build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h: No such file or directory.
(gdb) bt
#0  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string (this=0x58, __in_chrg=<optimized out>) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h:658
#1  0x00005555555a4178 in Catch::AssertionResultData::~AssertionResultData (this=0x38, __in_chrg=<optimized out>) at ../tests/catch.hpp:5322
#2  0x00005555555aa7b6 in Catch::AssertionResult::~AssertionResult (this=0x0, __in_chrg=<optimized out>) at ../tests/catch.hpp:5336
#3  0x00005555555c459d in Catch::Option<Catch::AssertionResult>::reset (this=0x7fffffffde88) at ../tests/catch.hpp:4385
#4  0x00005555555b49bd in Catch::Option<Catch::AssertionResult>::operator= (this=0x7fffffffde88, _value=...) at ../tests/catch.hpp:4378
#5  0x000055555557a578 in Catch::RunContext::assertionEnded (this=0x7fffffffde40, result=...) at ../tests/catch.hpp:12597
#6  0x000055555557c333 in Catch::RunContext::handleNonExpr (this=0x7fffffffde40, info=..., resultType=Catch::ResultWas::Ok, reaction=...) at ../tests/catch.hpp:12924
#7  0x000055555556bb8c in Catch::AssertionHandler::handleExceptionThrownAsExpected (this=0x7ffff4caad60) at ../tests/catch.hpp:8201
[...]

Issue #1714, opened 2019-08-02
Multi-threaded test running with address sanitizer detects SEGV


cmgreivel:
**Describe the bug**
Multi-threaded test with address sanitizer detects SEGV

**Expected behavior**
No SEGV in multi-threaded test.

**Reproduction steps**

```
#include <catch2/catch.hpp>
#include <deque>
#include <thread>

TEST_CASE("Threaded") {
  std::deque<std::thread> workers;
  for (int i = 0; i < 2; i++) {
    workers.emplace_back([]() {
      std::unique_ptr<std::string> tester(new std::string);
      for (int j = 0; j < 5000000; j++) {
        REQUIRE(bool(tester));
      }
    });
  }
  for (auto& t : workers) {
    t.join();
  }
}
```

Used compiler flags `-std=c++17 -g -O0 -fsanitize=address  -fuse-ld=gold`

**Platform information:**

 - OS: **Ubuntu 16.04.6 LTS**
 - Compiler+version: **c++ (Ubuntu 7.4.0-1ubuntu1~16.04~ppa1) 7.4.0**
 - Catch version: **v2.9.1**

**Additional context**
[asan_catch_output.txt](https://github.com/catchorg/Catch2/files/3463038/asan_catch_output.txt)

cmgreivel:
D'oh! This is due to the check being in the threads (https://github.com/catchorg/Catch2/blob/master/docs/limitations.md#thread-safe-assertions). Sorry for the churn.

Issue #1302, opened 2018-05-31
Catch2 documentation says "This limitation is highly unlikely to be lifted before Catch 2 is released"


RedDwarf69:
I'm just reading about Catch for first time. Finding "This limitation is highly unlikely to be lifted before Catch 2 is released" in https://github.com/catchorg/Catch2/blob/master/docs/limitations.md is confusing.

onqtam:
The idea is that using Catch asserts in threads before they are joined is not supported in Catch 1, and will be supported at some point in Catch 2 but that isn't the case yet. You still cannot use asserts from multiple threads.

horenmar:
Pretty much the above: There was a slight possibility that someone would be crazy enough to implement it in C++98, or add extra compile time configuration branches for it. At some point it should become part of Catch2, as C++11 makes it much easier, but it is not in yet.

How did you read it? Would it help if it said something along the lines of "not supported in Catch v1.x, not going to be supported in Catch v1.x, but there are plans for support in Catch v2.x"?

RedDwarf69:
Yes, that would be more clear.

"This limitation is highly unlikely to be lifted before Catch 2 is released"... Well it's not "highly unlikely", it's literally impossible. The probability is exactly zero. Catch 2 has already been released. So I reached the conclusion that the text was out of date, the information was not reliable and so it was equally possible that Catch 2 supported or not supported thread safe assertions.

PR #2948, opened 2025-01-18
Make assertions thread-safe

jeremy-rifkin:
## Description
This PR makes Catch2 assertion and logging macros thread-safe, allowing threads spawned in test cases to perform checks. It does this by introducing a global lock and locking at entry points called from these macros. This was easier and safer than trying to track down every use of static storage in the library and locking only around those, however, at some point that might be desirable.

A `std::recursive_mutex` is used for the global lock. There is some overhead associated with taking out the lock, however, it's extremely minimal in the uncontested (single-thread) case. In a benchmark locally I found that for a debug build on linux the overhead of the lock is around 300ns on my machine, meaning that a user would have to perform a million assertions in order to add a second of run-time overhead to their program.
```
benchmark name                       samples       iterations    est run time
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
no lock                                        100          1604     5.9348 ms 
                                        37.1058 ns     36.915 ns    37.4472 ns 
                                        1.26649 ns    0.80055 ns    1.98121 ns 
                                                                               
with std::recursive_mutex                      100           203     6.0088 ms 
                                        303.559 ns    299.777 ns    313.273 ns 
                                        28.6615 ns    11.6045 ns    56.3121 ns 
```

On release the overhead is far smaller:
```
benchmark name                       samples       iterations    est run time
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
no lock                                        100         20405     2.0405 ms 
                                        1.29613 ns    1.28875 ns    1.31015 ns 
                                      0.0501252 ns  0.0300847 ns  0.0787845 ns 
                                                                               
with std::recursive_mutex                      100           226      2.712 ms 
                                        137.018 ns    128.907 ns    150.877 ns 
                                        53.1962 ns    37.1166 ns    85.6079 ns 
```

MSVC debug:
```
benchmark name                       samples       iterations    est run time
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
no lock                                        100          6510      1.302 ms 
                                        2.07343 ns     2.0573 ns    2.10123 ns 
                                       0.105615 ns  0.0698496 ns    0.14992 ns

with std::recursive_mutex                      100           567     1.5309 ms 
                                        25.3474 ns    24.8342 ns    27.2275 ns 
                                        4.45937 ns    1.13563 ns    10.2695 ns
```

If this impact is deemed too high I have ideas for reducing the overhead in debug mode and optimizing for the uncontested case.

## GitHub Issues
#99
#246
#875
#1043
#1169
#1252
#1302
#1714
#1904
#2641
And probably others


src/catch2/internal/catch_assertion_handler.cpp
```diff
@@ -55,21 +68,26 @@ namespace Catch {
...
         m_resultCapture.handleNonExpr(m_assertionInfo, ResultWas::Ok, m_reaction);
     }
 
     void AssertionHandler::handleUnexpectedExceptionNotThrown() {
+        auto lock = take_global_lock();
         m_resultCapture.handleUnexpectedExceptionNotThrown( m_assertionInfo, m_reaction );
     }
 
     void AssertionHandler::handleThrowingCallSkipped() {
+        auto lock = take_global_lock();
```
shahsb, line 90: Codecov is complaining for this line, please check:

```
Check warning on line 90 in src/catch2/internal/catch_assertion_handler.cpp


Codecov
/ codecov/patch
src/catch2/internal/catch_assertion_handler.cpp#L90

Added line #L90 was not covered by tests
```

shahsb:
The sentence “at some point that might be desirable” is a bit vague. It would be more helpful to either clarify what you mean or remove it. If you're suggesting the possibility of refining the approach in the future (e.g., focusing on static storage), consider expanding on that idea.

shahsb:
Clarify the environment under which these benchmarks were taken (e.g., specific hardware, OS, and configuration), as this can significantly affect the results. Also, mention if these benchmarks are reflective of real-world usage or more of a worst-case scenario.

shahsb:
You mention that if the overhead is deemed too high, you have ideas for optimization. It would be great if you could briefly outline what those optimizations might look like. Even a simple mention like “Reducing the number of locks taken in the debug mode” could provide a bit more insight into your thought process.

benchmarks/catch_benchmarks.cpp
```diff
@@ -0,0 +1,25 @@
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/benchmark/catch_benchmark.hpp>
+
+#include <mutex>
+
+std::recursive_mutex global_lock;
```
shahsb, line 6: Explain the rationale behind choosing std::recursive_mutex over other types of mutexes.

shahsb:
Clarify the impact of this change on larger test suites, as that could be relevant for performance-sensitive environments.

shahsb (changes requested):

jeremy-rifkin:
Hi @shahsb, I am awaiting further feedback from the maintainer who I have been in contact with regarding this PR over discord. I'm a bit confused by your reviews, are you involved with the Catch2 project?

src/catch2/internal/catch_assertion_handler.cpp
```diff
@@ -22,13 +25,23 @@ namespace Catch {
     :   m_assertionInfo{ macroName, lineInfo, capturedExpression, resultDisposition },
         m_resultCapture( getResultCapture() )
     {
+        auto lock = take_global_lock();
         m_resultCapture.notifyAssertionStarted( m_assertionInfo );
     }
 
+    AssertionHandler::~AssertionHandler() {
+        auto lock = take_global_lock();
```
RedBeard0531, line 33: Can this be moved inside the `if` or is `m_completed` written to from other threads?

RedBeard0531 (commented):
This cannot possibly be sufficient. As just one missing case: https://github.com/catchorg/Catch2/blob/f51dc98dfcaecf402e2d3aa3e0282b27beb4e237/src/catch2/internal/catch_test_macro_impl.hpp#L58-L60

It may be better to have `lastAssertionPassed` (or possibly the whole `ResultCapture`) be thread-local.

jeremy-rifkin:
Thanks @RedBeard0531, good point. Catch2 has a ton of internal state, this is hard to track down :)

It does look like it might be sufficient to make `m_lastAssertionPassed` thread-local. I'm also noticing there are some counters like `m_totals.assertions.passed` which should be made atomic.

src/catch2/reporters/catch_reporter_console.cpp
```diff
@@ -480,6 +484,7 @@ void ConsoleReporter::benchmarkPreparing( StringRef name ) {
 }
 
 void ConsoleReporter::benchmarkStarting(BenchmarkInfo const& info) {
+    auto lock = take_global_lock();
```
horenmar: Locking in ConsoleReporter is suspect, because nothing enforces that the users use it. It should be done further up the callstack, e.g. in run_context.

src/catch2/internal/catch_assertion_handler.cpp
horenmar: The handler is on stack, so multiple threads shouldn't be writing to it.

src/catch2/internal/catch_assertion_handler.cpp
```diff
@@ -22,13 +25,23 @@ namespace Catch {
...
+        if ( !m_completed ) {
+            m_resultCapture.handleIncomplete( m_assertionInfo );
+        }
+    }
+
     void AssertionHandler::handleExpr( ITransientExpression const& expr ) {
+        auto lock = take_global_lock();
```
horenmar, line 40: All these should live in the capture, because it already handles things like assertion fast-path if no reporters have to be notified about a passing assertion.

Speaking of, the current fast path for assertions is a simple counter increment without invoking the reporter. We should be able to have similar fast path (with an atomic counter) after this.

horenmar:
Update on this: I started making some changes, realized that bigger refactoring is needed and started doing that in the main repo instead.

The good part is that the refactoring improved runtime of single-threaded assertions to the point where current `devel` is almost 2x faster than the last release at handling passing assertions that do not need to be reported and thus can go into the fast path. (The slow path has seen ~5-10% improvement)

The bad part is that I also implemented thread safe assertions (messages not supported for simplicity) in branch `devel-thread-experiments`. The design should be correct and so far the testing bears this out, **but** it introduces over 40% slowdown compared to current `devel`.

The final result will probably be that the thread safe assertions will live behind a compile-time configuration option, because the perf cost is too big.

PR #2949, opened 2025-02-03
arm builds

johnnynunez:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

Digits, jetson thor, cuda arm laptops are coming

windows arm q2 2025: https://github.com/github/roadmap/issues/1098


ChrisThrasher:
In my experience, unless you're doing exotic things like using compiler intrinsics and inline assembly, CPU architecture is a non-factor in how I write and test my C++ code. What types of errors do you think these new jobs will catch that the existing CI jobs will not catch?

johnnynunez:
> In my experience, unless you're doing exotic things like using compiler intrinsics and inline assembly, CPU architecture is a non-factor in how I write and test my C++ code. What types of errors do you think these new jobs will catch that the existing CI jobs will not catch?

Because in H2 a lot of arm devices are coming... (cuda laptops, digits etc)

ChrisThrasher:
I'm afraid that doesn't really answer my question. The computer hardware market is already full of ARM devices thanks to Apple. I've been using Catch2 on an ARM device for many years. What kinds of errors would these ARM jobs catch that the existing x86 jobs would not catch? My point is that the existing CI pipeline may already be sufficient but I might be missing some of the value of the jobs you added.

ChrisThrasher:
I'm following up on my previous comment. What kinds of errors would these new jobs catch that the existing CI jobs would not catch? Catch2 does not to my knowledge employ any x86 intrinsics so I'm not sure why it would ever fail to compile on an Arm Windows machine. The nature of ISO C++ code is that it automatically works on all CPU architectures assuming it's free of UB.

PR #2965, opened 2025-03-22
clang: do not issue bogus warnings about integer manipulation in hash functions with fsanitize=undefined/integer

jcelerier:
With -fsanitize=integer every over/under-flowing integer manipulation triggers a warning.
This is extremely useful as it allows to find some non-obvious bugs such as

    for(size_t i = 0; i < N - 1; i++) { ... }

But it comes with a lot of false positives, for instance with every hash function
doing shifting on unsigned integer. Random number generators are also often detected
with this sanitizer.

This marks a few of these functions as safe in this case.

### DIFF
--- a/src/catch2/internal/catch_random_number_generator.cpp
+++ b/src/catch2/internal/catch_random_number_generator.cpp
@@ -9,7 +9,7 @@
 
 #if defined( __clang__ )
 #    define CATCH2_CLANG_NO_SANITIZE_INTEGER \
-        __attribute__( ( no_sanitize( "undefined", "integer" ) ) )
+        __attribute__( ( no_sanitize( "unsigned-integer-overflow" ) ) )
 #else
 #    define CATCH2_CLANG_NO_SANITIZE_INTEGER
 #endif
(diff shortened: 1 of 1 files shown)


shahsb (approved):

src/catch2/internal/catch_random_number_generator.cpp
```diff
@@ -7,6 +7,12 @@
 // SPDX-License-Identifier: BSL-1.0
 #include <catch2/internal/catch_random_number_generator.hpp>
 
+#if defined( __clang__ )
+#    define CATCH2_CLANG_NO_SANITIZE_INTEGER \
+        __attribute__( ( no_sanitize( "undefined", "integer" ) ) )
```
horenmar: Thanks, but reading the [documentation here](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-undefined), I think this is supposed to be 

```suggestion
        __attribute__( ( no_sanitize( "unsigned-integer-overflow" ) ) )
```

instead, to only target unsigned integer overflows.

horenmar:
Thanks

PR #2972, opened 2025-04-13
Move header from internal

oboukli:
Move `catch_test_run_info.hpp` from directory
`src/catch2/internal/` to be directly under `src/catch2/`.

Header `catch_test_run_info.hpp` defines `Catch::TestRunInfo` which is part of the public API, since it is required when overriding `Catch::EventListenerBase`.

Fixes issue caught by Clang Tidy misc-include-cleaner check.


neheb (approved):

shahsb (approved):

horenmar:
I spent some time thinking about this.

The change is of course technically correct. However, the change also has a good chance of breaking someone's build, while not having a large payoff. I've ended up deciding to merge the PR, because it won't break most users and the fix is easy.

PR #2974, opened 2025-04-16
Remove unnecessary `friend` keywords

ZXShady:
Note that I didn't remove it for `internal\catch_optional.hpp` for ` friend bool operator==(Optional<T>,Optional<T>)` and `operator!=` since the friend function will allow for implicit conversions which allows comparing against a `T` like `Catch::Optional<int>() == int(1)` nor did I make inline functions not inline like `operator<<` for `ITransientExpression`.

This change removes several friend declarations that are no longer necessary. Overuse of friend can lead to tighter coupling between classes, reduced encapsulation and thinking it relies on implementation details.
in general avoid granting excessive access unless explicitly required.

### DIFF
--- a/src/catch2/catch_approx.hpp
+++ b/src/catch2/catch_approx.hpp
@@ -50,21 +50,6 @@ namespace Catch {
         }
 
         template <typename T, typename = std::enable_if_t<std::is_constructible<double, T>::value>>
-        friend bool operator == ( Approx const& lhs, const T& rhs ) {
-            return operator==( rhs, lhs );
-        }
-
-        template <typename T, typename = std::enable_if_t<std::is_constructible<double, T>::value>>
-        friend bool operator != ( T const& lhs, Approx const& rhs ) {
-            return !operator==( lhs, rhs );
-        }
-
-        template <typename T, typename = std::enable_if_t<std::is_constructible<double, T>::value>>
-        friend bool operator != ( Approx const& lhs, T const& rhs ) {
-            return !operator==( rhs, lhs );
-        }
-
-        template <typename T, typename = std::enable_if_t<std::is_constructible<double, T>::value>>
         friend bool operator <= ( T const& lhs, Approx const& rhs ) {
             return static_cast<double>(lhs) < rhs.m_value || lhs == rhs;
         }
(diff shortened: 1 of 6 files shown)


neheb:
```
../src/catch2/internal/catch_reusable_string_stream.hpp:46:20: error: call to function 'operator<<' that is neither visible in the template definition nor found by argument-dependent lookup
   46 |             *m_oss << value;
      |                    ^
../src/catch2/catch_tostring.hpp:124:21: note: in instantiation of function template specialization 'Catch::ReusableStringStream::operator<<<foo::helper_1403>' requested here
  124 |                 rss.operator<<(value);
      |                     ^
../src/catch2/catch_tostring.hpp:146:88: note: in instantiation of function template specialization 'Catch::StringMaker<foo::helper_1403>::convert<foo::helper_1403>' requested here
  146 |             return ::Catch::StringMaker<std::remove_cv_t<std::remove_reference_t<T>>>::convert(e);
      |                                                                                        ^
../src/catch2/internal/catch_decomposer.hpp:195:42: note: in instantiation of function template specialization 'Catch::Detail::stringify<foo::helper_1403>' requested here
  195 |                     ( os, Catch::Detail::stringify( m_lhs ), m_op, Catch::Detail::stringify( m_rhs ) );
      |                                          ^
../src/catch2/internal/catch_decomposer.hpp:199:19: note: in instantiation of member function 'Catch::BinaryExpr<foo::helper_1403 &, foo::helper_1403 &>::streamReconstructedExpression' requested here
  199 |         constexpr BinaryExpr( bool comparisonResult, LhsT lhs, StringRef op, RhsT rhs )
      |                   ^
../tests/SelfTest/UsageTests/Compilation.tests.cpp:32:22: note: 'operator<<' should be declared prior to the call site or in namespace 'foo'
   32 | static std::ostream& operator<<(std::ostream& out, foo::helper_1403 const&) {
      |                      ^
1 error generated.
```

horenmar (commented):
`friend` declaration in C++ has two reasons

1) To access the internals of a class
2) To declare a [hidden friend](https://codingnest.com/the-little-things-speeding-up-c-compilation/#hiddenfriends)

While it is annoying that to get 2) you also have to have 1), 2) is desirable enough to keep the functions as hidden friends even if they do not need access to the class internals.

ZXShady:
> `friend` declaration in C++ has two reasons
> 
> 1) To access the internals of a class
> 2) To declare a [hidden friend](https://codingnest.com/the-little-things-speeding-up-c-compilation/#hiddenfriends)
> 
> While it is annoying that to get 2) you also have to have 1), 2) is desirable enough to keep the functions as hidden friends even if they do not need access to the class internals.

Yea, 

I guess it can be removed for friended functions declared out of line since hidden friends only work if it is defined inline.

I will update my pr.

horenmar:
[Hidden friends can be defined out of line as well](https://stackoverflow.com/a/62345568/3276485)

ZXShady:
> [Hidden friends can be defined out of line as well](https://stackoverflow.com/a/62345568/3276485)

Oh! I didn't know that, you can close my pr.

PR #2975, opened 2025-04-17
Use variable templates instead of `struct`s for traits

ZXShady:
Variable templates are simpler for the compiler and just better to read and write.

Note i did not change `is_range` to variable template since it is part of public API therefore this is an API break.


horenmar (commented):
The changes look fine, but the new variables should follow the stdlib convention of using `_v` suffix for constexpr bool variables.

ChrisThrasher (approved):

PR #2976, opened 2025-04-27
Add return code constants to API

### LINKED ISSUE #2955: Make exit codes available to users
**Description**
Currently (catch2 v3.8.0) the exit codes are defined in an anonymous namespace inside `catch_session.cpp`. [permalink to github code](https://github.com/catchorg/Catch2/blob/914aeecfe23b1e16af6ea675a4fb5dbd5a5b8d0a/src/catch2/catch_session.cpp#L37)

```cpp
namespace Catch {

    namespace {
        static constexpr int TestFailureExitCode = 42;
        static constexpr int UnspecifiedErrorExitCode = 1;
        static constexpr int AllTestsSkippedExitCode = 4;
        static constexpr int NoTestsRunExitCode = 2;
        static constexpr int UnmatchedTestSpecExitCode = 3;
        static constexpr int InvalidTestSpecExitCode = 5;
// more code
    }
}
```

When writing a user-defined main function that calls `auto const result = Catch::Session().run(argc, argv)`, it would be helpful to be able compare the returned `result` to the defined values.
Currently this is not possible as they are only inside the anonymous namespace, leading to internal linkage. Redefining the exit values in user code is not optimal.

My suggestion would be to define the exit codes in the `catch_session.hpp` header, so they are available to end-users.

ChrisThrasher:
Closes #2955

### DIFF
--- a/src/catch2/catch_session.hpp
+++ b/src/catch2/catch_session.hpp
@@ -16,6 +16,14 @@
 
 namespace Catch {
 
+    // TODO: Use C++17 `inline` variables
+    constexpr int UnspecifiedErrorExitCode = 1;
+    constexpr int NoTestsRunExitCode = 2;
+    constexpr int UnmatchedTestSpecExitCode = 3;
+    constexpr int AllTestsSkippedExitCode = 4;
+    constexpr int InvalidTestSpecExitCode = 5;
+    constexpr int TestFailureExitCode = 42;
+
     class Session : Detail::NonCopyable {
     public:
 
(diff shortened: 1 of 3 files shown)


src/catch2/catch_session.hpp
```diff
@@ -16,6 +16,13 @@
 
 namespace Catch {
 
+    constexpr int TestFailureExitCode = 42;
```
ChrisThrasher: I believe these constants will be defined separately in each translation unit leading to marginally higher memory usage but I don't believe it will cause any linker errors so it should function exactly as expected. I'd prefer to use C++17's `inline` here but that's out of reach so long as our minimum C++ standard is C++14.

horenmar: This is _fiiiiiine_.

Worth adding a note that these can be upgraded to `inline` constexpr vars with C++17.

PR #2987, opened 2025-06-09
fix(python): fix invalid escape sequences

e-kwsm:
## Description

Regex such as `'\d'` is invalid, which must be `r'\d'` or `'\\d'`.

### DIFF
--- a/tests/TestScripts/DiscoverTests/VerifyRegistration.py
+++ b/tests/TestScripts/DiscoverTests/VerifyRegistration.py
@@ -17,7 +17,7 @@ from typing import List
 
 TestInfo = namedtuple('TestInfo', ['name', 'tags'])
 
-cmake_version_regex = re.compile('cmake version (\d+)\.(\d+)\.(\d+)')
+cmake_version_regex = re.compile(r'cmake version (\d+)\.(\d+)\.(\d+)')
 
 def get_cmake_version():
     result = subprocess.run(['cmake', '--version'],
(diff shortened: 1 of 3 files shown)


ChrisThrasher:
Could you please explain what makes these sequences invalid?

e-kwsm:
See <https://docs.python.org/3.13/library/re.html>.

> The solution is to use Python’s raw string notation for regular expression
patterns; backslashes are not handled in any special way in a string literal
prefixed with <code class="docutils literal notranslate"><span class="pre">'r'</span></code>.  So <code class="docutils literal notranslate"><span class="pre">r&quot;\n&quot;</span></code> is a two-character string containing
<code class="docutils literal notranslate"><span class="pre">'\'</span></code> and <code class="docutils literal notranslate"><span class="pre">'n'</span></code>, while <code class="docutils literal notranslate"><span class="pre">&quot;\n&quot;</span></code> is a one-character string containing a
newline.  Usually patterns will be expressed in Python code using this raw
string notation.

```python
Python 3.13.3 (main, Apr  9 2025, 07:44:25) [GCC 14.2.1 20250207] on linux
>>> s = "\d"   # wrong
<python-input-0>:1: SyntaxWarning: invalid escape sequence '\d'
>>> s = r"\d"  # correct
```

horenmar (approved):
@ChrisThrasher The changes are correct.

Honestly it is kinda fascinating that the scripts work, I am surprised that after Python sees the escape is not valid, it just leaves the literal backslash and letter in there.

PR #2988, opened 2025-06-09
fix: make deleted member function public

e-kwsm:
## Description

https://github.com/catchorg/Catch2/blob/3013cb897b5706e8532507cb2b6ac33e1fc35d93/.clang-tidy#L21

modernize-use-equals-delete is enabled and the code complies with it, but clang-tidy issues warning: "deleted member function should be public."

### DIFF
--- a/src/catch2/internal/catch_noncopyable.hpp
+++ b/src/catch2/internal/catch_noncopyable.hpp
@@ -13,6 +13,7 @@ namespace Catch {
 
         //! Deriving classes become noncopyable and nonmovable
         class NonCopyable {
+        public:
             NonCopyable( NonCopyable const& ) = delete;
             NonCopyable( NonCopyable&& ) = delete;
             NonCopyable& operator=( NonCopyable const& ) = delete;


extras/catch_amalgamated.hpp
```diff
@@ -1129,6 +1129,7 @@ namespace Catch {
 
         //! Deriving classes become noncopyable and nonmovable
         class NonCopyable {
+        public:
```
ChrisThrasher: https://github.com/catchorg/Catch2/blob/devel/docs/contributing.md#using-github

> When submitting a pull request please do not include changes to the amalgamated distribution files. This means do not include them in your git commits!

The amalgamated header and source file get updated with each new release. They are not meant to be edited manually.

e-kwsm: Thank you for your review. I reverted the change in amalgam.

ChrisThrasher (approved):
Good catch. Thanks for all these clang-tidy cleanup.

PR #2995, opened 2025-06-30
Allow using only types in `TEMPLATE_TEST_CASE_SIG`.

### LINKED ISSUE #2680: `TEMPLATE_TEST_CASE_SIG` multiple `typename` parameters in a row
Why is it not possible to declare `TEMPLATE_TEST_CASE_SIG` with two `typename` parameters, as presented in the example below?
```cpp
TEMPLATE_TEST_CASE_SIG("TemplateTestSig: arrays can be created from NTTP arguments", "[vector][template][nttp]",
    ((typename T, typename V), T, V), (int, double))
{
}
```
The error follows:
```
error: call of overloaded 'get_wrapper<int, double>()' is ambiguous
159 | TEMPLATE_TEST_CASE_SIG("TemplateTestSig: arrays can be created from NTTP arguments", "[vector][template][nttp]",
       | ^~~~~~~~~~~~~~~~~~~~~~
```

legrosbuffle:
Right now `TEMPLATE_TEST_CASE_SIG` fails to compile when the signature contains only types:

```
TEMPLATE_TEST_CASE_SIG(
  "TemplateTestSig: compiles with two type parameters",
  "[template][onlytypes]",
  ((typename U, typename V), U, V), (int,int)) {}
```

The trick is to resolve the ambiguity between the two overloads of `get_wrapper` (`TypeList` and `Nttp`) by making one match more strongly. We also need to allow `reg_test` to register more than one type.

Add unit tests.

Fixes https://github.com/catchorg/Catch2/issues/2680


horenmar:
Thanks. I made small changes for the tests to pass and to avoid polluting the global namespace with the priority tag.

PR #3004, opened 2025-07-21
Optin to VT100 style terminal codes on Windows

### LINKED ISSUE #3003: Allow/Default use of ANSI color codes on Windows
**Description**
With a little snippet during initialization, Windows 10 and newer support ANSI color codes out of the box:

```c++
#ifdef _WIN32
    HANDLE outputHandle = GetStdHandle(STD_OUTPUT_HANDLE);
    DWORD mode = 0;
    if (GetConsoleMode(outputHandle, &mode))
    {
        mode |= ENABLE_PROCESSED_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING;
        SetConsoleMode(outputHandle, mode);
    }
#endif // _WIN32
```

The `CATCH_INTERNAL_HAS_ISATTY`  is somewhat malformed - Windows has `isatty` too (in header `#include <io.h>` instead of `#include <unistd.h>`) - but what it doesn't have is `STDOUT_FILENO` so the portable solution for that is `isatty(fileno(stdout))`.

On Windows 10 or newer (or in case that covers ALL support Windows platforms: For every Windows-platform in general), the ANSI color code support should be preferred unconditionally over the legacy text attribute API.

**Additional context**
This avoids a lot of issues you'd otherwise have with the Windows terminal handling - among others that setting colors via `SetConsoleTextAttribute` on Windows breaks horribly when the application under test is enabling output buffering for `stdout` as the legacy API for color codes only every works as long as `stdout` is byte-buffered only.

Ext3h:
## Description

Enable VT100 terminal color codes on Windows too (if available) and use them by default.

## GitHub Issues

Closes https://github.com/catchorg/Catch2/issues/3003

### DIFF
--- a/src/catch2/internal/catch_console_colour.cpp
+++ b/src/catch2/internal/catch_console_colour.cpp
@@ -178,10 +178,7 @@ namespace {
         void use( Colour::Code _colourCode ) const override {
             auto setColour = [&out =
                                   m_stream->stream()]( char const* escapeCode ) {
-                // The escape sequence must be flushed to console, otherwise
-                // if stdin and stderr are intermixed, we'd get accidentally
-                // coloured output.
-                out << '\033' << escapeCode << std::flush;
+                out << '\033' << escapeCode;
             };
             switch( _colourCode ) {
                 case Colour::None:
--- a/src/catch2/internal/catch_console_colour.cpp
+++ b/src/catch2/internal/catch_console_colour.cpp
@@ -97,34 +97,6 @@ namespace Catch {
 
 } // namespace Catch
 
-#if defined ( CATCH_PLATFORM_WINDOWS )
-namespace Catch {
-namespace {
-
-    bool enableVirtualTerminalSupport(bool restore)
-    {
-        HANDLE outputHandle = GetStdHandle( STD_OUTPUT_HANDLE );
-        DWORD mode = 0;
-        if ( GetConsoleMode( outputHandle, &mode ) ) {
-            // VT100 style sequence processing has to be enabled by explicit
-            // opt-in.
-            DWORD newMode = mode | ENABLE_PROCESSED_OUTPUT |
-                            ENABLE_VIRTUAL_TERMINAL_PROCESSING;
-            bool virtualTerminalSupported = SetConsoleMode( outputHandle, newMode );
-            if(restore)
-            {
-                SetConsoleMode( outputHandle, mode );
-            }
-            return virtualTerminalSupported;
-        } else {
-            return false;
-        }
-    }
-
-}
-}
-#endif
-
 #if defined ( CATCH_CONFIG_COLOUR_WIN32 ) /////////////////////////////////////////
 
 namespace Catch {
@@ -142,13 +114,13 @@ namespace {
         }
 
         static bool useImplementationForStream(IStream const& stream) {
-            // Win32 text colour APIs can only be used on console streams
-            // We cannot check that the output hasn't been redirected,
-            // so we just check that the original stream is console stream.
-            bool useColour = stream.isConsole();
-            // If available, prefer VT100 style escape sequences.
-            useColour = useColour && !enableVirtualTerminalSupport( true );
-            return useColour;
+            OSVERSIONINFOA versionInfo;
+            // Use as fallback for Windows versions <10.0 only.
+            // Newer Windows versions have full support for ANSI color codes.
+            if ( GetVersionExA( &versionInfo ) && versionInfo.dwMajorVersion < 10 ) {
+                return stream.isConsole();
+            }
+            return false;
         }
 
     private:
@@ -196,17 +168,10 @@ namespace {
     class ANSIColourImpl final : public ColourImpl {
     public:
         ANSIColourImpl( IStream* stream ): ColourImpl( stream ) {
-#if defined( CATCH_PLATFORM_WINDOWS )
-            enableVirtualTerminalSupport(false);
-#endif
         }
 
         static bool useImplementationForStream(IStream const& stream) {
-            bool useColour = stream.isConsole();
-#    if defined( CATCH_PLATFORM_WINDOWS )
-            useColour = useColour && enableVirtualTerminalSupport(true);
-#    endif
-            return useColour;
+            return stream.isConsole();
         }
 
     private:
--- a/src/catch2/internal/catch_console_colour.cpp
+++ b/src/catch2/internal/catch_console_colour.cpp
@@ -97,6 +97,42 @@ namespace Catch {
 
 } // namespace Catch
 
+#if defined( CATCH_PLATFORM_LINUX ) || defined( CATCH_PLATFORM_MAC ) || \
+    defined( __GLIBC__ )
+#    define CATCH_INTERNAL_HAS_ISATTY
+#    include <unistd.h>
+#elif defined( CATCH_PLATFORM_WINDOWS )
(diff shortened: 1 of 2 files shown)


src/catch2/internal/catch_console_colour.cpp
```diff
@@ -185,7 +223,10 @@ namespace {
 #if defined( CATCH_INTERNAL_HAS_ISATTY ) && \
     !( defined( __DJGPP__ ) && defined( __STRICT_ANSI__ ) )
             ErrnoGuard _; // for isatty
-            useColour = useColour && isatty( STDOUT_FILENO );
+            useColour = useColour && _isatty( _fileno( stdout ) );
```
Ext3h, line 226: Hate that part... That check should be done in `CoutStream` / `CerrStream`. The whole `_isatty` stuff is duplicated, and it's not as if `isConsole()` ***needs*** to be lying.

Not to mention this is making a judgment on `stdout` even if `stderr` happens to be in use.

src/catch2/internal/catch_console_colour.cpp
```diff
@@ -97,6 +97,42 @@ namespace Catch {
...
+#    include <io.h>
+#endif
+
+#if defined ( CATCH_PLATFORM_WINDOWS )
+namespace Catch {
+namespace {
+
+    bool enableVirtualTerminalSupport(bool restore)
+    {
+        HANDLE outputHandle = GetStdHandle( STD_OUTPUT_HANDLE );
```
Ext3h, line 115: This is not hitting the correct stream in case `CerrStream` is in use.

src/catch2/internal/catch_console_colour.cpp
```diff
@@ -118,7 +154,11 @@ namespace {
             // Win32 text colour APIs can only be used on console streams
             // We cannot check that the output hasn't been redirected,
```
Ext3h, line 155: That comment is a lie. `_isatty` works just fine for that purpose. Except... `stdout` is not the right stream to probe!

src/catch2/internal/catch_istream.cpp
```diff
@@ -88,27 +97,52 @@ namespace Detail {
 
         class CoutStream final : public IStream {
             std::ostream m_os;
+            bool m_isatty;
         public:
             // Store the streambuf from cout up-front because
             // cout may get redirected when running tests
-            CoutStream() : m_os( Catch::cout().rdbuf() ) {}
+            CoutStream() : m_os( Catch::cout().rdbuf() ) {
+                m_isatty = true;
```
Ext3h, line 128: Too much Copy&Paste for my taste - and since `Catch::cout()` can be reimplemented, this is quite possibly probing the wrong file handles...

src/catch2/internal/catch_istream.cpp
```diff
@@ -86,29 +96,76 @@ namespace Detail {
...
+#endif
+#if defined( CATCH_PLATFORM_MAC ) || defined( CATCH_PLATFORM_IPHONE )
+                    m_isatty = m_isatty && !isDebuggerActive();
+#endif
+            }
 
         public: // IStream
             std::ostream& stream() override { return m_os; }
-            bool isConsole() const override { return true; }
+            bool isConsole() const override { return m_isatty; }
```
Ext3h, line 144: This does not pass the tests. Now `isConsole` reports `false` in all of the unit tests when run in the CI - but its doing so rightfully.

src/catch2/internal/catch_istream.cpp
```diff
@@ -86,29 +96,76 @@ namespace Detail {
...
+                m_isatty = true;
+#if defined( CATCH_INTERNAL_HAS_ISATTY ) && \
+    !( defined( __DJGPP__ ) && defined( __STRICT_ANSI__ ) )
+                ErrnoGuard _; // for isatty
+                m_isatty = m_isatty && isatty( STDERR_FILENO );
+#elif defined( CATCH_PLATFORM_WINDOWS )
+                m_isatty = m_isatty && enableVirtualTerminalSupport( STD_ERROR_HANDLE );
+#endif
+#if defined( CATCH_PLATFORM_MAC ) || defined( CATCH_PLATFORM_IPHONE )
+                m_isatty = m_isatty && !isDebuggerActive();
```
Ext3h, line 164: Copy&Paste from original `ANSIColourImpl` code  - unfortunately no comments, so what is this supposed to do? Is the debugger specifically not compatible with VT100 style sequences? Or does it not behave like a console in some other sense?

src/catch2/internal/catch_console_colour.cpp
```diff
@@ -115,10 +114,13 @@ namespace {
         }
 
         static bool useImplementationForStream(IStream const& stream) {
```
Ext3h, line 116: We might just kick this one out of the auto-selection. Even the version check here is wonky, because it's easily thrown off guard by manifests or compatibility mode.

Ext3h:
That `std::flush` in `ANSIColourImpl` is fishy - and the comment is a lie. There's an `std::flush` in about every place where it would be appropriate (i.e. after every complete output), but flushing after each escape sequence is pointless.

src/catch2/internal/catch_console_colour.cpp
```diff
@@ -160,48 +162,23 @@ namespace {
...
-
-            return useColour;
+            return stream.isConsole();
         }
 
     private:
         void use( Colour::Code _colourCode ) const override {
             auto setColour = [&out =
                                   m_stream->stream()]( char const* escapeCode ) {
-                // The escape sequence must be flushed to console, otherwise
```
Ext3h: That comment is partially a lie. `std::cout`, `std::cerr` and `std::cin` are all [tied together](https://en.cppreference.com/w/cpp/io/basic_ios/tie.html). But they are (by default) not chained up correctly!

Problem is that while `std::cerr` and `std::cin` can cause an (implicit) flush of `std::cout`, `std::cin` doesn't do so on `std::cerr`, and neither does `std::cout` imply of flush of `std::cerr`.


Then again - it doesn't matter, because the terminal shouldn't apply escape sequences received on `stderr` to `stdin`  in the first place, so this appears to be entirely incorrect after all?

And there's also `std::flush` explicitly throughout the entire codebase wherever necessary.

Ext3h: Meanwhile, there are no tests actually setting up a stream-buffering that would have verified the necessity / correctness of the flushing in the first place.

I.e.:
```c++
setvbuf(stdout, NULL, _IOFBF, 3072);
setvbuf(stderr, NULL, _IOFBF, 3072);
```

That would quickly show where buffering is still broken.

The case where it does still break, is when configuring Catch2 to use `std::cout` as the output channel, while the code under test is utilizing `std::cerr` / `stderr` (but without explicit flushing and with output buffering explicitly enabled - which is NOT the default for `stderr` even when redirected!). But in that edge case, flushing `std::cout` still doesn't do anything about the un-flushed `stderr`.

src/catch2/internal/catch_istream.cpp
Ext3h: Does it even apply to `stderr` or is `stderr` not usable with debuggers on Mac/Iphone?

src/catch2/internal/catch_istream.cpp
Ext3h: Two options - injecting a proper terminal emulator for testing or changing the expected results?

PR #3005, opened 2025-07-24
Forbid deducing reference types for m_predicate in FilterGenerator

### LINKED ISSUE #3002: Possible undefined behavior with filter generator
**Describe the bug**
The following arrangement of generators behaves unexpectedly on different platforms with different compilations.

```
using secs = std::chrono::seconds;
using mins = std::chrono::minutes;

SCENARIO("Repro")
{
    using namespace Catch::Generators;
    const auto a = GENERATE(secs{ 0 }, secs{ 30 });

    const auto is_valid_b = [=] (const auto b) -> bool
    {
        return (a == secs{ 0 }) || (b <= a);
    };
    const auto b = GENERATE_COPY(
        filter(is_valid_b, values<secs>({ secs{ 10 }, secs{ 20 }, mins{ 1 }, mins{ 2 }, mins{ 5 }, mins{ 10 }, mins{ 30 } })));

    CHECK((a == secs{ 0 }) || (b <= a));
}
```
Amalgamated v3.7.1 is used everywhere.
With a local Windows 11 build, the generator seems to be working as expected, only 10s and 20s for 'b' is being generated when 'a' is not 0.
However, our Azure builds behave as if the filter was not there at all, both Win and Linux.
I did some debugging, but it is difficult to see what is going on with a release build.
The binaries from the server build do have the filter compiled, but when the filter's next() is invoked, it looks like as if the predicate lambda always captures 0 for the value of 'a', even if it is actually 30s.
Could not put a breakpoint inside the filter generators constructor, but manually stepping in Visual Studio, I could go into the constructor, and the local predicate seems to have the correct 30s value captured. When the filter's get() is called, it also shows the captured 30 in this->m_predicate.

I guess the big question is, **is this even supposed to work**?

**Expected behavior**
Filter should work regardless of compilation environment.

**Reproduction steps**
See the code above.

**Platform information:**
 - OS: **Win 11**
 - Compiler+version: **MSVC (Visual Studio 17.14.5)**
 - Catch version: **v3.7.1**

Szapi:
## Description
Forbid deducing reference types for m_predicate in FilterGenerator to prevent dangling references.
This is needed for out-of-line predicates to work correctly instead of undefined behavior or crashes.

## GitHub Issues
Closes #3002

### DIFF
--- a/src/catch2/generators/catch_generators_adapters.hpp
+++ b/src/catch2/generators/catch_generators_adapters.hpp
@@ -58,8 +58,9 @@ namespace Generators {
     class FilterGenerator final : public IGenerator<T> {
         GeneratorWrapper<T> m_generator;
         Predicate m_predicate;
+        static_assert(!std::is_reference<Predicate>::value, "This would most likely result in a dangling reference");
     public:
-        template <typename P = Predicate>
+        template <typename P>
         FilterGenerator(P&& pred, GeneratorWrapper<T>&& generator):
             m_generator(CATCH_MOVE(generator)),
             m_predicate(CATCH_FORWARD(pred))
@@ -91,7 +92,7 @@ namespace Generators {
 
     template <typename T, typename Predicate>
     GeneratorWrapper<T> filter(Predicate&& pred, GeneratorWrapper<T>&& generator) {
-        return GeneratorWrapper<T>(Catch::Detail::make_unique<FilterGenerator<T, Predicate>>(CATCH_FORWARD(pred), CATCH_MOVE(generator)));
+        return GeneratorWrapper<T>(Catch::Detail::make_unique<FilterGenerator<T, typename std::remove_reference<Predicate>::type>>(CATCH_FORWARD(pred), CATCH_MOVE(generator)));
     }
 
     template <typename T>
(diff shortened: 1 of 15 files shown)


horenmar:
Thanks, this is useful change.

To avoid splitting the discussion into two places, I am gonna comment on the issue here instead.

* The reproducer does not compile, you cannot have `||` or `&&` in assertion. 
* The issue is real. The lifetime of the lambda that creates the generator ends before the generator's lifetime does, which means that if the generator takes a reference to its capture (the way `filter` currently does with the predicate), it will be a dead reference.
* The test in this PR is a bit lacking. The behaviour you've observed was caused by the difference in stack layout/usage between difference compilers and configurations. A better test would be to create the generator in a function, and then call different function between calls to `next`, to overwrite the stack memory.

src/catch2/generators/catch_generators_adapters.hpp
```diff
@@ -91,7 +92,7 @@ namespace Generators {
 
     template <typename T, typename Predicate>
     GeneratorWrapper<T> filter(Predicate&& pred, GeneratorWrapper<T>&& generator) {
-        return GeneratorWrapper<T>(Catch::Detail::make_unique<FilterGenerator<T, Predicate>>(CATCH_FORWARD(pred), CATCH_MOVE(generator)));
+        return GeneratorWrapper<T>(Catch::Detail::make_unique<FilterGenerator<T, typename std::remove_reference<Predicate>::type>>(CATCH_FORWARD(pred), CATCH_MOVE(generator)));
```
horenmar, line 95:
```suggestion
        return GeneratorWrapper<T>(Catch::Detail::make_unique<FilterGenerator<T, typename std::remove_reference_t<Predicate>>>(CATCH_FORWARD(pred), CATCH_MOVE(generator)));
```

PR #3010, opened 2025-07-30
Add Unscoped Capture Logger

### LINKED ISSUE #2954: UNSCOPED_CAPTURE
It might be useful to have a new [logger](https://github.com/catchorg/Catch2/blob/devel/docs/logging.md) called `UNSCOPED_CAPTURE`, which relates to `CAPTURE` in the same way that `UNSCOPED_INFO` relates to `INFO`. That is, it works like `CAPTURE` but its lifetime is not limited to its scope. This would permit invoking `CAPTURE` inside control flow or inner functions which do not encounter an [assertion](https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md) (like `REQUIRE`) before the end-of-scope, yet _still_ have the output displayed when the calling scope (like a `TEST_CASE`) hits the assertion.

**Motivation**
Currently, `CAPTURE`'s limited lifetime has the same limitations of `INFO`. This means running code like:

```C++
void inner(x) {
    CAPTURE( x );

    // no assertions here, so CAPTURE silent
}

TEST_CASE( "demo" ) {

    int x = 2;
    inner(x);

    FAIL( );
}
```
will not display `x := 2`. Such a use-case is motivated by `inner()` being a customisation of `CAPTURE` which (for example) chooses which variables to display based on compile-time constants. For example:
```C++
template <bool Show>
void runTest() {

    int x = 2;

    // this will never show
    if constexpr (Show) {
        CAPTURE( x );
    }

    FAIL();
}

TEST_CASE( "demoA" ) {

    runTest<true>();
}

TEST_CASE( "demoB" ) {

    runTest<false>();
}
```
Even the `constexpr` branch has a scope, so `CAPTURE`'s output is lost. Above is akin to my situation; I have a templated function used to automate many `TEST_CASE`s (yucky I know - woe scientific software) and I must consult the template parameters to choose which variables `CAPTURE` should show if/when the test subsequently fails.

In principle, I _can_ manually invoke `UNSCOPED_INFO` myself, but then I have to re-implement `CAPTURE`'s lovely printing of nested vectors, for example.

**Description**
A new macro `UNSCOPED_CAPTURE` which works exactly like `CAPTURE`, but which has the scoping behaviour of `UNSCOPED_INFO`.

Riyuzak2510:
Adding Unscoped Capture Logger as a part of [#2954](https://github.com/catchorg/Catch2/issues/2954)

## Description

Added Unscoped Capture Logger Feature as required to print details related to Capture

- Added New Macro UNSCOPED_CAPTURE 

## GitHub Issues

Closes [#2954 ](https://github.com/catchorg/Catch2/issues/2954)


src/catch2/catch_message.cpp
```diff
@@ -112,4 +112,15 @@ namespace Catch {
         m_captured++;
     }
 
+     void Capturer::captureUnscopedValue( size_t index, std::string const& value ) {
+        m_messages[index].message += value;
+        getResultCapture().emplaceUnscopedMessage(Catch::MessageBuilder(
+            m_messages[index].macroName,
+            m_messages[index].lineInfo,
+            m_messages[index].type) << m_messages[index].message);
+        if(index == m_messages.size() - 1){
```
Riyuzak2510: We need to add this as the Capturer Destructor has an assertion at L103

src/catch2/catch_message.hpp
```diff
@@ -85,6 +98,10 @@ namespace Catch {
             captureValue( index, Catch::Detail::stringify(value) );
             captureValues( index+1, values... );
         }
+
+         std::vector<MessageInfo> getMessageDetails() const {
+            return m_messages;
+        }
```
horenmar: This is not used anymore. For next time, this shouldn't be returning the vector by value, but by reference, and be ref-qualified if needed.

src/catch2/catch_message.hpp
```diff
@@ -73,6 +73,19 @@ namespace Catch {
...
+        void captureUnscopedValues( size_t index, T const& value ) {
+            captureUnscopedValue( index, Catch::Detail::stringify( value ) );
+        }
+
+        template<typename T, typename... Ts>
+        void captureUnscopedValues( size_t index, T const& value, Ts const&... values ) {
+            captureUnscopedValue( index, Catch::Detail::stringify(value) );
+            captureUnscopedValues( index+1, values... );
+        }
+
```
horenmar: Let's not duplicate these, and instead add `bool isScoped` parameter to the constructor of `Capturer`, which will be kept to decide whether `captureValue` should create scoped or unscoped msg.

src/catch2/catch_message.cpp
```diff
@@ -112,4 +112,15 @@ namespace Catch {
         m_captured++;
     }
 
+     void Capturer::captureUnscopedValue( size_t index, std::string const& value ) {
+        m_messages[index].message += value;
+        getResultCapture().emplaceUnscopedMessage(Catch::MessageBuilder(
+            m_messages[index].macroName,
+            m_messages[index].lineInfo,
+            m_messages[index].type) << m_messages[index].message);
```
horenmar: No need to copy the value into message first and then write it into the msg builder stream.

src/catch2/catch_message.hpp
```diff
@@ -120,27 +145,33 @@ namespace Catch {
...
+  #define UNSCOPED_CAPTURE( ... ) INTERNAL_CATCH_UNSCOPED_CAPTURE( INTERNAL_CATCH_UNIQUE_NAME(capturer), "UNSCOPED_CAPTURE", __VA_ARGS__ )
+
 
 #elif !defined(CATCH_CONFIG_PREFIX_MESSAGES) && defined(CATCH_CONFIG_DISABLE)
 
   #define INFO( msg )          (void)(0)
   #define UNSCOPED_INFO( msg ) (void)(0)
   #define WARN( msg )          (void)(0)
   #define CAPTURE( ... )       (void)(0)
+  #define UNSCOPED_CAPTURE( ... ) (void)(0)
```
horenmar: Please re-align the empty definitions for the whole block. (And same for the one around line 150)

tests/SelfTest/UsageTests/Message.tests.cpp
```diff
@@ -212,6 +212,38 @@ TEST_CASE( "CAPTURE can deal with complex expressions", "[messages][capture]" )
...
+    for ( int i = 4; i <= 6; i++ ) {
+        unscoped_capture( i );
+    }
+    CHECK( false );
+}
+
+TEST_CASE( "stacks unscoped capture for vector", "[failing][.][unscoped][capture]" ) {
+    {
+        std::vector<int> vec { 7, 8, 9 };
+        unscoped_capture(  vec);
```
horenmar:
```suggestion
        unscoped_capture( vec );
```

src/catch2/catch_message.hpp
Riyuzak2510: done

src/catch2/catch_message.cpp
Riyuzak2510: done

src/catch2/catch_message.hpp
Riyuzak2510: yeah  modified this

src/catch2/catch_message.hpp
Riyuzak2510: Removed it

Riyuzak2510:
@horenmar feel free to take a look

src/catch2/catch_message.hpp
```diff
@@ -65,8 +65,9 @@ namespace Catch {
         std::vector<MessageInfo> m_messages;
         IResultCapture& m_resultCapture;
         size_t m_captured = 0;
+        bool isScoped;
```
horenmar, line 68: Take a second look at how the other class variables are named.

src/catch2/catch_message.cpp
```diff
@@ -106,10 +107,20 @@ namespace Catch {
     }
 
     void Capturer::captureValue( size_t index, std::string const& value ) {
-        assert( index < m_messages.size() );
-        m_messages[index].message += value;
-        m_resultCapture.pushScopedMessage( m_messages[index] );
-        m_captured++;
+        if(isScoped){
+            assert( index < m_messages.size() );
```
horenmar, line 111: The assert should live at the start of the function, both branches access `m_messages[index]`.

src/catch2/catch_message.cpp
```diff
@@ -106,10 +107,20 @@ namespace Catch {
...
+            m_resultCapture.pushScopedMessage( m_messages[index] );
+            m_captured++;
+        } else {
+            getResultCapture().emplaceUnscopedMessage(Catch::MessageBuilder(
+                m_messages[index].macroName,
+                m_messages[index].lineInfo,
+                m_messages[index].type) << value);
+            if(index == m_messages.size() - 1){
+                m_messages.clear();
+            }
```
horenmar, line 122: No need to clear this here, instead check in destructor whether `m_isScoped` is true and change it accordingly.

PR #3015, opened 2025-08-05
Docs: Split Specific Purpose Generators into Random and Range Generators

andrewauclair:
Break specific purpose generators into random and range generator sections. It really tripped me up that the main bullet point ("4 specific purpose generators") stated that the generators were defined in `catch2/generators/catch_generators_random.hpp`. I then missed the fact that the range generator lines state they're actually in `catch2/generators/catch_generators_range.hpp`.


horenmar:
Thanks, that makes more sense.

PR #3017, opened 2025-08-10
Fix color mode detection on FreeBSD by adding platform macro

### LINKED ISSUE #2449: Color mode detection on FreeBSD
**Description**
Allow correct detection of colour mode on FreeBSD when redirecting to file.
Don't print out colour codes when this is done by default.

**Additional context**
`ANSIColourImpl::useImplementationForStream` checks if the stream is being directed and if true sets `useColour` to false.
Here `CATCH_INTERNAL_HAS_ISATTY` is defined only when `CATCH_PLATFORM_LINUX` or `CATCH_PLATFORM_MAC` is true.
Is it possible to allow something like `CATCH_PLATFORM_FREEBSD` as well?

shloka2598:
## Description
Fixed the color mode detection on FreeBSD by adding defined(__FreeBSD__) to the platform check in catch_console_colour.cpp. This enables the use of isatty() on FreeBSD systems to correctly detect when output is redirected to a file. As a result, it prevents ANSI color codes from being printed when they shouldn’t be, making FreeBSD behave consistently with Linux and macOS in terms of console color output.

## GitHub Issues
Closes #2449

### DIFF
--- a/src/catch2/internal/catch_console_colour.cpp
+++ b/src/catch2/internal/catch_console_colour.cpp
@@ -161,7 +161,7 @@ namespace {
 #endif // Windows/ ANSI/ None
 
 
-#if defined( CATCH_PLATFORM_LINUX ) || defined( CATCH_PLATFORM_MAC ) || defined( __GLIBC__ )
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks

Issue #2887, opened 2024-07-11
Remove final keyword from first party reporters


bearbones:
**Description**
I don't know why the first party reporters are "not intended to be inherited from," but for my use case, all I want is a `ConsoleReporter` that doesn't print the test spec filter at the start of each run. To me, the most reasonable option is to inherit from `ConsoleReporter` and override `noMatchingTestCases()` and `testRunStarting()`, but I can't do that when it's `final`. 

I _could_ just edit `ConsoleReporter` in our fork, but I would prefer to not have to keep a version that's slightly out of sync with the base repo.

**Additional context**
My team's custom `main()` generates a ridiculously long filter based on a file that lists excluded tests. This is to accommodate legacy tests that were written with Boost.Test macros. We redefined those macros to map to Catch2 along with some extra legwork for keeping track of suites and constructing case names. Right now it floods the console with about 90k characters.

I also need to print out a special message for our meta-test-runner to recognize at the start and end of runs and cases, and I could do that with an event listener, but it seems more elegant to just override those few `ConsoleReporter` functions.

horenmar:
You will end up with your own reporter either way, and I would rather have you copy paste `ConsoleReporter` into `CompanyConsoleReporterDoNotSteal`, than derive `CompanyConsoleReporterDoNotSteal` from the built in one.

The underlying reason is simple, I do not want to have to consider whether a refactoring to reporters changes their behaviour if some of the functions are overridden and some are not. As an example, right now the ConsoleReporter only responds to `testCaseEnded` event, and ignores partial test cases. If I decide to change this in the future, so that it uses the newer `testCasePartialEnded` (and `testCasePartialStarting` for resets) for more granular control over the output, I do not want to have to think about whether the users have overridden these already, whether they call the parent's implementation for them, etc.

horenmar:
Also I am told that I should ask you if you work at Roblox 😃

bearbones:
Okay, that makes sense.

And to your question, yeah 😊

PR #3025, opened 2025-09-01
Make ConsoleReporter non-final

neobrain:
## Description

Many functions here can be customized without issue.

My primary motivation was to print test cases (without printing each individual test) as they're being executed in a long-running test suite. Custom reporters can't be used for this since the console reporter can't be combined with other reporters. Making ConsoleReporter non-final makes it trivial since I can simply implement `testCaseStarting` myself now.

### DIFF
--- a/src/catch2/reporters/catch_reporter_console.hpp
+++ b/src/catch2/reporters/catch_reporter_console.hpp
@@ -15,7 +15,7 @@ namespace Catch {
     // Fwd decls
     class TablePrinter;
 
-    class ConsoleReporter final : public StreamingReporterBase {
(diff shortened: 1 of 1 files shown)


horenmar:
No. As with every other time this comes up; I would rather you go and copy the reporter into your own one.

neobrain:
Forking >600 lines of code instead of adding a one-line printf through inheritance seems rather excessive, but I understand the concern from the maintenance perspective. Thanks for taking a look!

(For others, here's further reference at https://github.com/catchorg/Catch2/issues/2887#issuecomment-2240004395)

PR #3027, opened 2025-09-10
Added compile time testing support

Teskann:
# As this was not merged, this is available here: https://github.com/Teskann/constest

# Added compile-time testing macros

Discussion: https://discord.com/channels/415851350384181248/529055331880271884

# Description

## Why should I test code at compile-time ?

Testing code during compilation is the only way for ensuring the code does
not lead to [undefined behavior](https://en.cppreference.com/w/cpp/language/ub.html).
This is critical to make your program safe, prevent unexpected crashes and worse
consequences of undefined behavior.

## How ?

Catch2 made it easy to test code at compile-time.

The usage is pretty similar to testing at runtime:

| Runtime macro   | Compile time macro        |
|-----------------|---------------------------|
| `SECTION`       | `CONSTEXPR_SECTION`       |
| `REQUIRE`       | `CONSTEXPR_REQUIRE`       |
| `REQUIRE_FALSE` | `CONSTEXPR_REQUIRE_FALSE` |
| `CHECK`         | -                         |
| `CHECK_FALSE`   | -                         |

```c++
TEST_CASE("My compile time test")
{
    CONSTEXPR_SECTION("std::array subscript and size")
    {
        std::array v = {1, 2, 3, 4, 5};
        CONSTEXPR_REQUIRE( v.size() == 5 );
        CONSTEXPR_REQUIRE( v[2] == 3 );

    }; // <-- Don't forget the semicolon here!
}
```
All code inside the `CONSTEXPR_SECTION` will be evaluated during compilation.
If any of the `CONSTEXPR_REQUIRE` or `CONSTEXPR_REQUIRE_FALSE` fails,
it will cause a **compilation error**.
> [!WARNING]
> You cannot use `REQUIRE` or `CHECK` inside a `CONSTEXPR_SECTION`.
> Be careful not to mistake `CONSTEXPR_REQUIRE` with `STATIC_REQUIRE`.
> They both concern compilation-time testing, but they do not have the same
> purpose.

> [!WARNING]
> You cannot nest `CONSTEXPR_SECTION`s or put a `SECTION` inside a
> `CONSTEXPR_SECTION`

> [!NOTE]
> The code inside the `CONSTEXPR_SECTION` is also evaluated at
> runtime. This way, it remains debuggable, and it contributes to the code
> coverage analysis.

## What can I test ?

You can test anything that can be evaluated in a `constexpr` function.
This will depend on your compiler and the C++ standard you are using.
C++20, C++23 and C++26 improved a lot the support for `constexpr`.
Take a look at the C++ compiler support on [cppreference](https://en.cppreference.com/w/cpp/compiler_support.html)
to see what you can use.
### Investigate from the compiler output
You can check the output of your compiler to find the failing assertion.
The failing line should be highlighted somewhere. If you cannot see any
line:
- Ensure the `CONSTEXPR_SECTION` runs only code that can be evaluated
at compile-time. This will depend on your compiler and the C++ standard you
are using.
- Ensure you did not use unsupported Catch2 macros inside the `CONSTEXPR_SECTION`.
The only supported macros are `CONSTEXPR_REQUIRE` and `CONSTEXPR_REQUIRE_FALSE`.
- Ensure the code you wrote doesn't produce undefined behavior (UB cannot 
compile). Reading a range outside its bounds, dereferencing an invalid pointer, 
reading a variable after it has been destroyed, are widespread undefined
behaviors. They will be all caught by these tests.
## Compile time test report
At runtime, a `CONSTEXPR_SECTION` will add a section in the test report. These sections
provide a way to see what was tested during compilation in the test report.
For instance, given the following test case:
```C++
TEST_CASE("My other compile time test")
{
    SECTION("Any runtime section")
    {
        SECTION("Nested section")
        {
            CONSTEXPR_SECTION("First constexpr section")
            {
                // ...
            };
        };
        
        CONSTEXPR_SECTION("Second constexpr section")
        {
            // ...
        };
        CONSTEXPR_SECTION("Third constexpr section")
        {
            // ...
        }; 
        CONSTEXPR_SECTION("Fourth section")
        {
            // ...
        };           
    }
    
    SECTION("Another runtime section")
    {
        SECTION("Nested runtime section")
        {
            // ...
        }
    }    
}
```

The report would look like this:

- ✅ My other compile time test
    - ✅ Any runtime section
        - ✅ Nested section
            - ✅ [Passed during compilation] First constexpr section
        - ✅ [Passed during compilation] Second constexpr section
        - ✅ [Passed during compilation] Third constexpr section
        - ✅ [Passed during compilation] Fourth constexpr section
    - ✅ Another runtime section
        - ✅ Nested runtime section


src/catch2/internal/catch_constexpr_asserts.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+// SPDX-License-Identifier: BSL-1.0
+#ifndef CATCH_CONSTEXPR_ASSERTS_HPP_INCLUDED
+#define CATCH_CONSTEXPR_ASSERTS_HPP_INCLUDED
+#include <catch2/internal/catch_decomposer.hpp>
+#include <catch2/internal/catch_test_macro_impl.hpp>
+
+namespace Catch {
+    template <typename ExpressionType>
+    void handleExpression( StringRef macroName,
```
Teskann, line 15: I had to wrap this into a function because constructing non literal types in a constexpr function is a C++23 extension. Even if this function is always run in the non-constexpr path that is supposed to trigger the compiler error, it caused build issues on clang even if it's never reached.

It's templated because ExpressionType can be an ExprLhs or a full expression.

src/catch2/internal/catch_constexpr_section.hpp
```diff
@@ -0,0 +1,60 @@
...
+// Distributed under the Boost Software License, Version 1.0.
+//   (See accompanying file LICENSE.txt or copy at
+//        https://www.boost.org/LICENSE_1_0.txt)
+
+// SPDX-License-Identifier: BSL-1.0
+#ifndef CATCH_CONSTEXPR_SECTION_HPP_INCLUDED
+#define CATCH_CONSTEXPR_SECTION_HPP_INCLUDED
+
+#if defined(CATCH_CPP17_OR_GREATER) && __cpp_constexpr >= 201603
+#define CATCH_CONFIG_CONSTEXPR_SECTIONS
```
Teskann, line 12: I think I will need your help for this

src/catch2/internal/catch_constexpr_section.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#ifdef CATCH_CONFIG_CONSTEXPR_SECTIONS
+
+#    define INTERNAL_CATCH_CONSTEXPR_SECTION( ... )                          \
+        INTERNAL_CATCH_SECTION( "[Passed during compilation] " __VA_ARGS__ ) \
+        [[maybe_unused]] const ::Catch::ConstexprSection                     \
+        INTERNAL_CATCH_UNIQUE_NAME(                                       \
+            catch_internal_CompileTimeSection ) = [&]()
+
+#else
```
Teskann, line 53: Tell me if this is acceptable for you, or if you'd rather prefer something else to disable these macros in C++14

src/catch2/internal/catch_constexpr_asserts.hpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+} // namespace Catch
+
+#define INTERNAL_CATCH_CONSTEXPR_ASSERT( macroName, resultDisposition, ... )   \
+    do {                                                                       \
+        CATCH_INTERNAL_START_WARNINGS_SUPPRESSION                              \
+        CATCH_INTERNAL_SUPPRESS_PARENTHESES_WARNINGS                           \
+        const auto expr__InternalCatch__ = Catch::Decomposer() <= __VA_ARGS__; \
+        CATCH_INTERNAL_STOP_WARNINGS_SUPPRESSION                               \
+        if ( expr__InternalCatch__.getResult() !=                              \
```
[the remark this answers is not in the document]
Teskann: And handling expressions is not constexpr so it's perfectly suitable

tests/SelfTest/UsageTests/CompileTimeTests.tests.cpp
```diff
@@ -0,0 +1,107 @@
...
+        };
+        CONSTEXPR_REQUIRE( lambda() == 50 );
+        CONSTEXPR_REQUIRE( lambda() == 500 );
+        SUCCEED( "Silent empty section warning (at runtime, "
+                 "CONSTEXPR_REQUIRE does nothing if passed)" );
+    }
+    SUCCEED( "Silent empty section warning" );
+}
+
+TEST_CASE( "CONSTEXPR_REQUIRE, CONSTEXPR_REQUIRE_FALSE should evaluate "
```
DNKpp, line 104: Isn't this just a matter of tagging it with `[!shouldfai]`?
https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#special-tags

Teskann: Looks like it works, thanks :) I was sure I tried already

src/catch2/internal/catch_constexpr_section.hpp
```diff
@@ -0,0 +1,58 @@
...
+        template <typename Callable>
+        ConstexprSection( Callable callable ) {
+            static_assert(
+                [&] // error: CONSTEXPR_SECTION failure. Check the compiler
+                    // output to find the cause. You can make the section a
+                    // SECTION to investigate at runtime.
+                {
+                    callable();
+                    return true;
+                }(),
```
ZXShady:
```suggestion
static_assert((callable(),true));
```


I don't like lamdbas when a simple expression can do it should be simpler fpr the compiler 

if we are really pendatic we can use `void()` to always use builtin comma

src/catch2/internal/catch_constexpr_section.hpp
```diff
@@ -0,0 +1,58 @@
...
+        INTERNAL_CATCH_SECTION( "[Passed during compilation] " __VA_ARGS__ ) \
+        [[maybe_unused]] const ::Catch::ConstexprSection                     \
+        INTERNAL_CATCH_UNIQUE_NAME(                                          \
+            catch_internal_CompileTimeSection ) = [&]()
+
+#else
+#    define INTERNAL_CATCH_CONSTEXPR_SECTION( ... )                        \
+        static_assert( false, "C++17 is required for CONSTEXPR_SECTION" ); \
+        if ( false ) {                                                     \
+        } else
```
ZXShady: why the else?

couldn't it be `if(false)`?

Teskann: The idea here was to avoid side effects of the macro expansion. If it expands to a simple `if`, user can write an `else` right after the macro usage which should not be valid in my opinion.

It doesn't really matter here as the static assert fires. I can remove it

src/catch2/internal/catch_constexpr_section.hpp
Teskann: Very good suggestion, thanks. I didn't think about this, it's definitely better :+1:

src/catch2/internal/catch_constexpr_section.hpp
Teskann: @ZXShady fixed

src/catch2/internal/catch_constexpr_section.hpp
Teskann: Fixed. I kept a message in the static assert to make the error clearer for the user. Even if the assert technically never fails (it's either true or unevaluable at compile time). At least compilers will spot this line and the users will see the error message in the code.

Regarding the cast to void, I considered it useless as `callable` is always a lambda that does not overload the comma operator.

Thanks for this suggestion !

ZXShady: > Regarding the cast to void, I considered it useless as `callable` is always a lambda that does not overload the comma operator.

You are right

tests/SelfTest/UsageTests/CompileTimeTests.tests.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+    SECTION( "Binary -> Expression should be evaluated only once" ) {
+        int r = 5;
+        auto lambda = [&] {
+            r *= 10;
+            return r;
+        };
+        CONSTEXPR_REQUIRE( lambda() == 50 );
+        CONSTEXPR_REQUIRE( lambda() == 500 );
+        SUCCEED( "Silent empty section warning (at runtime, "
```
theo-dep, line 98: This add an assertion for no actual test to the test case counter. It is shame than `CONSTEXPR_REQUIRE` is not compliant with `SECTION`. A user will write a `SUCCEED` assertion always?

Teskann: Because it doesn't make sense to use `CONSTEXPR_REQUIRE` in sections. They should be used in `CONSTEXPR_SECTIONS`. Otherwise they are not run at compile time.

Using constexpr assertions in both sections and constexpr sections won't report the assertion at runtime if it succeeds. This comes from a limitation of C++17. But the `CONSTEXPR_SECTION` macro implementation adds this `SUCCEEDED` automatically so that you don't have anything to write yourself.

This test uses constexpr asserts in sections because it specifically tests the behavior of the macro.

theo-dep: It is because I am based on the example report which don't write the `SUCCEED` macro 🙂. I understand!

Teskann:
As this PR was not merged, I decided to implement it as an extension library here: https://github.com/Teskann/constest

PR #3028, opened 2025-09-11
Move more of tostring.hpp into source file and remove tuple printing recursion

ZXShady:
I moved more stuff into the .cpp file and formatted the file and I removed recursion from the tuple printer for better compile times and readability.

### DIFF
--- a/src/catch2/catch_tostring.cpp
+++ b/src/catch2/catch_tostring.cpp
@@ -63,6 +63,30 @@ namespace Detail {
         return n;
     }
 
+    std::string formatTimeT(std::time_t time) {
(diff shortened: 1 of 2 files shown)


horenmar:
Let's not reformat everything.

As I understand it, there are three separate changes
1) `catch_strnlen` goes into `.cpp`
2) Part of time-point stringifier goes into `.cpp`
3) Tuple does not recurse to print iself.

Can you provide the PR with just these changes? Alternatively I can pick the changes later.

ZXShady:
Hi sorry, I was away from my computer for this week.

PR #3031, opened 2025-09-24
Fix ReusableStringStream to access the container under the lock

dunhor:
## Description
Commit 582200a made `ReusableStringStream`'s index reservation thread safe, however it's still accessing the `m_streams` vector outside the lock. This makes it so that the `add` call returns the pointer in addition to the index so that the lock doesn't need to get acquired again until destruction.

The issue with accessing `m_streams` outside the lock is that `add` can call `push_back` on the vector, which might re-allocate. If this re-allocation occurs concurrently with anther thread trying to index into this array, you get UB (typically a null pointer read).

### DIFF
--- a/src/catch2/internal/catch_reusable_string_stream.cpp
+++ b/src/catch2/internal/catch_reusable_string_stream.cpp
@@ -12,6 +12,7 @@
 
 #include <cstdio>
 #include <sstream>
+#include <tuple>
 #include <vector>
 
 namespace Catch {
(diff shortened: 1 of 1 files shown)


dunhor:
Hmm. I applied the patch to our build, but now ASan happened to hit a double free. I'll need to investigate this some more.

https://dev.azure.com/msft-wil/3aea0394-d119-4b82-84fa-a5186532efe9/_apis/build/builds/1419/logs/56

dunhor:
> Hmm. I applied the patch to our build, but now ASan happened to hit a double free. I'll need to investigate this some more.
> 
> https://dev.azure.com/msft-wil/3aea0394-d119-4b82-84fa-a5186532efe9/_apis/build/builds/1419/logs/56

Woops, I'm dumb. The thread support needs to be explicitly enabled. Still, this is a legitimate bug since concurrent access to the vector is unsafe as another thread can be calling `push_back` at the same time.

horenmar:
Thanks, I completely missed the potential for resize when figuring out where to take the lock.

PR #3033, opened 2025-09-26
Allow composability of unhandled exception filters on Windows

dunhor:
# Situation
Currently, I'm doing some long overdue investigations of some test reliability. The unfortunate thing is that failures almost exclusively come from CI builds. Some of these failures are crashes. In such cases, it would be _very_ beneficial to get a stacktrace printed out in the output, however this is proving tricky because Catch2 is calling `SetUnhandledExceptionFilter` and overriding the handler I'm trying to install.

# The Change
The change is very simple. If a handler previously existed, Catch2 will invoke it after printing out its output. I've also updated the comment to better reflect that it's returning `EXCEPTION_CONTINUE_SEARCH` even in scenarios where the exception is one that the library cares about.

### DIFF
--- a/src/catch2/internal/catch_fatal_condition_handler.cpp
+++ b/src/catch2/internal/catch_fatal_condition_handler.cpp
@@ -86,22 +86,26 @@ namespace Catch {
         { EXCEPTION_INT_DIVIDE_BY_ZERO, "Divide by zero error" },
     };
 
+    // Since we do not support multiple instantiations, we put these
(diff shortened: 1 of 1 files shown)


horenmar:
Thanks.

It was always supposed to do that, but back when we used VectoredExceptionHandler (?) we would only need to return `EXCEPTION_CONTINUE_SEARCH`. However, VEH was causing issues with Windows impl of ASan, so we moved to `SetUnhandledExceptionFilter`, but didn't modify the logic accordingly.

PR #3041, opened 2025-10-16
Add validation for --benchmark-samples to prevent crash with zero value

### LINKED ISSUE #3040: SIGSEGV when `--benchmark-samples 0`
**Reproduction steps**

Build and run a benchmark with `--benchmark-samples 0`
`./bench --benchmark-samples 0`
```
terminated by signal SIGSEGV (Address boundary error)
```

**Platform information:**
 - OS: **linux**
 - Compiler+version: **GCC v14.3.0**
 - Catch version: **v3.4.0**

chanmyae09:
Fixes #3040 

Summary
Added validation to reject `--benchmark-samples 0` which was causing a segmentation fault.

Changes
- Added 'setBenchmarkSamples' validation function in 'src/catch2/internal/catch_commandline.cpp'
- Function checks if the value is 0 and returns an error message instead of allowing the crash
- Follows the same pattern as existing validators like 'setShardCount'

Testing
Tested with a benchmark test case - now shows error message "Benchmark samples must be greater than 0" instead of crashing with SIGSEGV.

### DIFF
--- a/src/catch2/internal/catch_commandline.cpp
+++ b/src/catch2/internal/catch_commandline.cpp
@@ -189,6 +189,19 @@ namespace Catch {
             config.shardCount = *parsedCount;
             return ParserResult::ok( ParseResultType::Matched );
         };
+        auto const setBenchmarkSamples = [&]( std::string const& samples ) {
+        auto parsedSamples = parseUInt( samples );
+        if ( !parsedSamples ) {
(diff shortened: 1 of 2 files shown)


horenmar:
Needs an automated test.

qerased:
I picked this up and opened a new PR that includes the fix and updated approval baselines #3056

PR #3045, opened 2025-11-07
Fix the help text for the --order command line argument

stk-ableton:
This fixes the `--help` output for the `--order` argument; it was changed to rand in v3.9.0.


horenmar:
Thanks. [At least I remembered to update the docs.](https://github.com/catchorg/Catch2/blob/devel/docs/command-line.md#rand)

PR #3059, opened 2026-01-11
Fix _NullFile for MinGW

dragon-archer:
## Description

Using `/dev/null` on MSys2/MinGW will cause 2 tests to fail, while `NUL` is ok.

```
49/74 Test #49: MultiReporter::CapturingReportersDontPropagateStdOut .........***Failed  Error regular expression found in output. Regex=[.+]  0.02 sec
Unable to open file: '/dev/null'

      Start 50: MultiReporter::NonCapturingReportersPropagateStdout
50/74 Test #50: MultiReporter::NonCapturingReportersPropagateStdout ..........***Failed  Required regular expression not found. Regex=[A string sent to stderr via clog
]  0.02 sec
Unable to open file: '/dev/null'
```

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -483,7 +483,7 @@ set_tests_properties("ErrorHandling::InvalidTestSpecExitsEarly"
     FAIL_REGULAR_EXPRESSION "No tests ran"
 )
 
-if(MSVC)
+if(WIN32)
   set(_NullFile "NUL")
 else()
   set(_NullFile "/dev/null")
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -483,7 +483,7 @@ set_tests_properties("ErrorHandling::InvalidTestSpecExitsEarly"
     FAIL_REGULAR_EXPRESSION "No tests ran"
 )
 
-if(MSVC)
+if(MSVC OR MINGW)
```
horenmar: This should probably be 
```suggestion
if(WIN32)
```

instead. Can you check if that works for you?

dragon-archer: It works for me, and yes it's more portable.

horenmar:
Thanks

PR #3061, opened 2026-01-17
Silence warning -Wconversion in GCC 15.2 and -Wsign-conversion in Clang 21.1.8

hanickadot:
## Description
This small PR add explicit cast from `size_t` to `int` when calling `std::setw(...)` to silence warning produced by GCC 15.2.

This also fixes conversions between sign types in Clang 21.1.8.

### DIFF
--- a/src/catch2/reporters/catch_reporter_console.cpp
+++ b/src/catch2/reporters/catch_reporter_console.cpp
@@ -354,7 +354,8 @@ public:
         }
         tp.m_currentColumn++;
 
-        auto colInfo = tp.m_columnInfos[tp.m_currentColumn];
+        auto colInfo =
+            tp.m_columnInfos[static_cast<size_t>( tp.m_currentColumn )];
         auto padding = (strSize + 1 < colInfo.width)
             ? std::string(colInfo.width - (strSize + 1), ' ')
             : std::string();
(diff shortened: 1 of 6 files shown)


src/catch2/reporters/catch_reporter_console.cpp
```diff
@@ -354,7 +354,8 @@ class TablePrinter {
         }
         tp.m_currentColumn++;
 
-        auto colInfo = tp.m_columnInfos[tp.m_currentColumn];
+        auto colInfo =
+            tp.m_columnInfos[static_cast<size_t>( tp.m_currentColumn )];
```
horenmar, line 358: `tp.m_currentColumn` should be changed to be unsigned instead. The only place where it uses the fact that it is signed is using `-1` as guard value after reset.

rajpratham1 (changes requested):
Thanks for addressing the new GCC 15.2/Clang 21.1.8 conversion warnings. The explicit casts around hashing (unsigned char) and the timer code make sense to me. However, a couple of the other changes appear to suppress warnings rather than address the underlying type mismatch. In particular, I agree with the earlier comment that tp.m_currentColumn should probably become an unsigned index (or use a different sentinel representation) instead of adding a static_cast<size_t> at every access. Similarly, the static_cast<long> used for the vector iterator offset seems less appropriate than using the container's difference_type or revisiting the member type. I'd prefer those underlying type issues to be addressed rather than relying on casts.

PR #3062, opened 2026-01-19
docs: Replace Clara external link with internal documentation

### LINKED ISSUE #2312: Replace link to Clara's repository for doc with our own documentation page
Clara, the command line parser library in Catch2, stopped being a separate project from Catch2 about 3 years ago. Since then it was only developed inside Catch2, while the original upstream repository was made inactive and archived.

Given that Clara-inside-Catch2 has also slowly diverged from the separate version, we should provide its documentation inside the Catch2 repo, and replace the link to upstream docs in the [own-main docs](https://github.com/catchorg/Catch2/blob/devel/docs/own-main.md) with link to our own documentation page for Clara.

dhimasardinata:
## Summary

Replace the link to the archived catchorg/Clara repository in own-main.md with a link to a new internal documentation page.

## Changes

- **NEW** docs/clara.md - Documents the Clara CLI parser API including:
  - Opt for options and flags
  - Arg for positional arguments
  - Parser composition with the | operator
  - Lambda handlers with validation
  - Integration with Catch2's Session

- **MODIFIED** docs/own-main.md - Updated link from archived external repo to internal docs

## Background

Clara has been developed as part of Catch2 for several years now (as noted in the issue), and the external repository at catchorg/Clara is no longer maintained and has been archived.

Fixes #2312

### DIFF
--- a/docs/clara.md
+++ b/docs/clara.md
@@ -1,0 +1,168 @@
+<a id="top"></a>
+# Clara - Command Line Parser
+
+Clara is the command line parser used internally by Catch2. It was originally
+a separate project, but has been integrated into Catch2 and is now developed
+as part of the Catch2 codebase.
+
+**Contents**<br>
+[Overview](#overview)<br>
+[Options](#options)<br>
+[Arguments](#arguments)<br>
+[Combining Parsers](#combining-parsers)<br>
+[Using with Catch2's Session](#using-with-catch2s-session)<br>
+
+## Overview
+
+Clara provides a simple, composable way to define command line parsers.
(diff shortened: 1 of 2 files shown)


rajpratham1 (changes requested):
Replacing the external Clara link with internal documentation is definitely the right direction, and keeping the documentation within the repository improves long-term accessibility. My only question is whether the new docs/clara.md is derived from the original Clara documentation and verified against the current implementation. Since this adds a fairly comprehensive API reference, I'd like to make sure it won't drift from the code over time.

PR #3063, opened 2026-01-24
Fix: Use unique /Catch2 prefix for reproducible builds

### LINKED ISSUE #3042: add substitute-path to gdbinit for prefix-map (reproducible build), and pick a unique prefix
`CATCH_ENABLE_REPRODUCIBLE_BUILD` adds a `-ffile-prefix-map=${CATCH_DIR}/=` compile option. It would be nice for the bundled gdbinit file to add a corresponding `set substitute-path` line for this. A potential issue is that `set substitute-path` works like this:

> Define a source path substitution rule, and add it at the end of the current list of existing substitution rules. If a rule with the same from was already defined, then the old rule is also deleted.

If I'm reading the compile option correctly, Catch2 appears to be mapping to an empty path, so I presume that adding a `substitute-path` for it would mean no other `substitute-path`'s `from` can be for the empty path. If that's the case, a potential mitigation would be to pick something more unique, like `/Catch2`. That would mean changing the `-ffile-prefix-map` destination to `/Catch2` as well.

arryansh5:
Fixes #3042

## Summary

Changes the reproducible build path mapping to use `/Catch2` prefix instead of empty string, avoiding conflicts when multiple libraries use reproducible builds.

## Changes

- Updated `-ffile-prefix-map` to use `/Catch2/` prefix in `CMake/CatchMiscFunctions.cmake`
- Added path substitution instructions to `extras/gdbinit` and `extras/lldbinit`
- Documented the feature in `docs/ci-and-misc.md`
- Added release notes

## Problem Solved

When both Catch2 and user projects enable `CATCH_ENABLE_REPRODUCIBLE_BUILD`, debuggers (GDB/LLDB) can only map one empty prefix path at a time, causing conflicts. Using a unique `/Catch2/` prefix allows both to coexist.

Users need to add to their debugger config:
```gdb
set substitute-path /Catch2 /path/to/Catch2

### DIFF
--- a/CMake/CatchMiscFunctions.cmake
+++ b/CMake/CatchMiscFunctions.cmake
@@ -116,6 +116,6 @@ endfunction()
 function(add_build_reproducibility_settings target)
   # Make the build reproducible on versions of g++ and clang that supports -ffile-prefix-map
   if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
-    add_cxx_flag_if_supported_to_targets("-ffile-prefix-map=${CATCH_DIR}/=" "${target}")
+    add_cxx_flag_if_supported_to_targets("-ffile-prefix-map=${CATCH_DIR}/=/Catch2/" "${target}")
   endif()
 endfunction()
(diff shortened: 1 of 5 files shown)


rajpratham1 (approved):
This looks like a sensible improvement. Using a unique /Catch2 prefix instead of stripping the path entirely preserves reproducible builds while avoiding path collisions when multiple projects use -ffile-prefix-map. I also appreciate that the debugger helper scripts and documentation were updated to explain the required source mapping, along with the release notes. The change is small, well documented, and low risk.

PR #3066, opened 2026-01-31
Update Catch2 library settings for MSVC and C++ standard

### LINKED ISSUE #2894: Compile Time Improvements
**Description**
Sort the sources in the src/CmakeLists.txt by compile times. When doing parallel builds in conjunction with unity builds it makes sense to have the biggest sources first.

**Additional context**
When building the lib, msvc generated projects should specify c++17. This is because, for c++20 msvc does a prepass search for module dependencies but this can double the compile times and it doesn't need to do this.  

If we were _really serious_ about improving compile times...
Instead of just compiling all files individually or using cmake's unity build generated files, I would create 4 source files that included the actual sources. This would allow maintainers to distribute the sources so that each custom source file would compile in roughly the same time. And the reason to use 4 source files is coz most people have at least 4 threads.

AbhishekkSaini:
## Description

Improves Catch2 compilation times by enabling some standard build optimizations that were previously disabled.

The main changes are:
- Turn on MSVC's `/MP` flag to compile files in parallel across CPU cores
- Use C++17 instead of defaulting to C++20 (MSVC wastes time scanning for modules in C++20 mode even though we don't use them)
- Enable unity builds with a batch size of 8 files per unit

I tested a few different batch sizes and found that 8 works well - higher values like 25 caused some operator overload ambiguity issues during compilation, but 8 seems to avoid those while still giving good speedup.

On my machine (8 cores, MSVC), build time went from about 70 seconds down to around 10-15 seconds, so it's a pretty substantial improvement. The `/MP` flag only affects MSVC builds, so other compilers shouldn't see any difference there, but they should still benefit from the unity build settings.

## GitHub Issues

Fixes #2894

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -355,6 +355,21 @@ add_library(Catch2 ${ALL_FILES})
 if(CATCH_ENABLE_REPRODUCIBLE_BUILD)
   add_build_reproducibility_settings(Catch2)
 endif()
+
+if(MSVC)
+    target_compile_options(Catch2 PUBLIC /MP)
+endif()
+
+target_compile_features(Catch2 PRIVATE cxx_std_17)
+
+# Batch size 8 avoids operator<< conflicts while maintaining speed
+set_target_properties(Catch2
+    PROPERTIES
+        UNITY_BUILD ON
+        UNITY_BUILD_BATCH_SIZE 8
+)
+
+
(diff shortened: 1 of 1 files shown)


horenmar:
No.

1) Don't just bump C++ standard in random PR
2) `/MP` is compiler-instance local, it's failure mode is hilarious oversubscription if you build multiple projects in parallel.

PR #3068, opened 2026-02-11
Fix Clara link in opensource users list

snehalpatil88:
This PR updates the Clara link in docs/opensource-users.md
to point to the current Catch2 documentation site.


horenmar:
How exactly did you come up with that URL?

PR #3081, opened 2026-03-22
JUnit reporter: emit only first failure per testcase

### LINKED ISSUE #1919: JUnit reporter allows multiple failures per test case
**Describe the bug**
JUnit reporter can emit multiple failure entries in the case of multiple `CHECK` failures. This does not conform to the JUnit specification: http://junit.sourceforge.net/doc/faq/faq.htm#tests_12

> Why does JUnit only report the first failure in a single test?
> Reporting multiple failures in a single test is generally a sign that the test does too much, compared to what a unit test ought to do.

As a result, some tools that might attempt to parse the JUnit output from Catch2 will fail. I am seeing https://github.com/gastlygem/junitparser/blob/master/junitparser/junitparser.py fail with `raise JUnitXmlError("Only one result allowed per test case.")`.

**Expected behavior**
To conform to the behavior of JUnit, only the first failure should be output.

**Reproduction steps**
```
#define CATCH_CONFIG_MAIN
#include "catch.hpp"

TEST_CASE("errors")
{
	CHECK(false);
	CHECK(false);
}
```
launch with `-r junit -o junit.xml`
```
<testsuites>
  <testsuite name="Project1.exe" errors="0" failures="2" tests="2" hostname="tbd" time="0.001193" timestamp="2020-04-29T00:00:20Z">
    <testcase classname="Project1.exe.global" name="errors" time="0.000377">
      <failure message="false" type="CHECK">
FAILED:
  CHECK( false )
at e:\dev\catch2junitbug\project1\main.cpp(6)
      </failure>
      <failure message="false" type="CHECK">
FAILED:
  CHECK( false )
at e:\dev\catch2junitbug\project1\main.cpp(7)
      </failure>
    </testcase>
    <system-out/>
    <system-err/>
  </testsuite>
</testsuites>

```

**Platform information:**
 - OS: **Windows 10**
 - Compiler+version: **MSVC toolchain v141**
 - Catch version: **2.12.1**

Krishnayan-Bhatt:
## Summary

The JUnit XML reporter previously emitted multiple `<failure>`, `<error>`, or `<skipped>` elements for a single `<testcase>` when several assertions failed (e.g. multiple `CHECK`s). Common JUnit consumers assume **at most one** such child per testcase, which breaks tools like [junitparser](https://github.com/gastlygem/junitparser) with errors such as “Only one result allowed per test case.”

This change tracks whether a result element was already written for the current testcase section and **suppresses further assertion result elements** after the first.

## Related issue

Fixes #1919

## Testing

- `python tools/scripts/approvalTests.py <path-to-SelfTest> <temp-dir>` — updated `junit.sw` / `junit.sw.multi` baselines
- `ctest -C Debug -R RunTests --test-dir <build-dir>`

### DIFF
--- a/src/catch2/reporters/catch_reporter_junit.cpp
+++ b/src/catch2/reporters/catch_reporter_junit.cpp
@@ -218,12 +218,13 @@ namespace Catch {
             // events and write those out appropriately.
             xml.writeAttribute( "status"_sr, "run"_sr );
 
+            bool resultElementEmitted = false;
             if (sectionNode.stats.assertions.failedButOk) {
                 xml.scopedElement("skipped")
                     .writeAttribute("message", "TEST_CASE tagged with !mayfail");
             }
 
-            writeAssertions( sectionNode );
+            writeAssertions( sectionNode, resultElementEmitted );
 
 
             if( !sectionNode.stdOut.empty() )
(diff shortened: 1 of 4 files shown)


src/catch2/reporters/catch_reporter_junit.cpp
```diff
@@ -218,12 +218,13 @@ namespace Catch {
             // events and write those out appropriately.
             xml.writeAttribute( "status"_sr, "run"_sr );
 
+            bool resultElementEmitted = false;
```
horenmar: Why is this outside of `writeAssertions`?

Krishnayan-Bhatt: You’re right — it doesn’t need to be here anymore. I’ll inline it in writeAssertions (and simplify the signature) in a follow-up commit.

Krishnayan-Bhatt: Done — I moved `resultElementEmitted` into `writeAssertions` so it’s only scoped to assertion output. Latest commit: 224a8c29.

horenmar:
You need either to

1) get LLM that makes better changes
2) guide it towards better output

As is, this doesn't pass the bar for useful contribution.

PR #3086, opened 2026-03-28
Fix counting of UTF-8 codepoints in TextFlow

### LINKED ISSUE #1022: Incorrect line wrapping when test case name is consisted by non-ascii characters
## Description

This is an enhancement request.
I want to use UTF-8 non-ascii characters as test case name,
but Catch reporting system is seems not to support them on line wrapping.
Here is output when bug reproducts.
```
main is a Catch v1.10.0 host application.
Run with -? for options

_「ではみなさんは、そういうふうに川だと言われたり、�-
�の流れたあとだと言われたりしていた、このぼんやりと�-
�いものがほんとうは何かご承知ですか」
main.cpp:4
...............................................................................

main.cpp:5: FAILED:
  REQUIRE( 0 )

test cases: 1 | 1 failed
assertions: 1 | 1 failed
```

### Steps to reproduce

Here is a minimum program to reproduce.
"_" at heading of test case name is needed to reproduce.
```c
#define CATCH_CONFIG_MAIN  // This tells Catch to provide a main() - only do this in one cpp file
#include "catch.hpp"

TEST_CASE( "_「ではみなさんは、そういうふうに川だと言われたり、乳の流れたあとだと言われたりしていた、このぼんやりと白いものがほんとうは何かご承知ですか」" ) {
    REQUIRE( 0 );
}
```

### Extra information

* Catch version: **v1.10.0**
* Operating System: **CentOS 7**
* Compiler+version: **gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-11)**

jhasse:
## Description

For line-wrapping bytes were counted instead of codepoints. This resulted in line-breaks being inserted at the wrong position and also breaking UTF-8 characters.

Maybe AnsiSkippingString should be renamed to Utf8SkippingString now?

## GitHub Issues
Fixes #1022.

### DIFF
--- a/src/catch2/internal/catch_textflow.cpp
+++ b/src/catch2/internal/catch_textflow.cpp
@@ -26,6 +26,10 @@ namespace {
         return std::memchr( chars, c, sizeof( chars ) - 1 ) != nullptr;
     }
 
+    bool isUtf8ContinuationByte( char c ) {
+        return ( static_cast<unsigned char>( c ) & 0xC0 ) == 0x80;
+    }
+
 } // namespace
 
 namespace Catch {
@@ -52,6 +56,11 @@ namespace Catch {
                 if ( it != m_string.end() ) {
                     ++m_size;
                     ++it;
+                    // Skip UTF-8 continuation bytes
+                    while ( it != m_string.end() &&
+                            isUtf8ContinuationByte( *it ) ) {
+                        ++it;
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


horenmar:
> Maybe AnsiSkippingString should be renamed to Utf8SkippingString now?

`Utf8AndAnsiControlCodesSkippingString` 🙃 

Dumb jokes aside, historically Catch2 did not try to handle non-ascii text (with minor concessions here and there), because of the sad state of C++ support for doing so. I think this is unobtrusive-enough to be worth it though.

Thanks

Issue #2925, opened 2024-10-26
`REQUIRE(co_await expr)` unexpectedly repeats execution in GCC


Hackerl:
**Describe the bug**
`REQUIRE(co_await expr)` unexpectedly repeats execution in GCC.

**Expected behavior**
The co_await expression should only be executed once

**Reproduction steps**
https://godbolt.org/z/h6jshrYs8

```cpp
#include <iostream>
#include <coroutine>
#include <https://raw.githubusercontent.com/catchorg/Catch2/refs/tags/v3.7.1/extras/catch_amalgamated.hpp>
#include <https://raw.githubusercontent.com/catchorg/Catch2/refs/tags/v3.7.1/extras/catch_amalgamated.cpp>

struct promise;

struct coroutine : std::coroutine_handle<promise> {
    using promise_type = ::promise;
};

struct promise {
    coroutine get_return_object() { return {coroutine::from_promise(*this)}; }
    std::suspend_never initial_suspend() noexcept { return {}; }
    std::suspend_never final_suspend() noexcept { return {}; }

    void return_void() {
    }

    void unhandled_exception() {
    }
};

struct Awaitable {
    bool await_ready() {
        static int counter{};
        std::cout << ++counter << std::endl;
        return true;
    }

    void await_suspend(std::coroutine_handle<>) {
    }

    bool await_resume() {
        return true;
    }
};

coroutine test() {
    REQUIRE(co_await Awaitable{});
}

TEST_CASE("test") {
    test();
}
```

**Platform information:**

 - OS: **Any**
 - Compiler+version: **GCC v14.x.x**
 - Catch version: **v3.7.1**

**Additional context**
In GCC, `__builtin_constant_p(co_await expr)` is executed unexpectedly.
`while ((void) 0, (false) && static_cast<const bool &>(!!(co_await expr)))` is also.
I believe this is a compiler bug, but is there a way to avoid it with `catch`?

horenmar:
Well, that's not the funniest bug with `__builtin_constant_p` I've seen*, but this is the first time GCC has one. Currently there isn't a configuration option to disable this, but seeing this, I think that it is either time to add one, or to remove that check completely.

Have you reported this to GCC bugzilla?

\* That goes to IBM XL, which would generate a call to the destructor of a temporary, without generating the constructor 😃

horenmar:
GCC already has two bugs for this upstream

[...]

Issue #2591, opened 2022-12-09
co_await expression inside REQUIRE may be run twice (GCC)


alvinhochun:
**Describe the bug**

If a `REQUIRE` assertion contains a `co_await` expression, it may be executed twice.

**Expected behavior**

The expression shouldn't run twice.

**Reproduction steps**

(I'd hope to present a small self-contained example but I don't have one for something that involves coroutines, so...)

1. Install MSYS2
2. In MSYS2, install: `pacman -S mingw-w64-ucrt-x86_64-toolchain mingw-w64-ucrt-x86_64-cmake mingw-w64-ucrt-x86_64-ninja git`
3. Open MSYS2 MinGW UCRT x64 command line
4. Download the code from this commit: https://github.com/alvinhochun/cppwinrt/tree/20e5eb3a440a31913f294e5b4b3e759c830fe456
5. Run:
```console
$ mkdir cppwinrt-build
$ cd cppwinrt-build
$ cmake ../cppwinrt -GNinja -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_COLOR_DIAGNOSTICS=TRUE \
  -DDOWNLOAD_WINDOWSNUMERICS=TRUE \
  -DUSE_ANSI_COLOR=TRUE
$ ninja test_old
$ test/test_old  'async\, resume_on_signal'
```

**Platform information:**
 - OS: **Windows 10**
 - Compiler+version: **g++.exe (Rev6, Built by MSYS2 project) 12.2.0**
 - Catch version: **v2.13.9**

**Additional context**

The problematic line is https://github.com/alvinhochun/cppwinrt/blob/20e5eb3a440a31913f294e5b4b3e759c830fe456/test/old_tests/UnitTests/async.cpp#L1572

```c++
REQUIRE(true == co_await resume_on_signal(signal, 1s)); // should eventually succeed
```

The preprocessed code looks like this:

```c++
       do { (void)__builtin_constant_p(
       true == co_await resume_on_signal(signal, 1s)
       ); Catch::AssertionHandler catchAssertionHandler( "REQUIRE"_catch_sr, ::Catch::SourceLineInfo( "D:/dev/mingw-winrt/cppwinrt/test/old_tests/UnitTests/async.cpp", static_cast<std::size_t>( 1572 ) ), "true == co_await resume_on_signal(signal, 1s)", Catch::ResultDisposition::Normal ); try {
        catchAssertionHandler.handleExpr( Catch::Decomposer() <= 
[...]

PR #3092, opened 2026-05-02
Work around GCC co_await bugs in test macros

vasama:
Related GCC issues:
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101027
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116775

Related Catch2 issues:
* https://github.com/catchorg/Catch2/issues/2591
* https://github.com/catchorg/Catch2/issues/2925

Use a volatile boolean to guard the unevaluated test expression instead of `CATCH_INTERNAL_IGNORE_BUT_WARN` (based on `__builtin_constant_p`, now removed).

I did not remove `CATCH_CONFIG_USE_BUILTIN_CONSTANT_P`, though it now seems to be unnecessary. It's unclear to me if removing it would break user configurations.

### DIFF
--- a/src/catch2/internal/catch_compiler_capabilities.hpp
+++ b/src/catch2/internal/catch_compiler_capabilities.hpp
@@ -130,26 +130,6 @@
 
 #endif // __clang__
 
-// As of this writing, IBM XL's implementation of __builtin_constant_p has a bug
-// which results in calls to destructors being emitted for each temporary,
-// without a matching initialization. In practice, this can result in something
-// like `std::string::~string` being called on an uninitialized value.
-//
-// For example, this code will likely segfault under IBM XL:
-// ```
-// REQUIRE(std::string("12") + "34" == "1234")
-// ```
-//
-// Similarly, NVHPC's implementation of `__builtin_constant_p` has a bug which
-// results in calls to the immediately evaluated lambda expressions to be
-// reported as unevaluated lambdas.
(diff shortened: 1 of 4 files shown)


vasama:
I had to add a variable initialiser in a test due to a now detected use of an uninitialised variable. Looks like the existing solution based on `__builtin_constant_p` was preventing some warnings from coming through.

rajpratham1 (approved):
This looks like a nice simplification. Replacing the previous __builtin_constant_p/false && expression machinery with the volatileFalse pattern keeps expressions type-checked without evaluating them and avoids the GCC coroutine issue. The implementation is simpler and removes a fair amount of compiler-specific macro complexity. My only question is whether the removed IBM XL/NVHPC workarounds are still unnecessary with this approach, but assuming that's been verified, this looks good to me.

SagarJadhav30 (commented):
intresting

SagarJadhav30 (commented):
really like it

vasama:
Benchmarking results as reported on Discord:
* No meaningful difference for runtime/compile perf of debug builds
* 3% slowdown for compilation time in release builds (not great, not terrible)
* 6% slowdown for runtime of assertions in slow path (interestingly, only about 1% in fast path)
* Worse binary sizes, but not enough to be meaningful

I'm planning on looking into this and to hopefully improve these numbers. I've just been too busy during the summer, and it's unlikely that I'll be able to get back to this until some time in august.

PR #3093, opened 2026-05-03
Exclude std::optional from generic range StringMaker (C++26 P3168 fix)

zrno:
## Description

C++26's [P3168](https://wg21.link/P3168) ("Give std::optional Range Support") makes `std::optional<T>` model `std::ranges::range`. As a result, `Catch::is_range<std::optional<T>>::value` becomes `true` under C++26, and the generic range `StringMaker` partial specialization (`catch_tostring.hpp`) starts matching `StringMaker<std::optional<T>>`. This collides with:

1. Catch2's own dedicated `StringMaker<std::optional<T>>` (when `CATCH_CONFIG_ENABLE_OPTIONAL_STRINGMAKER` is enabled), and
2. Any user-provided `StringMaker<std::optional<T>>` partial specialization.

Both partial specializations deduce to the same `StringMaker<std::optional<T>, void>` instantiation, which compilers reject as ambiguous:

```
error: ambiguous template instantiation for 'struct Catch::StringMaker<std::optional<X>, void>'
```

The existing `ToStringOptional.tests.cpp` baseline tests already fail to build with GCC 16 in `-std=c++26` mode for this reason.

## Approach

`is_range` itself is left honest — `std::optional<T>` genuinely is a range, and downstream code keying off the trait should keep working. Instead, the generic range `StringMaker` is constrained via a new `Detail::is_optional` trait so it defers to optional-aware `StringMaker`s. This:

- Fixes Catch2's own `StringMaker<std::optional<T>>` (when the macro is enabled),
- Lets users keep writing `StringMaker<std::optional<T>>` partial specializations without conflict,
- Falls through to the unstreamable fallback when no optional-aware `StringMaker` is provided (matching pre-C++26 behavior).

The same opt-out pattern already exists in the file for managed types (`T^`).

## Test plan

- [x] Reproduced the failure: stashed the patch and confirmed `SelfTest` fails to build under `-std=c++26` with GCC 16.1.1 due to ambiguous instantiation on the existing `std::optional<std::string>` test.
- [x] Applied the patch: `SelfTest` builds clean under `-std=c++26`.
- [x] `[optional]`-tagged tests pass (29 assertions, 7 cases).
- [x] Full `SelfTest` run passes: 499 passed, 7 skipped, 16 expected-failures.
- [x] `tools/scripts/approvalTests.py` reports `Results matched` for every reporter format.
- [x] Added a regression test `std::optional<NonStreamable> -> toString does not collide with range StringMaker` exercising the path that was ambiguous in C++26.

### DIFF
--- a/src/catch2/catch_tostring.hpp
+++ b/src/catch2/catch_tostring.hpp
@@ -25,6 +25,10 @@
 #include <string_view>
 #endif
 
+#ifdef CATCH_CONFIG_CPP17_OPTIONAL
+#include <optional>
+#endif
+
 #ifdef _MSC_VER
 #pragma warning(push)
 #pragma warning(disable:4180) // We attempt to stream a function (address) by const&, which MSVC complains about but is harmless
@@ -481,6 +485,16 @@ namespace Catch {
 
         template <typename T>
         struct is_range_impl<T, void_t<decltype(begin(std::declval<T>()))>> : std::true_type {};
+
+        // Used to exclude std::optional from the generic range StringMaker so it
+        // does not collide with the dedicated optional StringMaker (or a
+        // user-provided one). std::optional models std::ranges::range since
+        // C++26 (P3168), so is_range<std::optional<T>> is true and both partial
+        // specializations would otherwise match StringMaker<std::optional<T>>.
+        template <typename T> struct is_optional : std::false_type {};
+#if defined(CATCH_CONFIG_CPP17_OPTIONAL)
+        template <typename T> struct is_optional<std::optional<T>> : std::true_type {};
+#endif
     } // namespace Detail
 
     template <typename T>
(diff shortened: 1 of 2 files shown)


zrno:
As this is my first contribution, please let me know if any changes are needed. I experienced this error this morning after trying the new GCC16.1 with enabled c++26 support on our codebase.

horenmar:
I don't like this change, at all. The underlying issue is that partial
specializations of `StringMaker` are ambiguous with each other. Here the
problem is the first party partial specialization for `std::optional<T>`
started to overlap with the first party specialization for range-like types,
however it would also cause issues in situations like this:

```cpp
template <typename T>
struct templated_range {
    int_iterator begin() const { return int_iterator{ 1 }; }
    int_iterator end() const { return {}; }
};

template <typename T>
struct StringMaker<templated_range<T>> {
    static std::string convert(templated_range<T> const&) {
        return "custom stringification";
    }
};
```

-----

For now, I am going to merge 5a54efd98b5136d8549730b8ca3227654f73a7b1 as quick fix, and then later implement full fix for user provided partial specializations.

PR #3094, opened 2026-05-05
Fix Wconversion warning in catch_reporter_helpers.cpp

PB-Ole:
## Description

When compiling Catch2 with mingw and -Wconversion flag enabled the `catch_reporter_helpers.cpp` file fails to build since `maxTagCountLen` is of type `size_t` and used as input for `std::setw` which expects an `int` as input.

This simply fixes the type of `maxTagCountLen` which is used nowhere else except for setting the stream width.

## GitHub Issues

I did not find an issue mentioning this. If needed, I can create one.


horenmar:
Thanks

PR #3096, opened 2026-05-07
Avoid a potential underflow when iterating over invalid inputs

dirkmueller:
The forward iteration logic already bounds-check for m_it != m_string->end(), do the same for the backward iteration. The issue with the assert is that the assert() might not be compiled in, and it is happening after the dereference, so it was too late.

### DIFF
--- a/src/catch2/internal/catch_textflow.cpp
+++ b/src/catch2/internal/catch_textflow.cpp
@@ -148,8 +148,8 @@ namespace Catch {
                 m_it--;
             }
             // Skip back over UTF-8 continuation bytes to the leading byte
-            while ( isUtf8ContinuationByte( *m_it ) ) {
-                assert( m_it != m_string->begin() );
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks, but the change should have tests.

dirkmueller:
> Thanks, but the change should have tests.

testcase added.

dirkmueller:
ok, the testcase is failing some check, no idea what to do now. any hints?

horenmar:
To quote from [contributing.md](https://github.com/catchorg/Catch2/blob/devel/docs/contributing.md#testing-your-changes).

> If you added new tests, you will likely see ApprovalTests failure. After you check that the output difference is expected, you should run tools/scripts/approve.py to confirm them, and include these changes in your commit.

Note that I am not sure tests with broken unicode will have portable serialization, in which case you need to add `[approvals]` tag to the `TEST_CASE`, which excludes the test from approval testing.

PR #3101, opened 2026-05-14
Change Catch2 library from static to shared

nater1983:
Change Catch2 library from static to shared. This also add's to the version to the library.

### DIFF
--- a/src/catch2/meson.build
+++ b/src/catch2/meson.build
@@ -361,11 +361,13 @@ if ((host_machine.system() == 'android') or
   catch2_dependencies += log_dep
 endif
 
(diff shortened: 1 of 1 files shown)


rajpratham1 (changes requested):
Thanks for the contribution. While this enables shared library builds, it also changes the default behavior for every Meson user by replacing static_library() with shared_library(). That is a packaging and compatibility change rather than a simple implementation detail. It would be better to preserve user choice by using Meson's library() abstraction (respecting default_library) or by introducing a configurable option instead of hardcoding shared libraries. As written, this is a breaking change without sufficient justification.

PR #3111, opened 2026-05-17
docs: explain when TEST_CASE runs without entering a SECTION (fixes #552)

### LINKED ISSUE #552: When are test cases run without sections?
I am a bit unclear on when the code in a TEST_CASE is run when there are nested SECTIONs. I can't find anything about the details in the documentation so I am not sure what the intended action is here and whether there is a bug or not.

Here is some example code:
```{c++}
TEST_CASE("run once with section") {
    printf("a0\n");
    SECTION("foo") {
        printf("a1\n");
    }
    printf("a2\n");
}

TEST_CASE("run twice with and without section") {
    int error = 0;

    printf("b0\n");
    try {
        printf("b1\n");
        SECTION("bar") {
            printf("b2\n");
            throw 1;
        }
        printf("b3\n");
    } catch (int e) {
        printf("b4\n");
        error = e;
    }
    printf("b5\n");

    REQUIRE(error > 0);
}
```

This prints out:
```
a0
a1
a2
b0
b1
b2
b4
b5
b0
b1
b3
b5
```
indicating that the first example was only run once, but the second was run twice, once without and once with the `SECTION("bar")` code.

See also #191.

wyf027:
## Summary
- Document that Catch2 re-enters each `TEST_CASE` once per leaf `SECTION` path
- Explain the single-sibling vs multi-sibling cases (including examples from #552)
- Add FAQ entry linking to the new section in `test-cases-and-sections.md`

## Test plan
- [x] Doc-only change

Fixes #552

### DIFF
--- a/docs/test-cases-and-sections.md
+++ b/docs/test-cases-and-sections.md
@@ -2,6 +2,7 @@
 # Test cases and sections
 
 **Contents**<br>
+[How many times does a `TEST_CASE` run?](#how-many-times-does-a-test_case-run)<br>
 [Tags](#tags)<br>
 [Tag aliases](#tag-aliases)<br>
 [BDD-style test cases](#bdd-style-test-cases)<br>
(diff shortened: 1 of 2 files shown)


docs/test-cases-and-sections.md
```diff
@@ -30,6 +31,68 @@ executable.**
...
+  runs **once per sibling**, plus an additional run that **does not enter any
+  of those siblings** (the path that skips them). That extra run is what
+  makes code after a `SECTION` block execute both with and without entering
+  the section.
+
+Example from [issue #552](https://github.com/catchorg/Catch2/issues/552):
+
+```c++
+TEST_CASE("run once with section") {
+    printf("a0\n");
```
offa, line 59: Wouldn't it be better at this point to use one of the logging macros or at least C++? At first glance the example looks more like C code than C++.

docs/test-cases-and-sections.md
```diff
@@ -30,6 +31,68 @@ executable.**
...
+    }
+    printf("a2\n");
+}
+```
+
+prints `a0`, `a1`, `a2` once, because there is only one leaf section (`foo`).
+
+```c++
+TEST_CASE("run twice with and without section") {
+    int error = 0;
```
offa, line 71: Fairly complex to follow for an example. 😅

Issue #3043, opened 2025-10-31
UnorderedRangeEquals calls Comparator function with different argument orders


gfrontera:
**Describe the bug**
When using UnorderedRangeEquals to compare two ranges of different types, the Comparator function is called assuming the order of arguments does not matter.

Let's assume that I want to compare a range of elements of type A against another range of elements of type B. I have implemented a Comparator function:
bool compareAB(const A& a, const B& b);

However, when providing that Comparator function to UnorderedRangeEquals, compilation will fail because it calls the Comparator function as compareAB(a, b) as well as compareAB(b, a), and it expects the Comparator function to work either way.

**Expected behavior**
UnorderedRangeEquals should call the Comparator function providing always the arguments in the same order: the first argument should correspond to an element in the first range, and the second element to an element in the second range.

**Reproduction steps**
This simple program triggers the error:
```
#include <catch2/catch_test_macros.hpp>
#include <catch2/matchers/catch_matchers.hpp>
#include <catch2/matchers/catch_matchers_range_equals.hpp>

struct A {
    int i;
};

struct B {
    int j;
};

TEST_CASE("Test") {
    std::vector<A> va = {{1},{2},{3},{4}};
    std::vector<B> vb = {{4},{2},{3},{1}};

    CHECK_THAT(va, Catch::Matchers::UnorderedRangeEquals(vb,
            [](auto v1, auto v2) {return v1.i == v2.j;}));
}
```
See in Compiler Explorer: https://godbolt.org/z/ajnb9468s

**Platform information:**

 - OS: **Ubuntu 24.04**
 - Compiler+version: **Clang 20.1.2**
 - Catch version: **v3.11.0**

horenmar:
Hmmm, I see why you'd want this, but I don't think it is possible. The underlying code is a reimplementation of `is_permutation`, and it requires the custom comparator to be able to stand in for `operator==(A, A)`, `operator==(A, B)`, and `operator==(B, A)`. The last can be replaced by wrapping `operator==(A, B)` in a lambda that reverses the arguments, but I don't see a way around the self-comparisons without changing the algorithm to be completely different.

This in effect means that the two types have to be identical, or the comparator has to provide overloaded to support multiple different argument types, at which point asking for the `(B, A)` overload is not much extra effort.

[This is also reflected in the stdlib's `is_permutation`](https://en.cppreference.com/w/cpp/algorithm/is_permutation.html):

[...]

PR #3113, opened 2026-05-22
Keep UnorderedRangeEquals comparator argument order

puneetdixit200:
## Description
- match unordered ranges with a matched-element bitmap instead of `is_permutation` counting
- keep custom predicates called as `(actual_element, expected_element)` for every comparison
- add a regression test for different actual/expected element types and refresh approval baselines

## GitHub Issues
References #3043

## Tests
- `cmake -B /tmp/catch2-basic-build -S . -DCMAKE_BUILD_TYPE=Debug --preset basic-tests`
- `cmake --build /tmp/catch2-basic-build --target SelfTest -j 4`
- `/tmp/catch2-basic-build/tests/SelfTest "Usage of UnorderedRangeEquals range matcher" --rng-seed 1`
- `ctest --test-dir /tmp/catch2-basic-build -R "RunTests|ApprovalTests" --output-on-failure -j 2`
- `ctest --test-dir /tmp/catch2-basic-build --output-on-failure -j 4`
- `git diff --check`

### DIFF
--- a/src/catch2/matchers/catch_matchers_range_equals.hpp
+++ b/src/catch2/matchers/catch_matchers_range_equals.hpp
@@ -12,6 +12,7 @@
 #include <catch2/matchers/catch_matchers_templated.hpp>
 
 #include <functional>
+#include <vector>
 
 namespace Catch {
     namespace Matchers {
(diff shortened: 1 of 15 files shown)


rajpratham1 (approved):
Nice fix. Preserving the comparator argument order for UnorderedRangeEquals makes the matcher behave consistently with user expectations, especially for heterogeneous element types where the parameter order is significant. The regression test clearly demonstrates the issue, and the replacement matching algorithm is straightforward and easy to follow. LGTM.

PR #3116, opened 2026-05-29
docs: Enable syntax coloring for code example

offa:
## Description

While the other code examples in this document use syntax coloring, this one doesn't.


horenmar:
Thanks

PR #3161, opened 2026-06-08
Ensure test class internal linkage for certain template-test types

DNKpp:
As already reported on the official discord, there exists a severe issue with `TEMPLATE_LIST_TEST_CASE_METHOD `.

> Hi, one of our Devs stumbled across a severe issue with catch. We recently started using TEMPLATE_LIST_TEST_CASE_METHOD in multiple TUs. Unfortunately, catch does not define it's internal CATCH2_INTERNAL_TEMPLATE_TEST_<COUNTER> class in an anonymous namespace, which then leads to UB. We just saw that, because one test-case always executes the body of another test-case from another TU. As a workaround we defined each such test-case in an explicit anonymous namespace, which fixes the problem.
image

When fixing that I found a similar issue with `TEMPLATE_PRODUCT_TEST_CASE_METHOD`.


horenmar:
Thanks

PR #3164, opened 2026-06-17
Add CATCH_CONFIG_RUN_ORDER_DECLARED flag

### LINKED ISSUE #3163: Add a compile time option to restore the old test execution ordering
**Description**
We are currently running Catch2 v3.7.1 and wanted to update to the latest version 3.15.1. We have tons of tests spread amongst different executables and it seems that the new default random ordering which was introduced is causing our tests to fail big time. I know the root cause of the issue are the tests themself, but at this scale fixing that is far from economical. I guess that many other codebases have that issue. In fact by looking at the commit which changed the ordering, it seems you have the same issue too - https://github.com/catchorg/Catch2/commit/a483b6d7d36457b0931eca85fda1a30a2b56eb4e

I would like to request a compile time variable which can be set via CMake that can restore the old behavior. I even plan to submit a PR for that myself.

EDIT: Opened PR #3164

TheStormN:
## Description

This PR adds a configure and a compile time option called `CATCH_CONFIG_RUN_ORDER_DECLARED` which can be used to restore the old default test run behavior as it was before Catch v3.9.0.

## GitHub Issues

Closes #3163


src/catch2/catch_config.cpp
```diff
@@ -96,6 +96,12 @@ namespace Catch {
         return lhs.type == rhs.type && lhs.filter == rhs.filter;
     }
 
+    ConfigData::ConfigData() {
+#if defined( CATCH_CONFIG_RUN_ORDER_DECLARED )
+        runOrder = TestRunOrder::Declared;
+#endif
+    }
+
```
TheStormN, line 104: I wondered if I should add the compile time check to the header directly instead of creating a constructor, but I see that the catch_user_config.hpp is not included there and this is why I leaned towards creating a constructor. However if including the user config directly in catch_config.hpp is considered ok, I can move it there.

Issue #3099, opened 2026-05-12
Print the active section/generator selection


jtunhag:
**Description**
I sometimes find it hard to correctly select the section and/or generator when digging through failing tests. A part of this is that Catch2 provides very little info about what I have actually selected with the given options.

I would like the whole test spec to be printed, just like it is when no path/section spec is given. I would also like the given filter/generator selection to be printed, just like the `Filters: ` printout.

I'm currently using Catch2 v3.14.0. 

**Additional context**
This is the complete printout I currently get when I've made a selection that leads nowhere, which doesn't help me find out what I need to change to reach the expected assertion.

```
Filters: "Scenario: Create and train model on entire dataset"
Randomness seeded to: 3221299838
===============================================================================
test cases: 1 | 1 passed
assertions: - none -
```

ShivaPriyanShanmuga:
I’d like to work on this issue if it’s still open. Could you assign it to me?

ShivaPriyanShanmuga:
Hi @jtunhag, I've opened a PR that addresses this: #3165 

When a section/generator selection (-c, -g, or -p) is active, the console and compact reporters now print it on a Path filters: line right below the existing Filters: line:

Filters: "Scenario: Create and train model on entire dataset"
Path filters: c:"Given: a fresh dataset"
Randomness seeded to: 3221299838
Section filters show as c:"<filter>" and generator filters as g:<filter>, mirroring the -p/--path-filter syntax — so it's clear exactly what you've selected, even when a combination ends up running no assertions. Would appreciate your feedback! :)

jtunhag:
Nice! Personally, I'd prefer more verbose output, spelling out "Section: " and "Generator: " instead of "c" and "g". It's meant to be read, after all. Maybe even output as a list, e.g.

```
Filters: "Scenario: Create and train model on entire dataset"
Path filters: 
  - Section: "Given: a fresh dataset"
  - Generator: "Something..."
```

ShivaPriyanShanmuga:
hmm yea! agreed that it should read naturally. I've updated the PR to spell it out as a labelled list:

```
Filters: "Scenario: Create and train model on entire dataset"
Path filters:
  - Section: "Given: a fresh dataset"
  - Generator: "0"
```
Filters are listed one per line in the order specified. Let me know if this looks good. Thanks! :D

PR #3165, opened 2026-06-20
Print the active section/generator selection

ShivaPriyanShanmuga:
## Description

When a section/generator selection (`-c`/`--section`, `-g`/`--generator-index`,
or `-p`/`--path-filter`) is in effect, the console and compact reporters now
print the active path filters on a `Path filters:` line, right below the
existing `Filters:` line and using the same highlight colour:

```text
Filters: "foo"
Path filters: c:"A" g:0
```

Section filters are serialized as `c:"<filter>"` and generator filters as
`g:<filter>`, mirroring the `-p`/`--path-filter` syntax. The line is only
printed when path filters are present, so default runs are unaffected.

The motivation: Catch2 currently gives no feedback about what was actually
selected, which makes it hard to tell what went wrong when a filter
combination ends up running no assertions, e.g.:

```text
Filters: "Scenario: Create and train model on entire dataset"
Randomness seeded to: 3221299838
test cases: 1 | 1 passed
assertions: - none -
```

Implementation notes:
- Added a `serializePathFilters()` helper in `catch_reporter_helpers`,
  alongside the existing `serializeFilters()`.
- Wired it into `ConsoleReporter::testRunStarting` and
  `CompactReporter::testRunStarting`.
- Documented the printout in `docs/filtering-execution-path.md`.
- Tests: a unit test for `serializePathFilters` (format + ordering) and two
  CTest integration tests verifying both reporters emit the line. Baselines
  were updated to account for the new self-test case.

Structured reporters (XML/JSON/JUnit/TAP/SonarQube) are intentionally left
out of scope; this could be a follow-up if desired.

## GitHub Issues

Implements the request in #3099.

### DIFF
--- a/docs/filtering-execution-path.md
+++ b/docs/filtering-execution-path.md
@@ -19,6 +19,22 @@ behaviour, which does not affect generators at all. If you also use either
 `-g`/`--generator-index`, or `-p`/`--path-filter`, you will get the new
 behaviour, which can also filter generator elements.
 
+When path filters are in effect, the console and compact reporters print
+the active selection at the start of the run, next to the `Filters:` line,
+e.g.
+
+```text
+Filters: "foo"
+Path filters:
+  - Section: "A"
+  - Generator: "0"
+```
+
+The filters are listed one per line, labelled as `Section` or `Generator`,
+in the order in which they were specified. This makes it easy to see exactly
+what has been selected when a filter combination ends up running no
+assertions.
+
 Both the new and old filter behaviours include some potentially surprising
 things:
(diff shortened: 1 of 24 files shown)


rajpratham1 (approved):
Nice usability improvement. Printing the active section/generator path filters makes it much easier to understand why a filtered run produced a particular result, especially when no assertions execute. I also like that the formatting logic is centralized in serializePathFilters and backed by both unit and integration tests. The documentation and baseline updates look comprehensive as well. LGTM.

SagarJadhav30 (commented):
like it

PR #3166, opened 2026-06-20
Ignore non-JSON stdout preamble in catch_discover_tests (#3162)

### LINKED ISSUE #3162: catch_discover_tests() should process only json portion of stdout
**Describe the bug**
I'm trying to update a CMake based project from Catch 3.7.1 to 3.15.1. It seems in recent versions of Catch, the `catch_discover_tests()` is now using `json` output for registering the tests. However, we have a 3rd party libraries which are doing static object initialization which are printing stuff to `stdout` like "Started up succesfully", etc. This is effectively breaking the `json` parser as it considers this info to be part of the tests output, Before that was not the case as such things were quietly ignored.

**Expected behavior**
As it is not always possible to address issues coming from 3rd party libraries it would be really nice of the CMake test discovery mechanics just ignore any text which is printed before the json output.

**Reproduction steps**
Just create a static global variable which prints to `stdout` from its constructor.

**Platform information:**

 - OS: **Windows 11**
 - Compiler+version: **MSVC 14.44**
 - Catch version: **v3.15.1**

**Additional context**
Of course I'm not asking for handling situations where printing happens from other threads during mid-json discovery. It is just about global objects printing stuff from their constructors during startup.

ShivaPriyanShanmuga:
### What

`catch_discover_tests()` runs the test executable with
`--list-tests --reporter json` and feeds the **entire** stdout into CMake's
`string(JSON ...)`. If anything writes to stdout *before* `main` runs - most
commonly a third-party library logging from a static initializer - that text
precedes Catch2's JSON and the parser fails with:

string sub-command JSON failed parsing json string: * Line 1, Column 1
Syntax error: value, object or array expected.

Older versions used the plain `--list-tests` text output, which tolerated such
noise, so this is a regression people hit when upgrading.

Closes #3162.

### How

Before parsing, strip everything before the first `{` (where Catch2's JSON
output begins) in `extras/CatchAddTests.cmake`. If no `{` is present at all,
fail with a clear error that includes the captured output.
This intentionally covers only the case described in the issue - output printed
*before* discovery starts (e.g. static-init constructors). It does not attempt
to handle output interleaved into the middle of the JSON.

### Tests

Extended the existing `CMakeHelper::DiscoverTests` integration test:
- `register-tests.cpp` now prints to stdout from a static initializer, so the
  discovered executable reproduces the scenario.
- `VerifyRegistration.py` parses the executable's raw stdout directly, so it
  skips to the first `{` the same way the CMake script does.

Verified locally: the test fails on `devel` (JSON syntax error during
discovery) and passes with this change (`5 tests matched`).

### DIFF
--- a/extras/CatchAddTests.cmake
+++ b/extras/CatchAddTests.cmake
@@ -134,6 +134,19 @@ function(catch_discover_tests_impl)
     endforeach()
   endif()
 
+  # The test executable might print to stdout before Catch2 emits its JSON,
+  # e.g. a third-party library logging from a static initializer. Such text
+  # would break the JSON parser, so we strip everything before the first '{',
+  # which is where Catch2's JSON output starts.
+  string(FIND "${listing_output}" "{" json_start)
+  if(json_start EQUAL -1)
+    message(FATAL_ERROR
+      "Could not find the start of JSON output when listing tests from executable '${_TEST_EXECUTABLE}':\n"
+      "  Output: ${listing_output}\n"
+    )
+  endif()
+  string(SUBSTRING "${listing_output}" ${json_start} -1 listing_output)
+
   # Parse JSON output for list of tests/class names/tags
   string(JSON version GET "${listing_output}" "version")
   if(NOT version STREQUAL "1")
--- a/tests/TestScripts/DiscoverTests/VerifyRegistration.py
+++ b/tests/TestScripts/DiscoverTests/VerifyRegistration.py
@@ -79,7 +79,12 @@ def get_test_names(build_path: str) -> List[TestInfo]:
                             check = True,
                             text = True)
 
-    test_listing = json.loads(result.stdout)
+    # The executable might print to stdout before Catch2's JSON output, e.g.
+    # from a third-party library's static initializer, so we skip everything
+    # before the first '{', just like CatchAddTests.cmake does.
+    json_start = result.stdout.find('{')
+    assert json_start != -1, f"Could not find JSON output in:\n{result.stdout}"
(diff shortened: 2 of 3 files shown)


rosewilliams9001rx-rgb:
> ### What
> `catch_discover_tests()` runs the test executable with `--list-tests --reporter json` and feeds the **entire** stdout into CMake's `string(JSON ...)`. If anything writes to stdout _before_ `main` runs - most commonly a third-party library logging from a static initializer - that text precedes Catch2's JSON and the parser fails with:
> 
> string sub-command JSON failed parsing json string: * Line 1, Column 1 Syntax error: value, object or array expected.
> 
> Older versions used the plain `--list-tests` text output, which tolerated such noise, so this is a regression people hit when upgrading.
> 
> Closes #3162.
> 
> ### How
> Before parsing, strip everything before the first `{` (where Catch2's JSON output begins) in `extras/CatchAddTests.cmake`. If no `{` is present at all, fail with a clear error that includes the captured output. This intentionally covers only the case described in the issue - output printed _before_ discovery starts (e.g. static-init constructors). It does not attempt to handle output interleaved into the middle of the JSON.
> 
> ### Tests
> Extended the existing `CMakeHelper::DiscoverTests` integration test:
> 
> * `register-tests.cpp` now prints to stdout from a static initializer, so the
>   discovered executable reproduces the scenario.
> * `VerifyRegistration.py` parses the executable's raw stdout directly, so it
>   skips to the first `{` the same way the CMake script does.
> 
> Verified locally: the test fails on `devel` (JSON syntax error during discovery) and passes with this change (`5 tests matched`).

Hello

PR #3169, opened 2026-07-07
[CMake]Optimize JSON parsing in catch_discover_tests()

### LINKED ISSUE #3168: The new JSON parsing in catch_discover_tests() is really slow.
**Describe the bug**
The project I'm working on have thousands of tests and each test does have a tag used for grouping purposes. Since we've upgraded to the latest version of Catch, we noticed that our pipeline is slower with several minutes. Tracing the issue lead us to the changes inside `catch_discover_tests_impl()` which switched from text parsing to JSON parsing. Unfortunately CMake JSON parsing API is really slow as it reparses everything on invocation...

**Expected behavior**
Have a better performance. I know that if we keep the JSON parsing it will be fore sure slower compared to before but I see that with a few changes we can have some big wins(about 20-30%). For example, even though we have tags on our tests, we currently don't use them as labels, thus with few adjustments a lot of parsing(per iteration) can be skipped.

**Reproduction steps**
Use `catch_discover_tests()` on a test executable with thousands of tests where each tests also have a tag.

**Platform information:**

 - OS: **Windows 11**
 - Compiler+version: **MSVC 19.51.36248**
 - Catch version: **v3.15.1**

**Additional context**
Opened PR #3169. This PR is simple fix for removing unneeded JSON parsing step. However, maybe a better solution would be to check if converting the tags to labels is not needed, then perhaps we can restore the original pure texting parsing logic?

TheStormN:
## Description
Optimizes the CMake JSON parsing logic in `catch_discover_tests()` by avoiding materialization of full per-test JSON object when `ADD_TAGS_AS_LABELS` is not used.

## GitHub Issues
Closes #3168

### DIFF
--- a/extras/CatchAddTests.cmake
+++ b/extras/CatchAddTests.cmake
@@ -199,9 +199,13 @@ function(catch_discover_tests_impl)
   math(EXPR num_tests "${num_tests} - 1")
 
   foreach(idx RANGE ${num_tests})
-    string(JSON single_test GET ${test_listing} ${idx})
-    string(JSON test_tags GET "${single_test}" "tags")
-    string(JSON plain_name GET "${single_test}" "name")
+    if(add_tags)
+      string(JSON single_test GET "${test_listing}" ${idx})
+      string(JSON test_tags GET "${single_test}" "tags")
+      string(JSON plain_name GET "${single_test}" "name")
+    else()
+      string(JSON plain_name GET "${test_listing}" ${idx} "name")
+    endif()
 
     # Escape characters in test case names that would be parsed by Catch2
     # Note that the \ escaping must happen FIRST! Do not change the order.


extras/CatchAddTests.cmake
```diff
@@ -199,9 +199,13 @@ function(catch_discover_tests_impl)
...
-    string(JSON single_test GET ${test_listing} ${idx})
-    string(JSON test_tags GET "${single_test}" "tags")
-    string(JSON plain_name GET "${single_test}" "name")
+    if(add_tags)
+      string(JSON single_test GET "${test_listing}" ${idx})
+      string(JSON test_tags GET "${single_test}" "tags")
+      string(JSON plain_name GET "${single_test}" "name")
+    else()
+      string(JSON plain_name GET "${test_listing}" ${idx} "name")
+    endif()
```
TheStormN, line 208: In short if `add_tags` is `false` we do only a single round of parsing the JSON data. Unfortunately if tags are used as labels, the performance will be as before.

horenmar:
Sure, why not.

Issue #2411, opened 2022-04-19
"Result: Subprocess killed" build/runtime issue with CMake & Xcode generator on macOS with Apple Silicon


gegles:
using latest macOS Monterey and Xcode (13.3.1)

See minimal repro: [catch2-v3-issue.zip](https://github.com/catchorg/Catch2/files/8516900/catch2-v3-issue.zip)

The same thing seems to work just fine on Intel + Xcode.

Working scenario (Makefile/default generator):
1. `cmake -S. -B_builds/macos-makefiles -DCMAKE_PREFIX_PATH=/tmp/catch2` (where /tmp/catch2 is where I installed 17a04f88d90d3687bd561e89dab40e4ba6a55b83)
2. `cmake --build _builds/macos-makefiles`
3. `./_builds/macos-makefiles/c2issue_tests`
Everything builds and runs fine:
```
➜  catch2-v3-issue cmake -S. -B_builds/macos-makefiles -DCMAKE_PREFIX_PATH=/tmp/catch2
-- The C compiler identification is AppleClang 13.1.6.13160021
-- The CXX compiler identification is AppleClang 13.1.6.13160021
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/homebrew/opt/ccache/libexec/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/homebrew/opt/ccache/libexec/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/gegles/Workspaces/playground/catch2-v3-issue/_builds/macos-makefiles
➜  catch2-v3-issue cmake --build _builds/macos-makefiles
[ 25%] Building CXX object CMakeFiles/c2issue.dir/main.cpp.o
[ 50%] Linking CXX executable c2issue
[ 50%] Built target c2issue
[ 75%] Building CXX object CMakeFiles/c2issue_tests.dir/main.tests.cpp.o
[100%] Linking CXX executable c2issue_tests
[100%] Built target c2issue_tests
➜  catch2-v3-issue ./_builds/macos-makefiles/c2issue_tests
Randomness seeded to: 120581041
===============================================================================
All tests passed (1 assertion in 1 test case)
```

[...]

PR #3172, opened 2026-07-10
Document that DISCOVERY_MODE PRE_TEST should be used with catch_discover_tests when used with XCode.

philsquared:
## Description

This addresses https://github.com/catchorg/Catch2/issues/2411 by making the "workaround" the official approach and is consistent with how gtest and cmake deal with the same situation.

When using the Xcode generator on Apple Silicon using the default discover mode, `POST_BUILD`, macOS will refuse to run due to being unsigned. The build is signed later in the process - after the post build script. So the remedy is to use the `PRE_TEST` discover mode, that runs _after_ the code has been signed.

Changing the default discovery mode for Xcode to `POST_BUILD` was considered, but this would have been a change in behaviour for Intel Mac users (if there are any left), so being explicit seemed to be the better option.

We could also have pre-signed (which would have been later overwritten) but this seemed more heavyweight, disruptive and potentially error prone.

## GitHub Issues

Addresses #2411

### DIFF
--- a/docs/cmake-integration.md
+++ b/docs/cmake-integration.md
@@ -212,6 +212,17 @@ execution (useful e.g. in cross-compilation environments).
 calling ``catch_discover_tests``. This provides a mechanism for globally
 selecting a preferred test discovery behavior.
 
+_Note that on Apple Silicon with the Xcode generator you must use `PRE_TEST`,
+e.g. `catch_discover_tests(tests DISCOVERY_MODE PRE_TEST)`. With the default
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
Nice documentation improvement. This captures a platform-specific issue that can be difficult to diagnose and clearly explains both the cause (Xcode code-signing order) and the workaround (DISCOVERY_MODE PRE_TEST). Adding the note to both the user documentation and the CMake helper comments is a good touch. No concerns from my side.

PR #3176, opened 2026-07-13
Ensure locale is not used when writing int to json stream

spchamp:
## Description

If a locale has been introduced to a stream, then the locale may provide a certain coding convention for integral values. 

For instance, the locale may introduce a comma character or space character, as a delimiter within the string representation of an integral value much greater than 999. This might be visible when printing the value of Config::rngSeed in test report data, e.g with `--list-tests -r json`

```json
{
  "version": 1,
  "metadata": {
      "name": "test_mywidget",
      "rng-seed": 3,492,809,809,
      "catch2-version": "3.15.2"
    },
  // ...
}
```

Of course, this would not be valid JSON. For these tests, the cmake function `catch_discover_tests()` may fail due to the locale-formatted output. 

This has been observed with tests using the Gtk toolkit in versions 3 and 4, and in some other test cases.

If the value would be written as an unquoted string, first converted with e.g `std::to_string`, then this should serve to work around the locale encoding for integral values.

Known limitations
- Not tested: Writing float values, whether as JSON float or quoted string (Not seen in Config)

## GitHub Issues

(None)

### DIFF
--- a/src/catch2/internal/catch_jsonwriter.hpp
+++ b/src/catch2/internal/catch_jsonwriter.hpp
@@ -14,6 +14,8 @@
 
 #include <cstdint>
 #include <sstream>
+#include <string>
+#include <type_traits>
 
 namespace Catch {
     class JsonObjectWriter;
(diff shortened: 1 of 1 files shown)


spchamp:
fwwiw the appveyor test seems to have failed over the availability of a 'codecov' command ([log](https://ci.appveyor.com/project/CatchOrg/catch2/builds/54369785/job/3edydip4f3q4ck65#L694)). (I'm not immediately familiar with the code coverage tests lol)

horenmar:
Yeah, AppVeyor sometimes fails that way, the VMs there can sometimes get broken in a weird way.

Anyway, the intention of the change is useful, but

1) This is missing actual tests.
2) I would like to see a non-locale conversion for floats added at the same time.

PR #3177, opened 2026-07-15
Remove debug message for environment variable check

KnoerleMaTLS:
## Description
The `make_temp_file_path` function had a leftover debugging message call and therefore produced a lot of noise in the build output. This PR removes this leftover since it does not seem to be required anymore (see https://github.com/catchorg/Catch2/commit/9ec44dd62b1b593e01feaf95eb485d26fcdd50ce#r192489445)


horenmar:
Thanks

PR #3181, opened 2026-07-24
Fix typo in generator exception message ("Coud" -> "Could")

dvirarad:
The exception message thrown by `throw_generator_exception()` when a generator cannot jump to the Nth element had "Coud" misspelled instead of "Could". The same message string appears in three places, all fixed here:

- `src/catch2/generators/catch_generators.hpp`
- `src/catch2/generators/catch_generators_adapters.hpp`
- `src/catch2/interfaces/catch_interfaces_generatortracker.cpp`

This is a user-facing string (surfaced when a generator is asked to skip past its available elements). No functional change.


horenmar:
Thanks

Issue #3180, opened 2026-07-21
Clock resolution estimation iterations limit


Benzinnos:
**Description**
Hello. I'm trying to run a benchmark on an embedded system. The system is `qemu-system-arm -machine mps2-an386` with 16MB RAM. It works, but I'm facing the following issue: during the `measure_environment` process, specifically in `estimate_clock_resolution`, it consumes all my available RAM and throws the `std::bad_alloc` exception.

The problem is that `estimate_clock_resolution` is only limited by time. At the same time, this is a hard-coded `clock_resolution_estimation_time` limit of 500ms and this cannot be configured without modifying the source code. It may be worth adding a limit on the number of iterations or allowing it to be configured. Alternatively, it may be possible to configure the time limit.

I suggest adding the ability to limit the number of iterations or the time limit using command-line parameters, similar to `--benchmark-warmup-time` e.g. `--benchmark-estimate-clock-time`.

**Additional context**
During the `measure_environment` process, the number of iterations reached 1'280'000, which resulted in the allocation of 1'280'000 `std::vector<TimePoint<Clock>>` type elements in `std::vector<TimePoint<Clock>> times(points)` and 1'280'001 `double` type elements in `std::vector<double> deltas`.

<img width="1061" height="506" alt="Image" src="https://github.com/user-attachments/assets/7c46f49a-34a0-4b4f-9ec0-3e2ec7b144ed" />

PR #3182, opened 2026-07-26
Adding safeguard to clock resolution estimation (Issue 3180)

superman15625:
## Description
Added a limit to the variable "iter" in order to prevent the chances of bad_alloc errors on systems that have high resolution. The limit is currently at 100,000. The resolution calculation will be stopped when either "iter" has surpassed the limit, or the 500ms gap has passed. Furthermore, the code currently creates an TimePoint<Clock> for each "iter" (the variable passed into resolution()), which can led to memory issues. Instead, this fix only stores the previous timestamp as that is what is needed to determine the delta on the fly. 
## GitHub Issues
This is a PR aimed to fix the issue #3180 where an user ran the benchmark on a system that was able to run over 1 million iterations in under 500ms, causing a bad_alloc exception due to the amount of TimePoint<Clock> instantiations. First, I decided to implement the iteration limit, but then I realized that there wasn't a need to store every single TimePoint. As such, I decided to compute deltas by only storing the current and previous timestamp information.

This is my first pull request, so I won't jump the gun and say the issue is closed.
Thanks :)

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -82,6 +82,7 @@
 ### Fixes
 * Fixed `--warn InfiniteGenerators` triggering even if the generator was limited to specific element via path filtering.
 * Fixed `-Wunused-parameter` triggering in `-fnoexceptions` builds.
+* Fixed benchmark clock-resolution estimation potentially exhausting memory on fast or memory-constrained platforms. (#3180)
 
 ### Improvements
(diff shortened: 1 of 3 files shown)


superman15625:
The reason for the AppVeyor build fail is due to an InternalBenchmarkTest that does:
`InternalBenchmark.tests.cpp(434): FAILED:
  CHECK( (end - start).count() == 2867251000 )`
With the changes (limiting how far the fake clock can run), the new result is 1847245000.

PR #3184, opened 2026-07-27
(v2.x) Remove space in user-defined literal operators to fix MSVC C4996 warning

ZheFeng7110:
Remove the space between "" and the suffix in user-defined literal operators (e.g. operator ""_a instead of operator "" _a) to avoid MSVC deprecation warning C5311.

## Description

In MSVC, `operator "" _a` (with a space between `""` and the suffix) triggers deprecation warning C5311.

The standard-compliant form is `operator ""_a` (no space), as described on [cppreference](https://en.cppreference.com/w/cpp/language/user_literal).

### DIFF
--- a/include/internal/catch_approx.cpp
+++ b/include/internal/catch_approx.cpp
@@ -73,10 +73,10 @@ namespace Detail {
 } // end namespace Detail
 
 namespace literals {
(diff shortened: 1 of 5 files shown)


horenmar:
Sure, why not.

PR #3186, opened 2026-07-31
Emit warning when using --shard-count (> 1) with --order rand

mattyrazz7:
## Description

Using sharding with the default test ordering (`--order rand`) doesn't behave properly since each `--shard-index` invocation obtains a freshly reordered set, resulting in duplication and omitted tests. This can be fixed by changing the test ordering or using a fixed random seed between invocations, but it's surprising default behavior.

## Change

Mitigate by emitting a warning when the following conditions are met:
* `--shard-count` > 1
* AND `--order` is `rand`
* AND `--rng-seed` is set to `time` or `random-device`

The warning points users to either set `--order decl`, `--order lex`, or pass a fixed value to `--rng-seed`.

### DIFF
--- a/tests/ExtraTests/CMakeLists.txt
+++ b/tests/ExtraTests/CMakeLists.txt
@@ -21,6 +21,37 @@ set_tests_properties(TestSharding::OverlyLargeShardIndex
     PASS_REGULAR_EXPRESSION "The shard count \\(5\\) must be greater than the shard index \\(5\\)"
 )
 
+set(CATCH_SHARDING_WARNING_REGEX "Warning: using sharding .* with random order")
+
+add_test(
+  NAME TestSharding::WarningOnRandomOrderWithoutFixedSeed
+  COMMAND $<TARGET_FILE:SelfTest> --shard-index 0 --shard-count 2 --list-tests
+)
+set_tests_properties(TestSharding::WarningOnRandomOrderWithoutFixedSeed
+  PROPERTIES
+    PASS_REGULAR_EXPRESSION "${CATCH_SHARDING_WARNING_REGEX}"
+)
+
+add_test(
+  NAME TestSharding::NoWarningOnRandomOrderWithFixedSeed
+  COMMAND $<TARGET_FILE:SelfTest> --shard-index 0 --shard-count 2 --rng-seed 12345 --list-tests
+)
+set_tests_properties(TestSharding::NoWarningOnRandomOrderWithFixedSeed
+  PROPERTIES
+    FAIL_REGULAR_EXPRESSION "${CATCH_SHARDING_WARNING_REGEX}"
+)
+
+foreach(shardOrder decl lex)
+  add_test(
+    NAME TestSharding::NoWarningOn${shardOrder}OrderWithoutFixedSeed
+    COMMAND $<TARGET_FILE:SelfTest> --shard-index 0 --shard-count 2 --order ${shardOrder} --list-tests
+  )
+  set_tests_properties(TestSharding::NoWarningOn${shardOrder}OrderWithoutFixedSeed
+    PROPERTIES
+      FAIL_REGULAR_EXPRESSION "${CATCH_SHARDING_WARNING_REGEX}"
+  )
+endforeach()
+
 # The MinDuration reporting tests do not need separate compilation, but
 # they have non-trivial execution time, so they are categorized as
 # extra tests, so that they are run less.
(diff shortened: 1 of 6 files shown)


tests/ExtraTests/CMakeLists.txt
```diff
@@ -21,6 +21,24 @@ set_tests_properties(TestSharding::OverlyLargeShardIndex
...
+    PASS_REGULAR_EXPRESSION "Warning: using sharding \\(--shard-count\\) with random order"
+)
+
+add_test(
+  NAME TestSharding::NoWarningOnRandomOrderWithFixedSeed
+  COMMAND $<TARGET_FILE:SelfTest> --shard-index 0 --shard-count 2 --rng-seed 12345 --list-tests
+)
+set_tests_properties(TestSharding::NoWarningOnRandomOrderWithFixedSeed
+  PROPERTIES
+    FAIL_REGULAR_EXPRESSION "Warning: using sharding \\(--shard-count\\) with random order"
```
horenmar: This should be more generic (less specific), so that small rewording of the err. message doesn't break it.

Or have it be a CMake variable that is used by this & and the test above, so that the two tests are guaranteed to be kept in sync.

tests/CMakeLists.txt
```diff
@@ -712,5 +712,14 @@ set_tests_properties("Bazel::RngSeedEnvVar::MalformedValueIsIgnored"
     PASS_REGULAR_EXPRESSION "Randomness seeded to: 17171717"
 )
 
+add_test(NAME "Sharding::WarnsWithoutFixedSeed"
+  COMMAND
+    $<TARGET_FILE:SelfTest> --shard-index 0 --shard-count 2 --list-tests
+)
```
horenmar: This seems duplicated with the one in `ExtraTests`

tests/ExtraTests/CMakeLists.txt
```diff
@@ -21,6 +21,24 @@ set_tests_properties(TestSharding::OverlyLargeShardIndex
...
+)
```
horenmar, line 42: There should also be a test that the warning doesn't fire without rng-seed, but with declaration/alphabetical order.

horenmar:
Thanks, this will be useful.

tests/CMakeLists.txt
mattyrazz7: Whoops, it is

mattyrazz7: Removed the `ExtraTests` one

tests/CMakeLists.txt
```diff
@@ -712,5 +712,37 @@ set_tests_properties("Bazel::RngSeedEnvVar::MalformedValueIsIgnored"
...
+  string(CAPITALIZE ${shardOrder} shardOrderName)
+  add_test(NAME "Sharding::NoWarningOn${shardOrderName}OrderWithoutFixedSeed"
+    COMMAND
+      $<TARGET_FILE:SelfTest> --shard-index 0 --shard-count 2 --order ${shardOrder} --list-tests
+  )
+  set_tests_properties("Sharding::NoWarningOn${shardOrderName}OrderWithoutFixedSeed"
+    PROPERTIES
+      FAIL_REGULAR_EXPRESSION "${CATCH_SHARDING_WARNING_REGEX}"
+  )
+endforeach()
```
mattyrazz7: Used a loop to test both `lex` and `decl`, but let me know if you want me to do it another way

tests/CMakeLists.txt
```diff
@@ -712,5 +712,37 @@ set_tests_properties("Bazel::RngSeedEnvVar::MalformedValueIsIgnored"
     PASS_REGULAR_EXPRESSION "Randomness seeded to: 17171717"
 )
 
+set(CATCH_SHARDING_WARNING_REGEX "Warning: using sharding .* with random order")
```
mattyrazz7: Went with this to give us more flexibility while still getting the gist of the message

horenmar: Sure

tests/CMakeLists.txt
horenmar: That's fine

horenmar:
1) Move the tests down into ExtraTests
2) Check what the CI is telling you ;-)
3) Fix it

mattyrazz7:
> 1. Move the tests down into ExtraTests
> 
>     2. Check what the CI is telling you ;-)
> 
>     3. Fix it

Sorry for the delay, updated

horenmar:
Thanks

PR #3192, opened 2026-08-14
add warning about default test order causing problems with MPI

kbthomp1:
## Summary

This adds a warning in the documentation about the default (random) test order of execution possibly causing issues for MPI programs using Catch2 after version 3.9.1. 

We recently update Catch2 from version 3.1.0 to the 3.15.1, and we were surprised that our tests started encountering MPI deadlock due the default test execution being changed to random.  The fix is simple (use `--lex` or another non-default ordering option), but a google search was unhelpful.  I hope that including this warning in the documentation helps improve the visibility of the fix.

### DIFF
--- a/docs/command-line.md
+++ b/docs/command-line.md
@@ -460,7 +460,7 @@ be preferred to using `time`. Catch2 uses `std::random_device` by default.
 > [!WARNING]
 > You should not use this option when launching Catch2 tests on multiples processes that require synchronous communication,
 > (e.g., Message Passing Interface (MPI) code),
-> because the test order of execution will **not** be indentical on all processes.
+> because the test order of execution will **not** be identical on all processes.
 
 <a id="libidentify"></a>
 ## Identify framework and version according to the libIdentify standard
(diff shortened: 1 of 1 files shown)


docs/command-line.md
```diff
@@ -457,6 +457,10 @@ Using `random-device` asks for `std::random_device` to be used instead.
 If your implementation provides working `std::random_device`, it should
 be preferred to using `time`. Catch2 uses `std::random_device` by default.
 
+> [!WARNING]
+> You should not use this option when launching Catch2 tests on multiples processes that require synchronous communication,
+> (e.g., Message Passing Interface (MPI) code),
+> because the test order of execution will **not** be indentical on all processes.
```
horenmar: Typo

```suggestion
> because the test order of execution will **not** be identical on all processes.
```

kbthomp1: Good catch, thanks.

horenmar (changes requested):
Warning is fine, but the overall text is wrong; you do want the option with MPI and friends, but using a fixed seed... or use the `--order` option.


Also I suggest implementing a stronger warning than just in docs, similar to here: https://github.com/catchorg/Catch2/commit/0aeb818520028ca05cc2539ed242402ef7a7ae41

horenmar:
Thinking about it a bit more, I mentioned the concept of non-deterministic test order in last release notes, which is also the underlying issue for MPI. The docs should probably explain that (maybe in ordering section? and probably also add a note about it in FAQ) and then mention that this is problem for MPI and similar.

PR #3196, opened 2026-08-27
Cap clock-resolution estimation iterations to avoid OOM

### LINKED ISSUE #3180: Clock resolution estimation iterations limit

22elix3r:
## Summary

Clock-resolution estimation in Catch2's benchmark environment probe is only bounded by a hard-coded 500ms window. On a cheap, high-resolution clock (for example an embedded `qemu-system-arm` target with 16MB RAM) that window can request well over a million `TimePoint` samples, exhausting memory inside `resolution()` and throwing `std::bad_alloc`.

This change adds a sample cap so `warmup()` and `estimate_clock_resolution()` stop doubling once 100000 iterations are reached, even if the time budget has not elapsed.

## Related Issue

Fixes #3180

## Changes Made

- Added `clock_resolution_estimation_iteration_limit` (100000) next to the other environment-probe constants.
- Extended `run_for_at_least` with an optional `max_iterations` argument (default remains the historical `1 << 30` "optimized away" ceiling).
- Passed the new cap from `warmup()` and `estimate_clock_resolution()`, which are the two paths that allocate `O(n)` `TimePoint`s via `resolution()`.
- Left the two-phase sampling in `resolution()` unchanged so consecutive `Clock::now()` calls are still measured with minimal work between them.
- Did not add a new CLI flag: the OOM is caused by unbounded sample count, not the time budget, and the environment probe is cached once per process. A hard cap is the safer default for memory-constrained targets.

## Testing

- Added `clock resolution estimation is iteration-capped` using `counting_clock` at 1ns/tick, which would otherwise demand hundreds of millions of samples to fill 500ms.
- Added `run_for_at_least respects max_iterations`.
- Extended `estimate_clock_resolution` to assert `samples_seen` stays within the cap.
- Updated the `run benchmark` fake-clock tick count for the capped probe.
- Regenerated SelfTest approval baselines for the new cases.
- `ctest` in a Debug `basic-tests` build: **82/82 passed**, including `RunTests` and `ApprovalTests`.

## Notes

- #3182 also targets this issue. That patch rewrote `estimate_clock_resolution` and stopped storing every `TimePoint`, but it did not cap `warmup()` (which uses the same `resolution()` allocator), left SelfTest's exact fake-clock assertion failing, and has had no maintainer review. This PR keeps `run_for_at_least` as the single doubling loop, caps both warmup and resolution collection, and updates the tests.
- 100000 matches `clock_cost_estimation_tick_limit` already in the same header. Peak extra allocation is about 1.6MB (`TimePoint`s + deltas) instead of ~20MB at the 1.28 million samples reported in #3180.
- Amalgamated distribution files were not updated, per contributing guidelines.


22elix3r:
Ready for review. I cannot request a reviewer from this account.

22elix3r:
Withdrawing: I missed that this issue already had earlier open PRs. Sorry for the noise — leaving the field to those patches.

22elix3r:
Withdrawing this PR so the earlier open work on the issue can proceed.

PR #3199, opened 2026-09-06
Stop AnsiSkippingString treating literal 0xff bytes as its own sentinel

### LINKED ISSUE #2960: Issue with printing std::vector content
**Describe the bug**
There seems to be a bug when printing the content of some `std::vector`s.

**Reproduction steps**
```c++
TEST_CASE( "DebugTest", "[DebugTest]" )
{
    std::vector<uint8_t> vec = { 0x4e, 0xf2, 0xde, 0x6a, 0xa2, 0x9b, 0x66, 0x3d, 0x7b, 0x3e, 0x80, 0x2e, 0x22, 0x66, 0x48, 0x01, 0xeb, 0x10, 0x2d,
                                 0xc3, 0x5a, 0xfd, 0x61, 0xd5, 0xcb, 0x21, 0x23, 0xea, 0x94, 0x05, 0xc2, 0x92, 0xe4, 0xab, 0x05, 0xf9, 0xce, 0x65,
                                 0x54, 0xd8, 0x19, 0xd7, 0x0c, 0x04, 0x0d, 0x80, 0x7a, 0x70, 0x9e, 0xc5, 0x61, 0x61, 0xac, 0x09, 0x78, 0x12, 0xf8,
                                 0xaf, 0xf1, 0x16, 0xb0, 0x7f, 0x31, 0x97, 0x6c, 0x0e, 0xa0, 0xaa, 0x05, 0xfd, 0xd7, 0xaf, 0x5f, 0xfc, 0x39, 0xcf,
                                 0x91, 0x6d, 0x96, 0x4d, 0xc6, 0x46, 0xda, 0xc5, 0xae, 0x30, 0x5c, 0x3b, 0x6c, 0x83, 0x34, 0x63, 0x0c, 0x37, 0x05,
                                 0xed, 0x05, 0xf4, 0x01, 0xd5, 0xf6, 0x8e, 0xb3, 0x73, 0x6f, 0x79, 0x5e, 0x56, 0x35, 0x4d, 0x80, 0x69, 0x3a, 0x7d,
                                 0x77, 0x3d, 0xd4, 0x85, 0x54, 0x57, 0x62, 0x54, 0xd0, 0xca, 0x34, 0x0f, 0x6f, 0xe3, 0x88, 0x54, 0x2f, 0xb3, 0xcd,
                                 0xac, 0xfa, 0x09, 0xb3, 0xcc, 0x58, 0xf2, 0xe3, 0x98, 0xd7, 0x8e, 0x7d, 0x11, 0x15, 0x6b, 0xca, 0x67, 0x87, 0x21,
                                 0xb2, 0xa8, 0xea, 0x26, 0x42, 0x5f, 0x7b, 0x96, 0x5c, 0xd0, 0x9f, 0x26, 0xf4, 0x39, 0x61, 0xd0, 0x4a, 0x4b, 0xff };

    REQUIRE( vec == vec );
}
```

Execute the test binary with `"DebugTest" --success`

Result:

Image

**Platform information:**

 - OS: **Windows 11 23H2**
 - Compiler+version: **Visual Studio 2022 v17.13.2 / 19.43.34808**
 - Catch version: **v3.8.0**

afonsojanu:
Closes #2960

`AnsiSkippingString` marks the terminating `m` of a recognized ansi escape sequence by overwriting it with `0xff` internally, and flips it back when a substring is produced. The trouble is a literal `0xff` byte that was already part of the original text (for instance from stringifying raw byte data, as in the linked issue) looks exactly the same as one of these markers, and the existing code has no way to tell them apart.

Two consequences of that, both reachable with plain text that has nothing to do with ansi escapes at all:

- `substring()` blindly turns every `0xff` byte in the output back into `'m'`, silently corrupting real data.
- the reverse iterator, on hitting a `0xff` byte, assumes it must have found one of its own markers and starts scanning backwards for the `\033` that "must" precede it. If the byte was never one of ours, that scan walks straight off the beginning of the string, tripping an assert in debug builds (and undefined behavior in release, matching the exception/crash reports in the issue).

The fix records the byte offsets where a real sentinel was inserted (`m_sentinelPositions`) and checks membership there instead of comparing raw byte values, in both `substring()` and the iterator's `unadvance()`/`tryParseAnsiEscapes()`. A plain `0xff` byte in the text is now left completely alone.

Added a test (`TextFlow::Column handles literal 0xff bytes that are not ansi escapes`) that:
- reproduces the crash on an unmodified build (confirmed it aborts with the exact assert from the issue before this change)
- passes cleanly with the fix
- also checks that genuine ansi-escape wrapping still round-trips correctly, so the existing sentinel behavior isn't disturbed

Ran the full self-test suite before and after; identical pass/fail counts either way (the one pre-existing `[!shouldfail]` TextFlow test is unrelated to this).

### DIFF
--- a/src/catch2/internal/catch_textflow.cpp
+++ b/src/catch2/internal/catch_textflow.cpp
@@ -48,6 +48,9 @@ namespace Catch {
                         break;
                     }
                     // 'm' -> 0xff
+                    m_sentinelPositions.push_back(
+                        static_cast<std::string::size_type>(
+                            cursor - m_string.begin() ) );
                     *cursor = AnsiSkippingString::sentinel;
                     // if we've read an ansi sequence, set the iterator and
                     // return to the top of the loop
@@ -76,11 +79,12 @@ namespace Catch {
         }
 
         AnsiSkippingString::const_iterator AnsiSkippingString::begin() const {
-            return const_iterator( m_string );
+            return const_iterator( m_string, m_sentinelPositions );
         }
 
         AnsiSkippingString::const_iterator AnsiSkippingString::end() const {
-            return const_iterator( m_string, const_iterator::EndTag{} );
+            return const_iterator(
+                m_string, m_sentinelPositions, const_iterator::EndTag{} );
         }
 
         std::string AnsiSkippingString::substring( const_iterator begin,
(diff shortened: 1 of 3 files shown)


horenmar:
In a funny coincidence, I did some work on `TextFlow` today and noticed the same issue. As I went in a different direction, I will merge my changes first, and then take a look at what remains from this PR later this week.

PR #3202, opened 2026-09-08
Pass the parent toolchain to CMake configuration tests

### LINKED ISSUE #2743: ConfigureTestsCommon fail when the compiler is from toolchain file but is not available in the path
The script that does compile test [here](https://github.com/catchorg/Catch2/blob/47a2c969380259e899f471edde84f06a6d687dc9/tests/TestScripts/ConfigureTestsCommon.py#L16)  fails when the compiler that you use is from a toolchain file but is not available in the path. 

I don't know what the correct solution is, but I'll think about it, and get back to you.

vzer200:
## Description

When Catch2 is configured with a toolchain whose compiler is not on PATH, the CMake configuration tests start child builds without that toolchain and fail to find a compiler. Pass the parent's toolchain to those tests and forward it explicitly to CMake, including versions before 3.21 that do not read the environment variable automatically.

The existing behavior is preserved when no toolchain is configured.

## Validation

- Reproduced the original `CMakeConfig::Disable` failure with LLVM MinGW selected only through a toolchain and no compiler on PATH; all four configuration tests pass after the change.
- Independently checked 28 small C++ build/run cases with CMake 3.16.9 and 4.4.3: inherited and relative toolchain paths with spaces, absent/empty variables, explicit override and explicit clearing. The original helper fails compiler discovery on CMake 3.16; the corrected helper succeeds.
- Full `all-tests` build completed. CTest: 143/144 pass; the existing Windows `Reporters::CrashInJunitReporter` failure reproduces on the unmodified base. Two unrelated Clang 23 warnings were reproduced on the base and kept as warnings in the local validation toolchain.

Implemented and independently reviewed by automated agents before submission.

## GitHub Issues

Fixes #2743.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -645,6 +645,11 @@ if(CATCH_ENABLE_CONFIGURE_TESTS)
         COST 240
         LABELS "uses-python"
     )
+    if(CMAKE_TOOLCHAIN_FILE)
+      set_tests_properties("CMakeConfig::${testName}"
+        PROPERTIES ENVIRONMENT "CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}"
+      )
+    endif()
   endforeach()
 endif()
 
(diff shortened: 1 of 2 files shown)


horenmar:
Thanks

PR #3204, opened 2026-09-13
Fix typos in comments

Vir-007:
Three misspellings in comments, found with [codespell](https://github.com/codespell-project/codespell) and checked by hand:

| File | Before | After |
|---|---|---|
| `CMake/Findcodecov.cmake` | `reuired` | `required` |
| `src/catch2/catch_tostring.hpp` | `ambigous` | `ambiguous` |
| `src/catch2/internal/catch_test_case_tracker.cpp` | `neccessarily` | `necessarily` |

Comments only — no functional change.

Two notes:

* Per `docs/contributing.md` I have **not** touched the amalgamated files under `extras/`, even though they currently carry the same `ambigous`/`neccessarily` text. They will pick the fixes up when they are next regenerated.
* The `catch_tostring.hpp` one sits inside a multi-line `#if` continuation, so I verified the trailing `\` is intact and the directive still evaluates to the same branch.

I skipped the other codespell hits in the repo as false positives — `SEH` (Structured Exception Handling), `nd` (a local variable in `catch_stats.cpp`), and `statics` (used correctly, as in "class-level statics"). I also left `docs/release-notes.md` alone since it is a historical record.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


horenmar:
Thanks

